PR #88, opened 2015-02-17
ACL update/revision Activities/small issue fix in company review

choutha:
added version of Activity --> not that important to rest of project. This is a simple version with key functionallity

changed isAllowed function in AbstractAclService --> important since it is about general authorization. Changed the method such that it checks the actual roles instead of the user object against several role strings

changed the approver variable in CompanyReview --> this was giving errors when doctrine was creating the database. Now changed the relation to a OneToMany relation as it should be (discussed with Geert)


module/Activity/src/Activity/Model/ActivitySignup.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Who is subscribed
+     *
+     * @ORM\Column(nullable=false)
+     * @ORM\ManyToOne(targetEntity="Activity\Model\Activity", inversedBy="roles")
+     * @ORM\JoinColumn(referencedColumnName="id")
+     */
+    protected $activity_id;
+
```
jszanto: De casing hier zou volgens mij camelCase moeten zijn om consistent te blijven met de rest.

choutha: You mean it should be $activityId instead of $activity_id?

choutha:
Already other pull request ready with a full implementation of organ ACL working. Not yet uploaded since this one first needs to be accepted

module/Activity/src/Activity/Model/ActivitySignup.php
jszanto: yes I think so.

module/Application/src/Application/Service/AbstractAclService.php
```diff
@@ -34,27 +34,69 @@ public function getRole()
...
+		
+		var_dump($roles);
+		
+		foreach($roles as $role){
+			if($this->getAcl()->isAllowed($role, $resource, $operation)){
+				return true;
+			}
+		}
+		return false;
+    }
```
kokx: Dit is dus juist fout. Ten eerste staat er nog een `var_dump` in. Maar vooral, de roles van een user zijn al goed gezet in de ACL, waardoor dit helemaal onnodig is. Zie https://github.com/GEWIS/gewisweb/blob/master/module/User/Module.php#L154

jszanto:
Vergeet niet om alle tekst in de views door translate() heen te halen.

PR #139, opened 2015-07-28
Photo: issue #55

jszanto:


module/Decision/src/Decision/Controller/MemberController.php
```diff
@@ -19,6 +20,23 @@ public function indexAction()
...
 
     /**
+     * Search action, allows searching for members.
+     */
+    public function searchAction()
+    {
+        $name = $this->params()->fromRoute('name');
+        $members = array();
+        foreach ($this->getMemberService()->findMemberByName($name) as $member) {
+            //TODO: this returns a lot of data, much more than is needed in most cases.
```
kokx: Wouldn't just limiting the number of results be good enough here?

module/Decision/src/Decision/Mapper/Member.php
```diff
@@ -39,6 +39,25 @@ public function findByLidnr($number)
...
+    public function findByName($firstName, $lastName)
+    {
+        $qb = $this->getRepository()->createQueryBuilder('m');
+
+        $qb->where('m.firstName LIKE ?1')
+            ->orWhere('m.lastName LIKE ?2')
+            ->setParameter(1, $firstName . '%')
+            ->setParameter(2, '%' . $lastName . '%');
+
+        return $qb->getQuery()->getResult();
```
kokx: This doesn't allow searching for full names. It might be better to solve this like I solved it in the GEWIS Database:

https://github.com/GEWIS/gewisdb/blob/master/module/Database/src/Database/Mapper/Member.php#L38

Also, limiting the number of results here would be nice as well.

module/Decision/src/Decision/Service/Member.php
```diff
@@ -27,9 +28,47 @@ public function getMembershipInfo()
...
+            throw new \User\Permissions\NotAllowedException(
+                $this->getTranslator()->translate('Not allowed to search for members.')
+            );
+        }
+        $parts = explode(' ', $name);
+        if (count($parts) == 1) {
+            return $this->getMemberMapper()->findByName($name, $name);
+        } else {
+            return $this->getMemberMapper()->findByName($parts[0], last($parts));
+        }
```
kokx: This wouldn't work well for people with multiple first names or a middle name (like 'van'). See my other comment for a DQL-based solution.

kokx:
Also, there is a merge conflict again. Can you fix that as well?

module/Decision/src/Decision/Controller/MemberController.php
jszanto: By aan lot of data I meant that toArray() returns a lot of data such as email, expiration, generation, etc. although limiting the number of results still is a good idea.

kokx: It's better to give too much information here. Because then we can use the same call for multiple search types.

jszanto:
all fixed and rebased

module/Decision/src/Decision/Service/Member.php
```diff
@@ -27,9 +28,43 @@ public function getMembershipInfo()
...
+     * @param string $query (part of) the full name of a member
+     * @pre $name must be at least 3 characters
+     *
+     * @return array|null
+     */
+    public function searchMembersByName($query)
+    {
+        if (strlen($query) < 3) {
+            throw new \Zend\Code\Exception\InvalidArgumentException(
+                $this->getTranslator()->translate('Name must be at least 3 characters')
```
kokx: Why does the name need to be at least 3 characters? 1 or 2 seems much more reasonable here actually.

jszanto: When searching through thousands of names it is unlikely that a 1 character search query will return any relevant results. I guess 2 characters would be reasonable.

kokx: Then make it 2. For names like 'vi'.

jszanto: Made it 2, also made it a const, so it will be even easier to adjust later on if needed.

PR #141, opened 2015-07-29
Fixed fatal error on membership info page.

### LINKED ISSUE #140: Fatal Error on membership info page
`Fatal error: Call to undefined method Decision\Model\OrganMember::getFoundation() in /home/pieter/public_html/gewisweb/module/Decision/view/decision/member/index.phtml on line 91`

kokx:
Also improved processing of memberships. Now only showing memberships of
current organs and they will only be shown once.

Fixes #140

### DIFF
--- a/module/Decision/view/decision/member/index.phtml
+++ b/module/Decision/view/decision/member/index.phtml
@@ -87,20 +87,6 @@ case Member::TYPE_HONORARY:
     </table>
     <h3><?= $this->translate('Membership of committees and fraternities') ?></h3>
 <ul>
-<?php
-$memberships = array();
-foreach ($member->getOrganInstallations() as $install) {
-    if (null !== $install->getDischargeDate()) {
-        continue;
-    }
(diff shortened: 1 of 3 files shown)


module/Decision/view/decision/member/index.phtml
```diff
@@ -87,8 +87,22 @@ case Member::TYPE_HONORARY:
...
+    }
+    if (!isset($memberships[$install->getOrgan()->getAbbr()])) {
+        $memberships[$install->getOrgan()->getAbbr()] = array();
+    }
+    if ($install->getFunction() != 'Lid') {
+        $memberships[$install->getOrgan()->getAbbr()][] = $install->getFunction();
+    }
+}
+?>
+<?php foreach ($memberships as $organ => $functions): ?>
```
jszanto: Might be better to just do this in the controller and then just return the $memberships variable to the view.

kokx: The controller would never be the right place for this. If you do this anywhere else, it's in the Service.

PR #145, opened 2015-07-29
Added list of active organs and organ information page

### LINKED ISSUE #144: Database: Pagina met alle actieve organen
Een pagina met alle actieve organen van GEWIS.

kokx:
Fixes #144 and #143 

Shows basic information about organs

### DIFF
--- a/module/Decision/src/Decision/Controller/OrganController.php
+++ b/module/Decision/src/Decision/Controller/OrganController.php
@@ -23,7 +23,9 @@ class OrganController extends AbstractActionController
      */
     public function showAction()
     {
-        return new ViewModel(array());
+        return new ViewModel(array(
+            'organ' => $this->getOrganService()->getOrgan($this->params()->fromRoute('organ'))
+        ));
     }
 
     /**
--- a/module/Decision/src/Decision/Controller/OrganController.php
+++ b/module/Decision/src/Decision/Controller/OrganController.php
@@ -23,9 +23,13 @@ class OrganController extends AbstractActionController
      */
     public function showAction()
     {
-        return new ViewModel(array(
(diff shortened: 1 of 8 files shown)


module/Decision/src/Decision/Controller/OrganController.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Show an organ.
+     */
+    public function showAction()
+    {
+        return new ViewModel(array(
+            'organ' => $this->getOrganService()->getOrgan($this->params()->fromRoute('organ'))
+        ));
```
jszanto: This should throw a 404 if an organ with the given id doesn't exist.

module/Decision/view/decision/organ/show.phtml
```diff
@@ -0,0 +1,48 @@
...
+            <li><?= $membership['member']->getFullName() ?> <?php if (!empty($membership['functions'])): ?> (<?= implode(', ', $membership['functions']) ?>)<?php endif ?></li>
+        <?php endforeach ?>
+        </ul>
+
+        <h2><?= $this->translate('Old members') ?></h2>
+        <ul>
+        <?php foreach ($oldMembers as $member): ?>
+            <li><?= $member->getFullName() ?></li>
+        <?php endforeach ?>
+        </ul>
```
jszanto: This list could get very long for some organs. Not sure if that is a problem or not.

kokx: The longest page here is from the AC. And it is about three pages. So it is not a problem at all.

module/Decision/src/Decision/Controller/OrganController.php
kokx: Fixed.

PR #162, opened 2015-08-02
Photo: issue #113

jszanto:


module/Photo/Module.php
```diff
@@ -105,4 +125,24 @@ public function getServiceConfig()
...
+     *
+     * @return array
+     */
+    public function getViewHelperConfig()
+    {
+        return array(
+            'factories' => array(
+                'isAllowed' => function ($sm) {
+                    $locator = $sm->getServiceLocator();
+                    $helper = new \Application\View\Helper\Acl();
```
kokx: Since this is an `Application` view helper, it makes more sense to include it in the view helper configuration there.

module/Application/src/Application/View/Helper/Acl.php
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Check if a operation is allowed for the current role.
+     *
+     * @param string $operation Operation to be checked.
+     * @param string|ResourceInterface $resource Resource to be checked
+     *
+     * @return boolean
+     */
+    public function __invoke($operation, $resource)
```
kokx: Since it's not possible to leave the resource empty here in any way, it is better to have the order of the parameters the other way around. So the order is the same as the `isAllowed` and other operations like `allow` and `deny` in the zf2 ACL implementation.

jszanto: I believe this would become a bit confusing since our isAllowed implementation in the service has them in this order https://github.com/GEWIS/gewisweb/blob/master/module/Application/src/Application/Service/AbstractAclService.php#L49

Also I can't think of many cases in which the operation should be empty.

module/Photo/Module.php
jszanto: There's one issue which needs to be resolved if we do this. Each module has its own acl config, how do we obtain the correct acl config?

module/Application/src/Application/View/Helper/Acl.php
kokx: In the service, it is only that way because the resource may be left empty. And we don't actually use the resource often there at all, so I don't think that it will become confusing.

module/Photo/Module.php
kokx: Ah indeed, that was the big problem with the `isAllowed` helper.

I do have an idea to solve this. Instead of a direct `isAllowed` helper, make a general `acl` helper which takes the service locator key for the acl as argument. Then you can load a specific ACL and use it from the view as follows:

```php
$this->acl('decision_acl')->isAllowed('member', 'view');
$this->acl('photo_acl')->isAllowed('photo', 'edit');
```

module/Application/src/Application/View/Helper/Acl.php
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Returns the Acl for a specific module
+     *
+     * @param string $factory Acl factory to load
+     *
+     * @return \Application\View\Helper\Acl
+     */
+    public function __invoke($factory)
+    {
+        $this->acl = $this->getServiceLocator()->get($factory);
```
kokx: It might be a good idea to strictly check if this is indeed an ACL instance, and throw an `InvalidArgumentException` if it is not.

jszanto:
all fixed

PR #169, opened 2015-08-02
Basic structure for the frontpage module

jszanto:
Just the basic structure and a bit of functionality. Might be handy to merge this already to prevent future merge conflicts.


module/Application/src/Application/View/Helper/Acl.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace Application\View\Helper;
+
+use Zend\View\Helper\AbstractHelper;
+
+
+class Acl extends AbstractHelper
```
kokx: This was in a different PR as well!

module/Decision/src/Decision/Service/Member.php
```diff
@@ -87,7 +87,8 @@ public function findMemberByLidNr($lidnr)
      */
     public function getBirthdayMembers($days = 0)
     {
-        if (!$this->isAllowed('birthdays')) {
+        // Allow everyone to view birthdays of today.
+        if ($days > 0 && !$this->isAllowed('birthdays')) {
```
kokx: You should actually add a privilege for showing today's birthdays, and check for that privilige for the case of `$days == 0`

module/Frontpage/config/module.config.php
```diff
@@ -0,0 +1,43 @@
...
+                        'controller'    => 'Frontpage',
+                        'action'        => 'home',
+                    ),
+                ),
+                'may_terminate' => true,
+                'child_routes' => array(
+                    'page' => array(
+                        'type'    => 'Segment',
+                        'options' => array(
+                            'route'    => '/[:category[/:name]]',
```
kokx: Since this is a child route of `/`, it shouldn't start with a `/`. Also, the category wouldn't be optional for this route. And defaults should be created for the optional part. And a controller and action still have to be determined.

Also, since it kinda is a catch-all route, you should set a priority.

module/Frontpage/src/Frontpage/Service/Frontpage.php
```diff
@@ -0,0 +1,131 @@
...
+        $birthdayInfo = $this->getBirthdayInfo();
+        return array(
+            'birthdays' => $birthdayInfo['birthdays'],
+            'birthdayPhoto' => $birthdayInfo['photo']
+        );
+    }
+
+    /**
+     * Retrieves all birthdays happening today, which should be shown on the home page.
+     * Includes the age and a random photo of someone whom has a birthday.
```
kokx: It shouldn't be a random photo. It should be the photo of the most active member (actually, the member with the most photos).

module/Frontpage/view/frontpage/frontpage/home.phtml
```diff
@@ -0,0 +1,39 @@
+<div class="jumbotron">
+    <a href="<?= $this->url('home') ?>">
+        <img src="<?= $this->basePath() ?>/img/logo.png"/>
+    </a>
+    <?php /*
+    <h1><?= sprintf($this->translate('Welcome to %sGEWIS%s'), '<span class="gewis-red">', '</span>') ?></h1>
+    <p><?= sprintf($this->translate('Welcome to the website of Study association %sGEWIS%s. This is just a bootstrap thingy for now, until the design is implemented.'), '<span class="gewis-red">', '</span>') ?></p>
+    <p><a class="btn btn-success btn-lg" href="https://github.com/GEWIS/gewisweb" target="_blank"><?php echo $this->translate('Fork the GEWIS website on GitHub') ?> &raquo;</a></p>
+ */ ?>
+</div>
```
kokx: This bit has changed a lot in the design. See the PR on the design for this one!

module/Photo/Module.php
```diff
@@ -95,6 +95,26 @@ public function getServiceConfig()
...
+                    $acl->allow('guest', 'album', 'view');
+
+                    // Users are allowed to view, remove and add tags
+                    $acl->allow('user', 'tag', array('view', 'add', 'remove'));
+
+                    // Users are allowed to download photos
+                    $acl->allow('user', 'photo', array('download', 'view_metadata'));
+
+                    return $acl;
+                },
```
kokx: These changes really belong in a different PR. Definitely not in one just for the frontpage.

module/Frontpage/config/module.config.php
jszanto: This will be changed in the future once the page part is implemented. (Actually it is changed already on another branch.)

kokx: In that case, it is bad to have the catch-all at all right now. Because it may cause weird bugs with route matching.

module/Photo/Module.php
jszanto: This one is based on the same branch as PR #162 which is why all the changes of PR #162 also get included here. This one probably should wait until PR #162 is resolved.

jszanto:
oops

PR #182, opened 2015-08-03
URL Helpers

jszanto:
This adds two new url helpers. A photo url helper to obtain the correct url for photos (similar to the ExamUrl view helper). And a javascript implementation of zf2's url view helper (which will be documented a bit if merged).

### DIFF
--- a/module/Application/src/Application/View/Helper/ScriptUrl.php
+++ b/module/Application/src/Application/View/Helper/ScriptUrl.php
@@ -13,12 +13,21 @@ use Zend\View\Exception;
  */
 class ScriptUrl extends AbstractHelper
 {
+    /**
+     * @return \Application\View\Helper\ScriptUrl
+     */
     public function __invoke()
     {
         return $this;
     }
 
-    public function requireUrl($name, $params)
+    /**
+     * Makes an url route available to the javascript url helper.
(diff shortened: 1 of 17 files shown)


module/Application/src/Application/View/Helper/ScriptUrl.php
```diff
@@ -0,0 +1,64 @@
...
+        $scriptParams = array();
+
+        foreach($params as $param) {
+            $scriptParams[$param] = '{' . $param . '}';
+        }
+
+        $url = $this->getView()->url($name, $scriptParams);
+        // Include data as inline script
+        $this->getView()->inlineScript()->captureStart();
+        echo 'URLHelper.addUrl("'. $name . '", "'. urldecode($url) . '");';
```
stefanmajoor: This violates the SRP. A PHP either defines logic, or it defines output. This file now does both. In this case there is a hidden dependency between the PHP and JavaScript (the JavaScript URLHelper object), which will be difficult to find since it is not in the template files.  Line 42 - 44 need to be moved to the template files, and the helper should keep an array of all the urls

jszanto: Fixed.

PR #196, opened 2015-08-09
Board approval of activities

stefanmajoor:
Board can now see activities, and approve/disapprove them

Admin url: activity/admin/queue


module/Activity/config/module.config.php
```diff
@@ -58,6 +58,56 @@
                                 'action' => 'create'
                             )
                         )
+                    ),
+                    'admin_queue' => array(
+                        'type' => 'Literal',
+                        'options' => array(
+                            'route' => '/admin/queue',
```
kokx: Admin routes currently have the structure `/admin/<module>/.....`. Please adhere to this in here. Also, it is better to make all these routes that start with `/admin/` childs of a general route.

module/Activity/src/Activity/Controller/AdminController.php
```diff
@@ -0,0 +1,115 @@
...
+                $activityService->approve($activity);
+                break;
+            case 'disapprove':
+                $activityService->disapprove($activity);
+                break;
+            case 'reset':
+                $activityService->reset($activity);
+                break;
+            default:
+                throw new \InvalidArgumentException('No sutch status ' . $status);
```
kokx: The text here should be translated.

stefanmajoor: this is not text that is returned to the user, so it does not have to be translated.

Also for the exception, I assume that will never reach the end user. and we are able to understand Dutch

module/Activity/config/module.config.php
stefanmajoor: I will fix this tomorrow

kokx:
Also, it has some merge conflicts. Please rebase on `master`.

stefanmajoor:
I merged with master (after rebase failed), and the routes have been fixed

module/Activity/src/Activity/Mapper/Activity.php
```diff
@@ -79,12 +94,27 @@ public function getUpcomingActivities($count = null)
      *
      * @return array
      */
-    public function getAllApproved()
+    public function getApprovedActivities()
```
kokx: Because of DRY, is it not better to make one method for this, `findActivitiesByStatus`?

stefanmajoor: Solved. Also only approved activities are shown in the upcomming  activities

PR #207, opened 2015-09-06
Fix upcomming activities (old activities are not displayed)

stefanmajoor:


kokx:
For the future, it is considered good practise for the first line of a commit message to have a maximum of 50 characters.

Source: http://chris.beams.io/posts/git-commit/

Issue #208, opened 2015-10-02
Activity: Costs field should have unknown as value


mstruijs:

kokx:
Was already fixed in #209

jszanto:
This was not properly fixed in #209, which is why it still was open.

stefanmajoor:
It was properly fixed in #209. However some more issues became apparent. These do not have anything to do with this issue, and I created issue #267 for that.

PR #209, opened 2015-10-02
Implemented unknown costs

stefanmajoor:
Will fix issue #208


mstruijs:
Actually, there is such a thing as variable costs, e.g. a cheaper price for freshman. This pretty much an important feature. So turning the costs into a string field is probably a better solution.

jszanto:
Probably best to indeed make it a string and then in the view display 'unknown' if the field is empty.

kokx:
I think this should be made into an additional feature?

PR #218, opened 2015-10-19
Create .codeclimate.yml

efueger:
Add .codeclimate.yml, disable CSSLint

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,19 @@
+engines:
+  eslint:
+    enabled: true
+  csslint:
+    enabled: false
+  phpcodesniffer:
+    enabled: true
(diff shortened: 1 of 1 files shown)


kokx:
Wow. Apparently there exist bots now that automatically create pull requests to push certain config files.

jszanto:
Actually not a a bot

stefanmajoor:
Why is this merged? Are we actually going to use codeclimate? If not, I think this should be reverted

jszanto:
I am currently testing it, it may be useful to help review PRs. For example code climate easily finds stuff like formatting issues, logic errors, unclear code, debug code, etc. This would save a dozen comments everytime we review a PR.

Issue #267, opened 2015-11-22
Activity: Costs should be a string


stefanmajoor:
At this moment costs of an activity is always a number. However, there is a variable costs possible. The easiest way to solve this is by making it a string 

Also see #209 and #208

Issue #195, opened 2015-08-09
Activity:Multiple languages in activity.


mstruijs:
At least support 2. More would probably be neater, but is not strictly necessary.
This includes the creation, representation and handling the absence of descriptions in certain languages.

stefanmajoor:
This has been fixed for the most part in #268. A more detailed ticket about what is left is already created (#297)

PR #268, opened 2015-11-22
 Activity refactor

stefanmajoor:
So I am refactoring activity a bit. The following three things are done

- Costs will be set to a string rather than to a number (see #267)
- English language support (see #195)
- Refactor my inexplainable design choices to better suite the rest of the application (Creating an activity is copied from the model to the service)

This PR is nearly done. Two questions remain:

- What are the constraints on creating an activity? Should it be ok that only the Dutch description is set? Or should both languages be set?
- Why the hell can I not set a creator directly? What is wrong with foreign key definition?


module/Activity/src/Activity/Service/Activity.php
```diff
@@ -110,12 +111,30 @@ public function createActivity(array $params)
...
+            $activity->setCosts($params['costs']);
+            $activity->setCostsEn($params['costs_en']);
+        }
+
+        $activity->setDescription($params['description']);
+        $activity->setDescriptionEn($params['description_en']);
+        $activity->setCanSignUp($params['canSignUp']);
+
+        // Not user provided input
+        $activity->setCreator($user);
```
jszanto: This is actually not a User\Model\User, it is a \User\Model\UserModel. Use $em->merge($user) to obtain a User\Model\User.

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -110,12 +111,30 @@ public function createActivity(array $params)
...
+        $activity->setEndTime(new \DateTime($params['endTime']));
+        $activity->setName($params['name']);
+        $activity->setNameEn($params['name_en']);
+        $activity->setLocation($params['location']);
+        $activity->setLocationEn($params['location_en']);
+
+        if (!$params['costs_unknown']) {
+            $activity->setCosts($params['costs']);
+            $activity->setCostsEn($params['costs_en']);
+        }
```
jszanto:
```
$form->bind($activity);
$form->setData($params);
```
Would save many lines of code here.

module/Activity/src/Activity/Service/Activity.php
stefanmajoor: It is a User\Model\User. I do not know why, it should not be,  but it is. Otherwise the code would fail. But the code does not fail.

jszanto: The code does fail since you indicated you can not set a creator directly. What exception/error does it give?

module/Activity/src/Activity/Service/Activity.php
stefanmajoor: There is till some conversions that are to be done, so it would not save that many lines of code.

The form object should only be used in the controller, since it mostly defines the form that a used sees. The data that is used there may or may not correspond to the data you want to set in your model. Therefore the form object should not be used in the model. Furthermore, that allows activities only to be created using forms (or at least implies this). This may not be the case. The only reason that you might use the form in the service because ZF2 put its validation in its form, which is a design flaw IMO.

module/Activity/src/Activity/Service/Activity.php
stefanmajoor: The code fails in the model, after PHP's type checking confirmed that it was a user object.

The failure is that DBAL complains that it can not convert an User object to a string.

module/Activity/src/Activity/Service/Activity.php
jszanto: If a form only defines what a user sees, then we might as well not use forms, since the form is being rendered manually anyways (https://github.com/stefanmajoor/gewisweb/blob/03a585790c2e17bcf8df33da60a396aedfe2b709/module/Activity/view/activity/activity/create.phtml).

stefanmajoor: That would make the templates much better readable.  I know for example that Laravel takes that approach. No forms, just validators

module/Activity/src/Activity/Service/Activity.php
kokx: Actually, `$user` may be a `User\Model\User` or a string (`guest` or `apiuser` or something else). It is better to actually see if the user has an identity (obtain the auth service and call `hasIdentity()`) and also obtain the identity from the auth service.

module/Activity/src/Activity/Service/Activity.php
kokx: Forms mostly are an interface around validators. Not so much around the visual aspect of forms, especially how they are in ZF2.

stefanmajoor: So, I should use the form instead? The form thingy is too much magic for my liking.

stefanmajoor:
So, @kokx What are the constraints on creating an activity? Should it be ok that e.g. only the Dutch description is set? Or should both languages be set?

kokx:
At least one of the languages should be set. Both is preferable, but we don't need both.

module/Activity/src/Activity/Service/Activity.php
kokx: I don't really see much magic in the forms actually. If you use `bind()` (which is the most magic there is), it will be filled by a hydrator (which you have to set yourself).

module/Activity/src/Activity/Service/Activity.php
jszanto: I just tested this and it works perfectly fine for me. Are you sure your database schema is up to date in your testing environment?

stefanmajoor: That is because this is not the part of the code where weird things happen, and because I have hacked around it. The problem occurs at the following line:

https://github.com/GEWIS/gewisweb/pull/268/files#diff-5f81752c423c0c2d5ddbabad479f495dR346

Why is the ->getLidNr() needed?

kokx: I actually think it shouldn't be needed at all. I even think that that should give an error.

jszanto: It works because of the getLidNr(), it doesn't work without it according to @stefanmajoor

jszanto: I just looked at this and I figured out the actual causes of the issue:
1. Your ORM annotations aren't correct, nullable should be listed inside the JoinColumn statement (don't ask me why)
2. The authentication service returns a User\Model\User object, however this object is not known to doctrine and therefore doctrine can't do anything with it. $em->merge(..) fixes that.

The following patch fixes the issue: https://gist.github.com/jszanto/97741da52d9407e2e5fe

mstruijs:
>At least one of the languages should be set. Both is preferable, but we don't need both.
In that case, I think it would be a good idea to show the description in the other language if the current one is unavailable. (e.g. show the English description if the Dutch one is unavailable on the Dutch site)

stefanmajoor:
So activity creation is completely refactored now, such that one of the languages must be used.

The only thing left to do is that there needs to be some system which displays the correct language for the activity, but that is something for another issue/pull request.

kokx:
There are some recent changes to the master branch that might break a few things in the Activity module. Things that introduce 'plaintext' roles other than `guest` for example. Hence, before this is merged, I am going to check locally if this merge will work out.

module/Activity/src/Activity/Model/Activity.php
```diff
@@ -127,22 +163,202 @@ class Activity
...
+     * @return string
+     */
+    public function getLocation()
+    {
+        return $this->location;
+    }
+
+    /**
+<<<<<<< HEAD
+     * @param string $location
```
jszanto: It seems like you forgot something here.

module/Activity/view/activity/activity/view.phtml
```diff
@@ -81,12 +89,28 @@ echo $this->form()->openTag($form); ?>
...
 <?php echo $this->formSubmit($form->get('submit'));
       echo $this->form()->closeTag();?>
+>>>>>>> c0f9bb68975cf370f3672fdec02a831f85f67fad
 <?php elseif ($canSignUp):?>
        <?php echo $this->translate('Login to subscribe');?>
 <?php endif; ?>
 	</div>
 </div>
 
+<<<<<<< HEAD
```
jszanto: And here in this while file too.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -63,13 +63,18 @@ public function createAction()
                 $data['costs'] = '-1';    // Hack. Because empty string is seen as 0
             }
 
-            $form->setData($this->getRequest()->getPost());
+            $postData = $this->getRequest()->getPost();
+            $form->setData($postData);
 
             if ($form->isValid()) {
```
mstruijs: I think it would be better to validate the form in `createActivity` instead, since it simplifies the logic here and makes the `createActivity` method more robust.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -63,13 +63,18 @@ public function createAction()
...
-                $activity = $activityService->createActivity($form->getData(\Zend\Form\FormInterface::VALUES_AS_ARRAY));
+                $activity = $activityService->createActivity(
+                    $form->getData(\Zend\Form\FormInterface::VALUES_AS_ARRAY),
+                    $postData['language_dutch'],
```
mstruijs: Why is the language data passed separately? Is that not already contained in `$form->getData()` ?

mstruijs:
There is no bilingual support for the additional fields yet. This needs to be done, but that may be put in a later PR (possibly including a refactor of the creation of these fields)

module/Activity/src/Activity/Controller/ActivityController.php
stefanmajoor: No it is not, since the value is not part of the activity model

module/Activity/src/Activity/Controller/ActivityController.php
stefanmajoor: I think that is something for another pull request.

PR #270, opened 2015-11-23
API authentication

kokx:
Makes it possible to add API tokens and to authenticate as API application via the `X-Auth-Token` header.


module/User/src/User/Model/ApiUser.php
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Get the API user's role ID.
+     *
+     * @return string
+     */
+    public function getRoleId()
+    {
+        return 'api_' . $this->getLidnr();
+    }
```
jszanto: This won't work

jszanto:
Being able to revoke tokens might be handy

kokx:
Now also supports removing (or revoking) tokens

PR #273, opened 2015-11-24
Uploading and viewing summaries

kokx:
Added the ability to upload and view summaries


module/Education/src/Education/Form/Upload.php
```diff
@@ -4,9 +4,13 @@
...
 use Zend\Form\Form;
 use Zend\InputFilter\InputFilter;
+use Zend\InputFilter\InputFilterProviderInterface;
 use Zend\I18n\Translator\TranslatorInterface as Translator;
 
-class Upload extends Form
+/**
+ * Upload a summary
+ */
+class Upload extends Form implements InputFilterProviderInterface
```
stefanmajoor: Why not call the file/class UploadSummary instead? That makes the comment superfluous

module/Education/src/Education/Service/Exam.php
```diff
@@ -166,7 +166,7 @@ public function tempUpload($post, $files)
     public function upload($post, $files)
```
stefanmajoor: why does the upload method in the Exam service upload a summary?

stefanmajoor:
Why are there so many diffs where code for uploading exams is changed by code to upload summaries. Should exams no longer be uploaded?

kokx:
There is a new way to upload exams. Which uploads exams in bulk (which is the case every quartile, when the education officer receives new exams of the exam weeks). Summaries, however, won't need to be uploaded in bulk.

module/Education/src/Education/Service/Exam.php
kokx: The naming may be improved here indeed. Will take a look at that.

module/Education/src/Education/Form/Upload.php
kokx: The naming may be improved here indeed. Will take a look at that.

kokx:
Fixed all the previous problems. Can this be reviewed @jszanto or @stefanmajoor ?

PR #274, opened 2015-11-24
Added support for dynamic study years

### LINKED ISSUE #174: Education: make year dynamic
https://github.com/GEWIS/gewisweb/blob/master/module/Education/src/Education/Oase/Client.php#L69

kokx:
Fixes #174

### DIFF
--- a/module/Education/src/Education/Oase/Client.php
+++ b/module/Education/src/Education/Oase/Client.php
@@ -49,10 +49,11 @@ class Client
      *
      * @param array $studies
      * @param string $lang
+     * @param string $year
      *
      * @return array
      */
-    public function ZoekActiviteitenOpDoelgroep($groups, $lang)
+    public function ZoekActiviteitenOpDoelgroep($groups, $lang, $year)
     {
         $vraag = new Vraag(__FUNCTION__);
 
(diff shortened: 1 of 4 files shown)


stefanmajoor:
Why are some method/class names in Dutch?

Apart from that, the PR seems ok.

kokx:
Because the API that is being consumed is entirely in Dutch. And I basically made a service that consumes the API, and it is a good practice that local names reflect the API names to avoid confusion.

stefanmajoor:
Interesting, do you have a source on that explaining why?

Since the file basically is an adapter for the OaseClient, it seems for me the perfect place to change the dutch to english language. At this moment the naming of Oase leaks to our internal interfaces, which I think is bad.

kokx:
The source why the API is in dutch, is because it was created by Dienst ICT.....

kokx:
@stefanmajoor @jszanto Is this PR going to be merged sometime soon? Or are there changes needed?

stefanmajoor:
I still do not fully agree with the Dutch naming scheme. Apart from that the PR seems good.

@jszanto Could you give you opinion about the Dutch/English naming? If you agree with @kokx about the Dutch naming, you can merge this PR

jszanto:
I agree that it would be better to give them english names (and name them in camelCase). However, this PR was made to fix issue #174 and since it properly fixes #174 I think it should be merged. Perhaps create a seperate issue/PR for the naming @stefanmajoor?

Issue #252, opened 2015-11-22
Add board and organ ACL role


jszanto:
It also would be convenient if the board and organ role would automatically be applied and removed.

PR #292, opened 2015-12-08
ACL configuration

kokx:
Configures ACL to:

- [x] Allow users in TU/e range to view exams
- [x] Give board members the same rights as admins.
- [x] Make it possible to check if a user has access because of an organ relation.

This belongs with issues #252 and #253

### DIFF
--- a/module/User/Module.php
+++ b/module/User/Module.php
@@ -202,7 +202,7 @@ class Module
                      */
                     $acl->addRole(new Role('guest'));
                     $acl->addRole(new Role('tueguest'), 'guest');
-                    $acl->addRole(new Role('user'), 'guest');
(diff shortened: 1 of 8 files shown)


module/User/Module.php
```diff
@@ -168,6 +168,10 @@ public function getServiceConfig()
                     );
                 },
 
+                'user_remoteaddress' => function ($sm) {
+                    $remote = new \Zend\Http\PhpEnvironment\RemoteAddress();
+                    return $remote->getIpAddress();
+                },
```
jszanto: Zend framework's ability to add abstraction layers to every single php function or global variable still amazes me every time.

PR #299, opened 2015-12-14
Allow logging in using pincodes

jszanto:


module/User/src/User/Authentication/Adapter/PinMapper.php
```diff
@@ -0,0 +1,135 @@
...
+
+        /**
+         * Users logging in in this way should not have all their regular roles. Since this login
+         * method is less secure.
+         */
+        $userRole = new UserRoleModel();
+        $userRole->setRole('sosuser');
+        $userRole->setLidnr($this->lidnr);
+        $user->setRoles([$userRole]);
+
```
jszanto: Hier moet nog even naar gekeken worden, want dit is geen goede oplossing. Bijvoorbeeld door de IsBoardMember() assertion kan iemand nog steeds alle rechten krijgen.

kokx: Je kan een `deny` rule zetten op de `sosuser` role. Dit zal zeer waarschijnlijk boven de `IsBoardMember()` assertion gaan.

kokx:
How are you handling a timeout? When you log into the screen, you should be logged out automatically after some time.

Also, might it be better to implement this screen mostly in JavaScript? And use some kind of API to communicate with the website? You could actually resend login information (membership number and pincode) with every request that needs a login. This way you don't need complicated session management.

jszanto:
The plan was to implement the interface almost fully in javascript and then use the REST API to communicate with the website. Using the existing session/authentication setup simplifies things here, since that allows me to use the existing activity sign up and sign out methods. Timeouts can be handled by the client. E.g. logout after x seconds and logout after someone has subscribed to an activity.

Issue #296, opened 2015-12-09
Activity: Implement ACL


kokx:
Now that we can check for organs, the Activity module should implement ACL to check user rights.

PR #301, opened 2015-12-15
Activity ACL 

stefanmajoor:
Fixes issue #296 

@kokx can you confirm that this is the correct way to do the acl?

### DIFF
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -90,8 +90,6 @@ class Module
 
                     $acl->allow('user', 'activity', 'create');
 
-                    $acl->allow('admin', 'activity', 'manage');
-
                     return $acl;
                 },
             ]
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -88,6 +88,7 @@ class Module
                     $acl->allow('guest', 'activitySignup', 'signUp');
                     $acl->allow('guest', 'activitySignup', 'view');
 
+
                     $acl->allow('user', 'activity', 'create');
 
                     return $acl;
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -85,11 +85,12 @@ class Module
                     $acl->addResource('activitySignup');
 
                     $acl->allow('guest', 'activity', 'view');
-                    $acl->allow('guest', 'activitySignup', 'signUp');
+
                     $acl->allow('guest', 'activitySignup', 'view');
 
-
                     $acl->allow('user', 'activity', 'create');
+                    $acl->allow('user', 'activitySignup', 'signUp');
+                    $acl->allow('user', 'activitySignup', 'checkUSerSignedUp');
 
                     return $acl;
                 },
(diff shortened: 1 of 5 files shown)


module/Activity/Module.php
```diff
@@ -90,6 +90,8 @@ public function getServiceConfig()
 
                     $acl->allow('user', 'activity', 'create');
 
+                    $acl->allow('admin', 'activity', 'manage');
+
                     return $acl;
```
jszanto: This is not needed since admins have all permissions by default.

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -32,6 +32,17 @@ protected function getDefaultResourceId()
...
 
     /**
+     * Return the form for this activity
+     *
+     * @return ActivityForm
+     */
+    public function getForm()
+    {
+        $this->permissionCreateActivityOrException();
+        return new ActivityForm();
```
kokx: It is better to obtain the form from the Service Manager. Such that you can inject things like translation from there.

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -197,9 +222,54 @@ public function reset(ActivityModel $activity)
...
     }
+
+    /*
+     * Some general ACL functions
+     */
+
+    /**
+     * Checks if the user is allowed to view the activity, and if not, throws an used permission exception
+     */
+    public function permissionViewActivityOrException()
```
kokx: It would be a bit better to make these permissions a little more granular. Like having separate `approve` and `reset` privileges, instead of a single privilege.

jszanto:
There are still some problems in the controllers, like a bunch of `==='guest'`. I believe these checks can simply be removed since you are doing ACL checks in the service now.

module/Activity/src/Activity/Service/Activity.php
stefanmajoor: I could do that, however I think that permissions should not be too granular (It hurts the simplicity of the acl), the approve/disapprove/reset actions are actions are quite closely related I think.

Also what do you think of mu using other methods to do the permission check. In other Services, e.g. (https://github.com/GEWIS/gewisweb/blob/master/module/Photo/src/Photo/Service/Photo.php) this is not done, which imo both clutters the code inside the function, and is anything but DRY.

kokx: I don't think that we have to worry a lot about the ACL being too granular, because we configure it in one place. I do agree that approve and disapprove are closely related (it would actually be weird to split those), but I'm not so sure about reset there.

Anyway, I am not sure on my opinion on the second part. Indeed, it is true that including that check in the function itself isn't DRY. But neither is this solution (to be DRY here, you would have to combine everything into one function, but that would also clutter things in a different way).

stefanmajoor:
So, as @jszanto pointed out, there are some controllers that still have === 'guest'. They have to do with the activity signup. The activity signup should have ACL as well, however, this is a bit pointless to do before the activity signup is refactored just like the activity was done. Therefore I marked this PR as 'in progress', which will allow people to give some pointers if there are still things to be changed. Once I did the activity signup refactor, I will reopen this PR with the ACL done in the activity refactor as well.

stefanmajoor:
So I think I have fixed all your comments now

PR #303, opened 2015-12-15
Organ mutations

kokx:


stefanmajoor:
As I understand correctly, the data is fully truncated during the progress of this script, and the data is not available at the GEWIS website. Since the data is read only anyway, and this procedure does not allow this data to be refered to (since foreign keys might be violated), why not do it in a transaction?

kokx:
@stefanmajoor It is now using transactions.

kokx:
@jszanto @stefanmajoor Can this be merged, or are there things that prevent that?

jszanto:
If @stefanmajoor has no further concerns, then this can be merged.

PR #304, opened 2015-12-17
Poll XSS fix

kokx:


stefanmajoor:
Seems ok to me

But since XSS-injections are not really my thing, can somebody with more experience confirm and merge?

Issue #327, opened 2015-12-22
Activity: validate fieldset name and options of multiple languages


mstruijs:
Specifically, validate that the required fields in `Activity\Form\ActivityFieldFieldset` corresponding to each _selected_ language in the `Activity\Form\Activity` are filled in.

tuupke:
Cool, ik wist niet dat ik aan de website mee ontwikkelde ^^

PR #328, opened 2015-12-22
Issue297

mstruijs:
Fixes Issue #297, apart from some validation, which is postponed to #327.


kokx:
Please also fix the codeclimate issues that aren't only about style

PR #341, opened 2015-12-22
Comments for polls

kokx:


module/Frontpage/src/Frontpage/Form/PollComment.php
```diff
@@ -0,0 +1,77 @@
...
+                    ]
+                ]
+            ],
+            'content' => [
+                'required' => true,
+                'validators' => [
+                    [
+                        'name' => 'string_length',
+                        'options' => [
+                            'min' => 16
```
jszanto: This min length seems a bit high. There currently are many valid poll comments with less than 16 characters.

kokx: Fixed.

PR #348, opened 2015-12-23
Feature/design education index

armandduijn:
schermafbeelding 2015-12-23 om 01 21 29


public/scss/components/_layout.scss
```diff
@@ -16,3 +16,43 @@ body {
...
+    position: relative;
+
+    &:hover {
+        .glyphicon {
+            animation: 4s linear 0s dealWithIt;
+        }
+    }
+
+    &:after {
+        content: "DEAL WITH IT";
```
jszanto: Not sure if we can deal with it...

PR #357, opened 2015-12-23
Styled the administrator interface of the photo module a bit

gafderks:


module/Photo/src/Photo/Service/Admin.php
```diff
@@ -178,7 +178,8 @@ public function upload($files, $album)
             }
         } else {
             throw new \Exception(
-                $translator->translate('The uploaded file is not a valid image')
+                $translator->translate("The uploaded file is not a valid image \nError: " .
+                    json_encode($imageValidator->getMessages()))
```
jszanto: Does this work? If so, what does it do?

gafderks: I didn't get the uploader to work. This was added to get more debug information.

jszanto: I don't think `json_encode` returns a very readable error message. Perhaps something like `implode(',', array_values($imageValidator->getMessages()))` would be better?

mstruijs: The manual (http://framework.zend.com/manual/current/en/modules/zend.validator.html) suggests something like
```
foreach ($imageValidator->getMessages() as $messageId => $message) {
        $translator->translate("Validation failure '$messageId': $message\n");
    }
```

jszanto: Implode pretty much does the same, except for it results in a slightly more compact message. Keeping the message as small as possible is good in this case, since there is very little room in the view to display it.

kokx: Also, when using translation, use `sprintf. Like this:

```php
sprintf($translator->translate("The uploaded file is not a valid image \nError: %s"), 
    json_encode($imageValidator->getMessages()))
```

gafderks: I have fixed this issue.

jszanto:
`37906ea` added something which automatically places a container on all admin pages (since none of those have to be full width. So adding another container here is not needed.

gafderks:
@jszanto I have removed the containers.

PR #367, opened 2015-12-23
Logo's can now be uploaded in edit company form

### LINKED ISSUE #364: Bedrijven: uploading logo does not work when editing a company

wsinnema:
Also fixed deletion of logo when company is deleted. Fixes #364

### DIFF
--- a/module/Company/src/Company/Service/Company.php
+++ b/module/Company/src/Company/Service/Company.php
@@ -72,12 +72,28 @@ class Company extends AbstractACLService
      * @param mixed $company
      * @param mixed $data
      */
-    public function saveCompanyByData($company,$data)
+    public function saveCompanyByData($company, $data, $files)
     {
         $companyForm = $this->getCompanyForm();
         $companyForm->setData($data);
         if ($companyForm->isValid()){
             $company->exchangeArray($data);
+            foreach ($company->getTranslations() as $translation) {
+                $file = $files[$translation->getLanguage() . '_logo'];
+                try {
+                    $oldPath = $translation->getLogo();
+                    $newPath = $this->getFileStorageService()->storeUploadedFile($file);
+                    $translation->setLogo($newPath);
+                    echo $newPath;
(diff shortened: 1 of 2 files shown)


module/Company/src/Company/Service/Company.php
```diff
@@ -72,12 +72,28 @@ public function savePackageByData($package,$data)
...
+                    $newPath = $this->getFileStorageService()->storeUploadedFile($file);
+                    $translation->setLogo($newPath);
+                    echo $newPath;
+                    if ($oldPath != '' && oldPath != newPath) {
+                        $this->getFileStorageService()->removeFile($oldPath);
+                    }
+                    echo $translation->getLogo();
+                } catch (\Exception $exception) {
+                    var_dump($exception);
+
```
jszanto: There is still a bunch of debug code over here.

wsinnema:
Debug code is gone now

module/Company/src/Company/Service/Company.php
```diff
@@ -136,24 +148,15 @@ public function insertCompanyByData($data,$files)
...
-            //echo "hoi";
-            //var_dump($files);
             foreach ($company->getTranslations() as $translation) {
                 $file = $files[$translation->getLanguage() . '_logo'];
-                //var_dump($file);
                 try {
                     $newPath = $this->getFileStorageService()->storeUploadedFile($file);
-                    echo $newPath;
                     $translation->setLogo($newPath);
                 } catch (\Exception $exception) {
```
jszanto: Don't we need to do anything with these exceptions? Like return an error message?

wsinnema: Error message is not needed, because when the upload fails, the logo will not be set, and the next page will display an error, or the user will still see the old image there.

jszanto: Perhaps return false then such that the user can try to upload the logo again?

wsinnema: The user can try again on the next screen.

PR #370, opened 2015-12-23
Updated the favicon with the version that Koen has made.

gafderks:


jszanto:
What is the difference between this one and the original?

Mesoptier:
Hogere kwaliteit!

PR #371, opened 2015-12-23
Require activity costs to be specified.

### LINKED ISSUE #368: Activiteiten: Disallow empty cost
You need to specify costs: e.g. can be 'gratis' or a number.

mstruijs:
Fixes #368


stefanmajoor:
Please fix the Code Climate issue

Furthermore, I think that this approach is much simpeler and better than my previous approach. So, if you fixed the issue, I think it can be merged

PR #377, opened 2015-12-24
Fixes uploading a banner when editing a company and when editing a banner package

### LINKED ISSUE #365: Bedrijven: uploading banner does not work when editing a banner package

wsinnema:
Fixes #365 and #364

### DIFF
--- a/module/Company/src/Company/Service/Company.php
+++ b/module/Company/src/Company/Service/Company.php
@@ -97,7 +97,10 @@ class Company extends AbstractACLService
             $company->exchangeArray($data);
             foreach ($company->getTranslations() as $translation) {
                 $file = $files[$translation->getLanguage() . '_logo'];
-                if ($file['error'] === 0) {
+                if ($file['error'] !== 4) {
+                    if ($file['error'] !== 0) {
+                        return false;
+                    }
                     $oldPath = $translation->getLogo();
                     $newPath = $this->getFileStorageService()->storeUploadedFile($file);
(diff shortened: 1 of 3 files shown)


module/Company/src/Company/Service/Company.php
```diff
@@ -56,12 +56,25 @@ public function getHiddenCompanyList()
...
+                $package->exchangeArray($data);
+                $file = $files['banner'];
+                try {
+                    $oldPath = $package->getImage();
+                    $newPath = $this->getFileStorageService()->storeUploadedFile($file);
+                    $package->setImage($newPath);
+                    if ($oldPath != '' && oldPath != newPath) {
+                        $this->getFileStorageService()->removeFile($oldPath);
+                    }
+                } catch (\Exception $exception) {
```
kokx: Why an empty catch block? This is really bad practise. Also, catch statements that catch all exceptions are bad as well. Please try to catch less here.

jszanto: It might be better to fix this in #367 and then rebase your other branches once that PR is merged. Now we have this same issue in 3 of your PRs.

wsinnema: I need to catch an \Exception here. Sure, I could rewrite this part to check for the error condition too, but that would result in code duplication, which is also bad practise.

wsinnema: Everything is fixed now

PR #392, opened 2015-12-27
Fix signup

### LINKED ISSUE #372: Activity: Notice: undefined variable
When not logged in and viewing an activity that has a form, I get the following notice:
```
Notice
: Undefined variable: form in
C:\xampp\htdocs\gewisweb\module\Activity\src\Activity\Controller\ActivityController.php
on line
59
```

### LINKED ISSUE #373: Activity: Not added to table after subscription
When you subscribe to an activity, you do not show up in the table. After a refresh, you *are* in the table.

### LINKED ISSUE #374: Activity: success is not set on successful signoff
The success alert is never shown because success is not set

### LINKED ISSUE #375: Activity: CSRF vulnerability on activity signoff
You do not need to use the form to signoff for an activity. It should not be possible to signoff by only requesting an url like `/gewisweb/public/activity/signoff/9`.

mstruijs:
Fixes several bugs related to the signup process for activities:

- Define the undefined variables in the views (fixes #372 )
- Provide the proper ACL rights
- Make sure that the order of the persons that subscribed for the activity is displayed correctly.
- Fix some bugs in the controller (fixes #374 )
- Ensure that the subscription table is updated immediately (fixes #373 )
- Use a form for unsubscribing (fixes #375 )

### DIFF
--- a/module/Activity/src/Activity/Controller/ActivityController.php
+++ b/module/Activity/src/Activity/Controller/ActivityController.php
@@ -7,6 +7,7 @@ use Activity\Service\Signup;
 use Zend\Mvc\Controller\AbstractActionController;
 use Zend\Session\Container as SessionContainer;
 use Activity\Form\ActivitySignup as SignupForm;
+use Activity\Form\ModifyRequest as RequestForm;
 use Zend\View\Model\ViewModel;
 
 class ActivityController extends AbstractActionController
@@ -59,6 +60,7 @@ class ActivityController extends AbstractActionController
             'signedUp' => $signupService->getSignedUpUsers($translatedActivity),
             'signupData' => $translatorService->getTranslatedSignedUpData($activity, $session->lang),
             'form' => $form,
+            'signoffForm' => new RequestForm('activitysignoff','Unsubscribe'),
             'fields' => $fields
         ];
     }
@@ -105,7 +107,6 @@ class ActivityController extends AbstractActionController
 
         $translator = $activityService->getTranslator();
 
-//        $params = $this->viewAction();
         //Assure the form is used
         if (!$this->getRequest()->isPost()){
             $params = $this->viewAction();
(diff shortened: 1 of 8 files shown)


module/Activity/src/Activity/Service/Signup.php
```diff
@@ -87,10 +87,14 @@ public function getSignedUpData(ActivityModel $activity)
...
 
         $fieldValueMapper = $this->getServiceManager()->get('activity_mapper_activity_field_value');
+        $memberMapper = $this->getServiceManager()->get('decision_mapper_member');
         $result = [];
         foreach($activity->getSignUps() as $signup){
             $entry = [];
-            $entry['member'] = $signup->getUser()->getMember()->getFullName();
+            //$signup->getUser()->getMember() yields an empty result if the signup is new.
+            //(i.e. a signup has taken place in the same request)
+            $member = $memberMapper->findByLidnr($signup->getUser()->getLidnr());
```
mstruijs: Since `$signup->getUser()->getMember()` gets a doctrine proxy object (`DoctrineORMModule\Proxy\__CG__\Decision\Model\Member`), filled with null-values if the signup has been created in the same request, I have used a different way to get the name. I don't know enough about Doctrine to either tell if this a symptom of another bug or how to properly do this. This way does seem to work, though.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -7,6 +7,7 @@
 use Zend\Mvc\Controller\AbstractActionController;
 use Zend\Session\Container as SessionContainer;
 use Activity\Form\ActivitySignup as SignupForm;
+use Activity\Form\ModifyRequest as RequestForm;
```
jszanto: This form does not seem to exist? Also it is better to retrieve new form instances from the service using the service manager. (To allow easy passing of constructors for translation, etc.)

module/Activity/src/Activity/Service/Signup.php
jszanto: I can't quite produce any issues when I try replacing that by $signup->getUser()->getMember() again.

module/Activity/src/Activity/Controller/ActivityController.php
mstruijs: It seems I had forgotten to add it. Retrieving form instances from the service is a good idea in general, but does not benefit this particular form (no instances living in the service are required) and this has not been applied yet in the rest of the module, so I think that that can be deferred.

module/Activity/src/Activity/Service/Signup.php
mstruijs: I cannot reproduce the issues I got myself either. Which version should be kept? ('either one' is a valid answer)

jszanto: I think the previous version is better (and it saves a database query)

jszanto:
Also note that this branch has some merge conflicts.

PR #394, opened 2015-12-27
Fix activity signup screen layout

jszanto:
This is an attempt to make the activity screen work again after the layout got slightly messed up. The scrolling part still seems to only work at random times. I have no idea why this is. All I know is that it worked fine before PR #312 was merged. Note that my CSS is just a large collection of hacks, it probably should be redone sometime by someone else.


gafderks:
Maybe it's a good idea to use [jQuery Mobile](http://demos.jquerymobile.com/1.4.5/).

jszanto:
@gafderks jQuery Mobile could work to provide some more touch friendly UI elements, but it won't solve certain problems such as scrolling. Remember that the activity screen is not like a mobile device, it does not generate any touch events, all input is mouse input. It also would require more work to restyle everything again.

PR #399, opened 2015-12-28
Activity: Set a deadline for subscription for an activity

### LINKED ISSUE #192: Activity: Set a deadline for subscription for an activity
Allow the subscriptions to stop before the actual activity.

stefanmajoor:
Fixes #192

### DIFF
--- a/module/Activity/src/Activity/Controller/ActivityController.php
+++ b/module/Activity/src/Activity/Controller/ActivityController.php
@@ -5,9 +5,7 @@ namespace Activity\Controller;
 use Activity\Model\Activity;
 use Activity\Service\Signup;
 use Zend\Mvc\Controller\AbstractActionController;
-use Zend\Session\Container as SessionContainer;
 use Activity\Form\ActivitySignup as SignupForm;
-use Zend\View\Model\ViewModel;
 
 class ActivityController extends AbstractActionController
 {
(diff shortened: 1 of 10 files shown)


module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -176,13 +174,4 @@ public function signoffAction()
...
     }
-
-    public function touchAction()
-    {
-        $viewModel = new ViewModel();
-        $viewModel->setVariables(array('key' => 'value'))
-            ->setTerminal(true);
-
-        return $viewModel;
-    }
```
jszanto: Why was this removed?

jszanto:
It appears something went wrong when you merged the master in to this branch (which is a bad practice too).

stefanmajoor:
Something went horribly wrong when the changes from the master were applied. Let me have some time to figure it out, and I will make another pull request if I have fixed what went wrong.

Also git automatically merges if you pull. That is why a merge was doen.

jszanto:
The easiest would be to remove that last merge commit, fetch the master (instead of pulling) and then rebase upon the master,

stefanmajoor:
Except it did not go wrong in the merge, but rather stuff just disappeared in the first commit I made at that branch.

PR #410, opened 2015-12-30
Production configuration

jszanto:
This defines a production and development configuration and enables caching, classmap autoloading and other fancy things in production. Still a WIP (so don't merge this unless you want to break everything).


kokx:
Please fix the travis build as well. Or remove the PHPUnit tests (I don't think we really need those at the moment).

jszanto:
@kokx travis build has been fixed

config/application.config.php
```diff
@@ -1,22 +1,29 @@
...
-        'Company',
-        'Decision',
-        'Frontpage',
-        'Activity',
-        'Photo',
-        'Education',
-        'Migration',
-        'MaglMarkdown'
-    ],
+$env = getenv('APP_ENV') ?: 'production';
```
kokx: I think that it is bad to include such code in a config file. It would be better to distuinguish between both in the `index.php`.

jszanto: You mean we should define `$env` in `index.php` instead?

stefanmajoor: The configuration for the module should depend on the environment variables, and the other code (ideally) should not. There are several ways to do this. Either by setting environment related variables in .env files (or a mechanism that does the same, like the .local files). This is especially a good solution for passwords etc. Or different config files need to be loaded according to the environment. I.e. you have a common config file, and a config file for production, development, staging and testing. You then merge the common config file, with the environment specific file.

jszanto: The zf2 docs suggest to do it this way... http://framework.zend.com/manual/current/en/tutorials/config.advanced.html

Of course following the zf2 docs is not necaserely the right thing to do here. However the differences between the production and development config are very minimal, so I'm doubting if it is worth the effort to maintain different config files for each.

stefanmajoor: Well, the ZF2 tutorial made me sad.

-  It is a mess IMO
-  They use == instead of ===, so very weird things will happen if you set the environment to True.
- If you are using  ($env == 'production') three times, at least make an intermediate variable $isProduction = ($env === 'production'), which makes the code much more readable.

jszanto: The ZF2 tutorial indeed is a bit saddening now that I think about it.

jszanto:
I have now placed the production and development configs in to separate files. The development config inherits the production config since the two are almost identical.

stefanmajoor:
I personally think that this is a good solution for this problem.

jszanto:
In that case, feel free to merge this.

PR #411, opened 2015-12-31
Fix activity signup redirect

### LINKED ISSUE #393: Activity: Redirect activity/signup and activity/signoff to activity/view
Since the user should not visit the signup and signoff routes, but only POST to them, remaining on these routes gives unneeded error messages when reloading the page (e.g. when logging in or out). Note that the messages are printed in the signup and signoff routes, so this should be changed as well.

mstruijs:
The user is now redirected to the view route after (un)subscribing for an activity and messages (stored in the session) can be displayed in the view route.
Fixes #393.


module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -181,31 +191,48 @@ public function signoffAction()
...
+     * $success was false)
+     *
+     * @param int $id
+     * @param boolean $success Whether the request was successful
+     * @param string $message
+     */
+    protected function redirectActivityRequest($id, $success, $message)
+    {
+        $activityRequestSession = new SessionContainer('activityRequest');
+        $activityRequestSession->success = $success;
```
jszanto: Zend framework actually has a built in function for sending messages http://framework.zend.com/manual/current/en/modules/zend.mvc.plugins.html#zend-mvc-controller-plugins-flashmessenger

PR #424, opened 2016-01-05
Add testdata module as dependency

jszanto:
Including this module as a dev dependency makes it much easier to use.

### DIFF
--- a/config/application.config.php
+++ b/config/application.config.php
@@ -3,6 +3,7 @@ return [
     // This should be an array of module namespaces used in the application.
     'modules' => [
         'ZendDeveloperTools',
+        'TestData',
         'DoctrineModule',
         'DoctrineORMModule',
         'User',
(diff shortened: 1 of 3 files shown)


config/application.config.php
```diff
@@ -3,6 +3,7 @@
     // This should be an array of module namespaces used in the application.
     'modules' => [
         'ZendDeveloperTools',
+        'TestData',
```
kokx: Adding it in `application.config.php` is not such a good idea, since this will break the application when the module is not there. So it will break in non-dev environments/

jszanto: #410 adds something which only loads development related modules in development environments, so this won't be a problem.

kokx: Then maybe this should be added in #410 instead?

jszanto:
Added this to #410

PR #432, opened 2016-01-10
Disallowed users to view unapproved polls

### LINKED ISSUE #397: Polls: Unapproved polls can be viewed and commented on

kokx:
Fixes #397

### DIFF
--- a/module/Frontpage/src/Frontpage/Service/Poll.php
+++ b/module/Frontpage/src/Frontpage/Service/Poll.php
@@ -110,7 +110,7 @@ class Poll extends AbstractAclService
         }
 
         // check if poll is approved
-        if (null === $poll->getApprover()) {
+        if (is_null($poll->getApprover())) {
             return false;
         }
 
--- a/module/Frontpage/src/Frontpage/Service/Poll.php
+++ b/module/Frontpage/src/Frontpage/Service/Poll.php
@@ -25,10 +25,19 @@ class Poll extends AbstractAclService
      * Retrieves a poll by its id
      * @param int $pollId the id of the poll to retrieve
      * @return PollModel|null
+     *
+     * @throws \User\Permissions\NotAllowedException if the user isn't allowed to see unapproved polls
      */
     public function getPoll($pollId)
     {
-        return $this->getPollMapper()->findPollById($pollId);
+        $poll = $this->getPollMapper()->findPollById($pollId);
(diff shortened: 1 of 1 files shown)


jszanto:
This won't prevent users from voting on unapproved polls. There should also be a check in the getPollOption() function or canVote() function.

kokx:
Fixed.

module/Frontpage/src/Frontpage/Service/Poll.php
```diff
@@ -25,10 +25,19 @@ public function getNewestPoll()
...
      * @param int $pollId the id of the poll to retrieve
      * @return PollModel|null
+     *
+     * @throws \User\Permissions\NotAllowedException if the user isn't allowed to see unapproved polls
      */
     public function getPoll($pollId)
     {
-        return $this->getPollMapper()->findPollById($pollId);
+        $poll = $this->getPollMapper()->findPollById($pollId);
+        if ((null === $poll->getApprover()) && !$this->isAllowed('view_unapproved')) {
```
stefanmajoor: Please refrain from Yoda conditions.

Also === null breaks consistency with the is_null that is used in the rest of the code. is_null has another benefit, that it does not allow you to accidentally type == null

kokx: I know for a fact that `is_null` isn't used everywhere else in the code (since I have always been using `null ===`). I usually prefer language constructions over functions. Neither am I sure that it is good to use `is_null`.

And the fact that it's a Yoda condition kinda is an old habit that dies hard (I think it was a code standard in ZF1 if you used a comparison with `null`).

stefanmajoor: In the same file is_null is used consequently, that is what I meant. Personally I don't mind if $a === null and is_null($a) is used interchangeably, since both can be understood easily (Although a lot of people would mind). Yoda conditions I do mind.

module/Frontpage/src/Frontpage/Service/Poll.php
```diff
@@ -100,6 +109,11 @@ public function canVote($poll)
             return false;
         }
 
+        // check if poll is approved
+        if (null === $poll->getApprover()) {
```
jszanto: Might want to remove this yoda condition too..

PR #436, opened 2016-01-10
Organ selection for activity

stefanmajoor:
Nearly finished, there are however some nasty hacks still in place.

Anyone get any idea how to do this better?

### DIFF
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -44,7 +44,6 @@ class Module
             'invokables' => [
                 'activity_service_activity' => 'Activity\Service\Activity',
                 'activity_service_activityTranslator' => 'Activity\Service\ActivityTranslator',
-                //'activity_form_activity' => 'Activity\Form\Activity',
                 'activity_form_activityfield_fieldset' => 'Activity\Form\ActivityFieldFieldSet',
                 'activity_form_activity_signup' => 'Activity\Form\ActivitySignup'
             ],
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -55,18 +55,14 @@ class Module
                     return $sm->get('doctrine.entitymanager.orm_default');
                 },
                 'activity_form_activity' => function ($sm) {
-                    /** @var \Decision\Service\Member $userService */
+                    /** @var \Decision\Service\Member $memberService */
                     $memberService = $sm->get('decision_service_member');
                     /** @var \User\Model\User $identity */
                     $identity = $sm->get('user_role');
-
-                    // This is a necessary hack, since the user and corresponding member model is cached
-                    // I could not get the organ from the Member, because this resulted in all kinds of exceptions
-                    // This forces doctrine to fetch a clean object from the database
-                    $user = $memberService->findMemberByLidNr($identity->getLidnr());
+                    $organs = $memberService->getOrgans($identity->getMember());
 
                     $translator = $sm->get('translator');
-                    return new \Activity\Form\Activity($user, $translator);
+                    return new \Activity\Form\Activity($organs, $translator);
 
                 },
                 'activity_service_signup' => function ($sm) {
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -2,6 +2,8 @@
 namespace Activity;
 
 
+use User\Service\User;
+
 class Module
 {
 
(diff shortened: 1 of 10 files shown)


module/Activity/Module.php
```diff
@@ -52,12 +54,27 @@ public function getServiceConfig()
...
+                    /** @var \Decision\Service\Member $userService */
+                    $memberService = $sm->get('decision_service_member');
+                    /** @var \User\Model\User $identity */
+                    $identity = $sm->get('user_role');
+
+                    // This is a necessary hack, since the user and corresponding member model is cached
+                    // I could not get the organ from the Member, because this resulted in all kinds of exceptions
+                    // This forces doctrine to fetch a clean object from the database
+                    $user = $memberService->findMemberByLidNr($identity->getLidnr());
+
```
jszanto: Perhaps it is easier to just add a function (if it doesn't exist already) to the member mapper which returns all organs which a member is currently part of. This will save some queries, since you don't need the entire member object.

stefanmajoor: Is there any need to save some queries? I do not see any performance issues yet. To quote someone famous: "Premature optimization is the root of all evil." That is also why I disagree with caching the member object, since it probably does not have a lot of influence on server load/rendering time, and it has all sorts of complications (this piece of code is one of them)

jszanto: Saving queries is not needed, but I think retrieving the data you need directly from the database will simplify the code. Caching the member object also has other non-performance related benefits.

module/Activity/src/Activity/Form/Activity.php
```diff
@@ -16,13 +21,22 @@ class Activity extends Form implements InputFilterProviderInterface
...
 
+        // all the organs that the user belongs to in organId => name pairs
+        $organs = [0 => $translator->translate('No organ')];
+
+        // Find user that wants to create an activity
+        $user->getCurrentOrganInstallations()->forAll(function ($index, OrganMember $organMember) use (&$organs) {
+            $organ = $organMember->getOrgan();
+            $organs[$organ->getId()] = $organ->getName();
+        });
+
```
jszanto: This kinds of stuff would not be needed if we query the database directly for the data we need.

module/Activity/Module.php
stefanmajoor: The most simplest thing would be if I just could use the object methods to get to the needed information, as is (a) expected, and (b) is the way to do it in OOP and (c) minimizes dependencies.

There is already a method in the member service to get the organs a member is in I see now. Since I get the feeling that fixing the behaviour is a lot of work, I will use this method instead.

stefanmajoor:
Fixed the hacks, so can be reviewed now

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -170,15 +173,45 @@ public function createActivity(array $params, $dutch, $english)
...
+            // The organs that the user belongs to with the correct organId (either 0 or 1)
+            $organs = $memberService->getOrgans($member);
+
+            // An array only containing the organ that this member belongs to and with the correct id
+            $correctOrgan = array_filter($organs, function (Organ $organ) {
+                return $organ->getId();
+            });
+
+            // Check if the member belongs to the organ
+            if (count($organ) === 0) {
```
jszanto: What does this do? Isn't $organ equal to null? I don't quite understand this whole block of code does actually. Are you trying to determine if the current user is in an organ with id $organId?

jszanto:
What happens if the organ for an activity is null? Can

jszanto:
Oops, I mistook the close button for a cancel button.

module/Activity/Module.php
```diff
@@ -42,9 +44,9 @@ public function getServiceConfig()
             'invokables' => [
                 'activity_service_activity' => 'Activity\Service\Activity',
                 'activity_service_activityTranslator' => 'Activity\Service\ActivityTranslator',
-                'activity_form_activity' => 'Activity\Form\Activity',
-                'activity_form_activity_signup' => 'Activity\Form\ActivitySignup',
-                'activity_form_activityfield_fieldset' => 'Activity\Form\ActivityFieldFieldSet'
+                //'activity_form_activity' => 'Activity\Form\Activity',
```
kokx: Why not simply remove this line?

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -60,7 +60,7 @@ public function viewAction()
             'isSignedUp' => $identity !== 'guest' && $signupService->isSignedUp($translatedActivity, $identity->getMember()),
             'signedUp' => $signupService->getSignedUpUsers($translatedActivity),
             'signupData' => $translatorService->getTranslatedSignedUpData($activity, $langSession->lang),
-            'form' => $form,
+            'form' => $form ? $form : null,
```
kokx: I guess this needs an `isset()` or another actual condition to check?

module/Activity/src/Activity/Form/Activity.php
```diff
@@ -16,13 +21,26 @@ class Activity extends Form implements InputFilterProviderInterface
...
+     */
+    public function __construct(array $organs, Translator $translator)
     {
         parent::__construct('activity');
         $this->setAttribute('method', 'post');
         $this->setHydrator(new ClassMethodsHydrator(false))
             ->setObject(new \Activity\Model\Activity());
 
+        // all the organs that the user belongs to in organId => name pairs
+        $organOptions = [0 => $translator->translate('No organ')];
```
kokx: I think the '0 =>` is quite superfluous here?

stefanmajoor: It is superfluous from a parsing perspective, it is not superfluous from an understanding perspective. Since it is a map, and not an array, the key should be provided for understandability

stefanmajoor:
I cleaned up the code for @kokx comment, and fixed a minor bug that I can along, where the subscriptions where not shown if the subscription deadline had passed.

PR #442, opened 2016-01-13
Added API routes for the photo's

kokx:


stefanmajoor:
There are some weird things going on in the route definitions. We do now have two endpoints, one for admin, and one for the api, and they both refer to the AlbumAdmin controller. And the different functions that are necessary are divided seemingly randomly over the two controllers:

- Creating an album can only be done by the admin
- Editting an album can only be done by the admin
- Importing an album can only be done by the api
- Uploading an album can only be done by the api
-  Deleting an album can only be done by the api

What is the reason it is done this way?

As far as I understand, this is because all the api routes return json, and all the admin routes return json, and this makes the admin module use the api for the things that are implemented in the api.

Still, what I would do is split the controller up into two parts: an admin controller, and an api controller, to group the functionallity together (SRP). Further more, I would have a good look at all things the api is able to do, and which it is not able to do. It makes no sense that you need the admin route to create an album, but you can only delete it through the api. Lastly, I only see one function of the API properly documented.

jszanto:
Also, the `move_photo` routes is not an api route while the `move_album` route is. Both routes return JSON.

kokx:
@stefanmajoor The reason everything was done this way was simple. I wanted this to be done quickly, so AViCo could start testing against this. Thus, I took all the routes that returned a `JsonModel`, and turned them into API routes (without really looking at context). The fact that I missed the `move_photo` route was a mistake.

kokx:
What I'm going to do, is create a separate API controller, which does only one thing right now (return a list of photo's from an album). Such that AViCo can start using it.

kokx:
Closed in favor of PR #471

PR #445, opened 2016-01-19
Fixed possible redeclaration of moduleIsActive

### LINKED ISSUE #437: Fatal error: cannot redeclare moduleIsActive()
Rebased my branch with master, and suddenly got this error:

```
Fatal error: Cannot redeclare moduleIsActive() (previously declared in /var/www/html/gewisweb2/module/Application/view/layout/layout.phtml:2) in /var/www/html/gewisweb2/module/Application/view/layout/layout.phtml on line 2
````

Tried to update to some older revision (one I knew that was good), but got the same issue still, then did a clean clone of the gewisweb repository, and the error was still there, which makes me suspect that this is either a caching problem, or a dependency problem

Is there somebody else with the same problem?

### PR DESCRIPTION
**Pull request by kokx**
Fixes #437

### DIFF
--- a/module/Application/Module.php
+++ b/module/Application/Module.php
@@ -93,6 +93,12 @@ class Module
                     $helper = new \Application\View\Helper\ScriptUrl();
                     return $helper;
                 },
+                'moduleIsActive' => function ($sm) {
+                    $locator = $sm->getServiceLocator();
+                    $helper = new \Application\View\Helper\ModuleIsActive();
+                    $helper->setServiceLocator($locator);
+                    return $helper;
+                },
                 'fileUrl' => function ($sm) {
                     $locator = $sm->getServiceLocator();
                     $helper = new \Application\View\Helper\FileUrl();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by stefanmajoor** [comment]
Is ZF2 using some kind of templating engine? Because I think that this is a fix that masks the underlying problem: that there is a function defined in a view, which violates the principle that PHP and html should be seperated as much as possible.

If that is something which is necessary because we do not use a templating engine, then I suggest that this approached is used for now, and we move to a templating engine for the next version. 

But before this PR should be merged, at least the  moduleIsActive function needs cleaning up. There are two problems: first a Yoda condition. And second of all, the line:
```
 $viewModel = current($view->viewModel()->getCurrent()->getChildren());
```
Depends on the internal state of the iterator of the array. This means that this code can be (unintendently) affected by other code that runs on the array, which means that after a refactor of other code this code may work differently, which can cause nasty bugs.

jszanto:
The nice way to do this would be to make a view helper for those two functions so we can remove them from the view.

kokx:
That is definitely true. I'll see what I'm going to do there.

kokx:
Refactored the `moduleIsActive` function to a proper view helper.

Issue #238, opened 2015-11-19
Activity: Let organiser view subscription email adresses


mstruijs:
The organiser is the person or group(= committee) that has created the activity. This may be useful to do in a separate screen.

jszanto:
Fixed in #447

Issue #237, opened 2015-11-19
Activity: export list of subscriptions


mstruijs:
To some printable format. A more printable web page of the list would do.

mstruijs:
In particular, the additional options should be displayed, with the application of dinner lists ('eetlijsten')

jszanto:
Fixed in #447

PR #447, opened 2016-01-27
More details for organizing committee of an activity

stefanmajoor:
This pull requests adds two things:

- A link to view all email adresses (and the ability to mail all the subscribed people)
- An export to pdf function for an activity

See #237 and #238 

The only slight issue that I still have is that only an organizing committee can view this details, and not the organizer itself. This might be good (if the creator leaves the committee), or bad (if there is no activity behind it, you can not view the information of the activity you created). It can be fixed by adjusting the permissions. However, I am currently now aware of a way to give a permission if you have either created the activity, or are in the organizing activity.

### DIFF
--- a/module/Activity/Module.php
+++ b/module/Activity/Module.php
@@ -1,8 +1,7 @@
 <?php
 namespace Activity;
 
-
-use User\Service\User;
+use User\Permissions\Assertion\IsOrganMember;;
 
 class Module
 {
@@ -115,6 +114,7 @@ class Module
 
                     $acl->allow('user', 'activity', 'create');
                     $acl->allow('user', 'activitySignup', ['signup', 'signoff', 'checkUserSignedUp']);
+                    $acl->allow('user', 'activity', 'viewDetails', new IsOrganMember());
 
                     $acl->allow('sosuser', 'activitySignup', ['signup', 'signoff', 'checkUserSignedUp']);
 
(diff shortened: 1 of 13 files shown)


module/Activity/view/activity/activity/view.phtml
```diff
@@ -144,4 +144,9 @@ $this->headTitle($this->translate('Activities')); ?>
         echo $this->formSubmit($submit);
         echo $this->form()->closeTag();?>
     <?php endif; ?>
+
+    <?php if ($this->acl('activity_acl')->isAllowed('activity', 'create')): ?>
```
jszanto: This doesn't quite seem right. Since not everyone can export this list for all activities. I think it would be better to add a page to the admin interface which lists all activities which someone has access to. Adding such a page also is handy if we implement things such as editing later.

module/Activity/Module.php
```diff
@@ -115,6 +114,7 @@ public function getServiceConfig()
 
                     $acl->allow('user', 'activity', 'create');
                     $acl->allow('user', 'activitySignup', ['signup', 'signoff', 'checkUserSignedUp']);
+                    $acl->allow('user', 'activity', 'viewDetails', new IsOrganMember());
```
jszanto: I'm wondering if it wouldn't make more sense to just have an organ member role which is automatically applied to all organ members. However that isn't really part of this PR.

module/Activity/view/activity/activity/view.phtml
stefanmajoor: This is outside the scope for this pull requests, and is something for a new issue, that can be resolved later

PR #451, opened 2016-01-31
Allow polls to have anonymous votes

jszanto:
This is required in order to migrate polls from the previous GEWIS website.

### DIFF
--- a/module/Frontpage/src/Frontpage/Model/PollOption.php
+++ b/module/Frontpage/src/Frontpage/Model/PollOption.php
@@ -137,7 +137,7 @@ class PollOption implements ResourceInterface
      */
     public function getVotesCount()
     {
-        return $this->anonymousVotes + (is_null($this->votes) ? null : $this->votes->count());
+        return $this->anonymousVotes + (is_null($this->votes) ? 0 : $this->votes->count());
     }
 
     /**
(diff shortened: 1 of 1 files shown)


module/Frontpage/src/Frontpage/Model/PollOption.php
```diff
@@ -116,13 +123,21 @@ public function setEnglishText($englishText)
...
+
+    /**
      * Get the number of votes for this poll option.
      *
      * @return integer
      */
     public function getVotesCount()
     {
-        return is_null($this->votes) ? null : $this->votes->count();
+        return $this->anonymousVotes + (is_null($this->votes) ? null : $this->votes->count());
```
stefanmajoor: null should be 0

jszanto: Fixed.

kokx:
Isn't this solution way too hacky? Shouldn't we have separate `PollVote` instances for every vote?

jszanto:
@kokx `PollVote` instances must be bound to a user id to prevent duplicate voting. However there is no way we can do this when this data is missing.

kokx:
Well, in that case we don't really have any other option indeed. Thanks for the explanation!

PR #459, opened 2016-02-03
Improve design of the photo module

jszanto:
Should be good enough for an initial release now.
screenshot from 2016-02-03 15 15 50
screenshot from 2016-02-03 15 12 11
screenshot from 2016-02-03 12 12 50
screenshot from 2016-02-03 12 13 37


kokx:
<3

PR #479, opened 2016-02-04
Password change and reset

kokx:
Is not done completely, but may be reviewed already

### DIFF
--- a/module/User/src/User/Controller/UserController.php
+++ b/module/User/src/User/Controller/UserController.php
@@ -146,13 +146,6 @@ class UserController extends AbstractActionController
     }
 
     /**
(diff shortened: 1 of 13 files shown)


module/User/src/User/Controller/UserController.php
```diff
@@ -104,6 +104,55 @@ public function registerAction()
...
+        ]);
+    }
+
+    /**
+     * Action to activate reset password.
+     */
+    public function activateResetAction()
+    {
+    }
+
```
jszanto: What does this do?

kokx: Haha. Nothing :D. Let's remove this.

PR #516, opened 2016-02-07
Activity: various fixes

### LINKED ISSUE #455: Activity: Unapproved activities can be subscribed to

### LINKED ISSUE #484: Activity: Subscription deadline should not be required when subscriptions disabled
Currently, an activity requires a subscription deadline, even if subscriptions are not enabled.

### LINKED ISSUE #523: Activity: Check editing rights for organ instead of membership of organ in activity creation form
Since editing rights are needed to create an activity and being a member of an organ does not imply having rights (since those are granted after logging in)

mstruijs:
This PR fixes the following:

- After creating an activity, the organiser is informed of the succes. (fixes #455 )
- Specifying a deadline is no longer required, it defaults to the end date of the activity. (fixes #484 )
- The deadline date is only shown if an activity allows subscription.
- The dates are now properly translated (i.e. no "Sunday" in the dutch locale)
- Display the correct organs in the creation form (fixes #523 )


jszanto:
Wouldn't it be better to not show the subscription deadline when it is equal to the end time?

module/Activity/view/activity/activity/create.phtml
```diff
@@ -5,9 +5,10 @@ $this->headLink()
 $this->headScript()
     ->prependFile($this->basePath() . '/js/activity-create.js')
     ->prependFile($this->basePath() . '/js/jquery.datetimepicker.full.min.js');
-$form->prepare();
+if (!$success) {$form->prepare();}
```
jszanto: This probably could be formatted nicer.

module/Activity/view/activity/activity/create.phtml
```diff
@@ -5,9 +5,10 @@ $this->headLink()
...
 ?>
 <div class="container">
+    <?php if (!$success): ?>
```
stefanmajoor: This is a if else structure that now spans more than 200 lines of code. I think it would be better to split it into more than one template. Either having a succes and a non success template, or doing this:
```
if (a):
    loadTemplate('notsuccess')
else:
    loadTemplate('succes')
endif;
```

Otherwise the templates become very incomprehensible quite fast

mstruijs: Having two separate templates is indeed a cleaner solution.

Note that the success page is probably only temporary, since, after more organiser features are implemented (e.g. editing after submitting), redirecting to the organiser page on success makes more sense.

jszanto: In that case I think this is fine for now, since the if-else block will leave when we add the redirect.

mstruijs:
I'm going to refactor the Activity service a bit. I think putting all the queries for groups of activities in a separate service would be better.

mstruijs:
Someone should probably check if the refactor doesn't break anything.

module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -0,0 +1,195 @@
...
+        return $activity;
+    }
+
+    /**
+     * Get the activity with additional details
+     *
+     * @param $id
+     * @return ActivityModel
+     */
+    public function getActivityWithDetails($id)
```
mstruijs: This does the same as `getActivity()`, apart from an additional permission check. Why? Is this supposed to return more details (which?) in a later iteration?
@stefanmajoor

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -306,6 +131,43 @@ public function createActivity(array $params, $dutch, $english)
...
+     * Find the organ the activity belongs to, and see if the user has permission to create an activity
+     * for this organ.
+     *
+     * @param \User\model\User $user The user creating the activity
+     * @param int $organId The id of the organ associated with the activity
+     * @return Organ The organ associated with the activity, if the user is a member of that organ
+     * @throws \User\Permissions\NotAllowedException if the user is not a member of the organ specified
+     */
+    protected function findOrgan(\User\model\User $user, $organId)
+    {
```
jszanto: I believe this whole function could be replaced by something like:

```
$organ = null;
try {
    $organ = $organService->getOrgan($id);
} catch (\Exception $e) {
    // Organ not found
}

if (!$decisionService->canEditOrgan($organ)) {
    // User is not allowed to create activity for this organ
}

return $organ;
```

It would make it slightly better understandable in my opinion.

mstruijs: The function is also incorrect, in the sense that a permission exception is generated when there exists no organ with the given id. I don't see why an exception should be caught here, though. (i.e. I don't see how or why an incorrect id should be handled here)

jszanto: Handling an incorrect id is not necessary here, since there is no way it can be incorrect through normal use of the website.

PR #540, opened 2016-02-12
Added breadcrumbs to album and photo view.

### LINKED ISSUE #515: Show which album you are browsing on the photo view page

gafderks:
Fixes #515

### DIFF
--- a/module/Photo/view/photo/album/index.phtml
+++ b/module/Photo/view/photo/album/index.phtml
@@ -2,6 +2,42 @@
         ->prependFile($this->basePath() . '/js/masonry.pkgd.min.js')
         ->prependFile($this->basePath() . '/js/imagesloaded.pkgd.min.js');
 ?>
+<section class="section section-breadcrumb">
+    <div class="container">
+        <ol class="breadcrumb">
+            <li>
+                <a href="<?= $this->url('photo') ?>">
+                    <?= $this->translate('Photos') ?>
+                </a>
+            </li>
+            <?php
+            // first we need to get all parents
+            $crumbs = [];
+            if (!is_null($album->getParent())) {
+                $currentAlbum = $album;
+                while(!is_null($currentAlbum->getParent())) {
+                    // prepend parent to array
+                    array_unshift($crumbs, [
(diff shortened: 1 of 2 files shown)


kokx:
This makes the 'Back to album' button seem superfluous. See:

untitled

(made from https://test.gewis.nl/photo/view/55229 on this branch)

gafderks:
I think it is better to keep both since the 'Back to album' button is more salient and this improves the usability. It is not bad to have multiple ways in the UI to achieve the same action.

kokx:
It is bad if they are literally next to each other. If the button was in another place then it might indeed be a great idea.

kokx:
Either way, an opinion from @Mesoptier or @armandduijn might be nice here.

armandduijn:
I understand @kokx's point but I have to agree with @gafderks. It also makes the navigation more uniform. Some pages have breadcrumbs, others don't.

Mesoptier:
I think we can get rid of the button. The breadcrumbs provide more functionality anyway.

Is it possible to change "Current photo" to something more descriptive.

jszanto:
I think "current photo" can be omitted.

gafderks:
I think 'Current photo' cannot be omitted because it is standard to breadcrumbs that the current page is also included at the end. Although I would love to put a more descriptive text there, I cannot find one since photos do not have names and the time-stamp of the photo is also not that informative.

kokx:
I would then simply make it 'photo' instead of 'current photo'

mhverbakel:
I think that users would try to find a back button in the upper left corner. The breadcrumbs may provide the same functionality, but that is less intuitive than clicking a back button in the upper left corner.

armandduijn:
Lucid does it too naamloos

kokx:
But at lucid there is at least some content between the buttons. But in this implementation there is literally only some whitespace between them.

kokx:
Also, when I actually look at that lucid page, I don't really see any usability (just fluff). So I wouldn't really compare our photo page to that.

kokx:
@gafderks What are your current plans on this?

gafderks:
Well, personally, I'm happy with the way this functionality is implemented in this PR.

jszanto:
I think we should keep it this way and await user feedback

PR #564, opened 2016-02-25
Company slug names are now validated

### LINKED ISSUE #541: Company: Validate slugs on company creation
Currently you can create a company with an invalid slug, after which it is impossible to edit it because the company admin also uses slugs for editing companies (why?).

wsinnema:
They are checked for uniqueness and forbidden characters. Fixed #541

### DIFF
--- a/module/Company/src/Company/Form/EditCompany.php
+++ b/module/Company/src/Company/Form/EditCompany.php
@@ -219,58 +219,56 @@ class EditCompany extends Form
                 ),
             ],
             'filters' => [
-                ['name' => 'StripTags'],
-                ['name' => 'StringTrim'],
             ],
         ]);
 
-        $filter->add(array(
+        $filter->add([
             'name' => 'en_website',
             'required' => false,
-            'filters' => array(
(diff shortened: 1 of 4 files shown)


module/Company/src/Company/Form/EditCompany.php
```diff
@@ -205,6 +205,24 @@ protected function initFilters()
...
+                    'fields' => 'slugName',
+                ]),
+                new \Zend\Validator\Regex([
+                        'pattern' => '/^[0-9a-zA-Z_\.\-]*$/',
+                    ]
+                ),
+            ],
+            'filters' => [
+                ['name' => 'StripTags'],
+                ['name' => 'StringTrim'],
```
jszanto: Why are these filters needed when you're already validating the slug properly?

wsinnema: No, they are not needed

wsinnema:
There eas still a code climate issue

jszanto:
Oops, I missed that

Issue #468, opened 2016-02-03
Company: Allow jobs to have dates


jszanto:

wsinnema:
Needs discussion

PR #565, opened 2016-02-25
Added dates to jobs

wsinnema:
Jobs now have dates, however, they are not shown yet on the public website. Possibly, we should first ask people if these dates have to be visible on the public site. However, technically, this fixes issue #468.

### DIFF
--- a/module/Company/src/Company/Model/Job.php
+++ b/module/Company/src/Company/Model/Job.php
@@ -77,6 +77,20 @@ class Job
     protected $description;
 
     /**
+     * The job's description.
+     *
+     * @ORM\Column(type="text")
+     */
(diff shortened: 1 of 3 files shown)


module/Company/src/Company/Model/Job.php
```diff
@@ -77,6 +77,20 @@ class Job
     protected $description;
 
     /**
+     * The job's description.
+     *
+     * @ORM\Column(type="text")
+     */
+    protected $description;
```
jszanto: Jobs have two descriptions now?

wsinnema: This is fixed now

PR #567, opened 2016-02-26
Randomized the jobs on the joblist

### LINKED ISSUE #566: Company: Randomize job order on joblist

wsinnema:
Fixes #566

### DIFF
--- a/module/Company/view/company/company/job-list.phtml
+++ b/module/Company/view/company/company/job-list.phtml
@@ -59,6 +59,8 @@ $this->headTitle($this->translate('Jobs'));
         <?php
 
 
+        // Randomize the order of companies
+        shuffle($jobList);
         foreach ($jobList as $job):
             if ($job->isActive()):
                 $company = $job->getPackage()->getCompany(); ?>
--- a/module/Company/view/company/company/job-list.phtml
+++ b/module/Company/view/company/company/job-list.phtml
@@ -57,15 +57,14 @@ $this->headTitle($this->translate('Jobs'));
         </div>
 
         <?php
-
-
-        // Randomize the order of companies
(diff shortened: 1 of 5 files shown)


module/Company/view/company/company/job-list.phtml
```diff
@@ -59,6 +59,8 @@ $this->headTitle($this->translate('Jobs'));
         <?php
 
 
+        // Randomize the order of companies
+        shuffle($jobList);
```
jszanto: Doing this, as I mentioned earlier, also randomizes the order on the page which shows all jobs for a company. This is very anoying in my opinion when trying to browse through all the jobs for a specific company.

wsinnema: I didn't notice, now it is fixed

module/Company/view/company/company/job-list.phtml
```diff
@@ -59,6 +59,10 @@ $this->headTitle($this->translate('Jobs'));
         <?php
 
 
+        // Randomize the order of companies
+        if ($randomize):
+            shuffle($jobList);
+        endif
```
kokx: Within view code, we have two styles of using control statements like `if`:

```php
<?php if (/* condition*/): ?>
HTML
<?php endif ?>
```

And:

```php
<?php
if (/* condition */) {
    /* PHP code */
}
```

Mixing them only makes code less readable.

wsinnema: It's fixed now

kokx:
Why is this closed? Shouldn't it be merged?

wsinnema:
Because the request listed commits that are already merged. I opened the same request again with the correct commits

PR #586, opened 2016-04-12
Fix the modulesisactive view helper

jszanto:


module/Application/src/Application/View/Helper/ModuleIsActive.php
```diff
@@ -23,6 +23,9 @@ public function __invoke($condition)
     {
         $info = $this->getRouteInfo();
         foreach ($condition as $key => $cond) {
+            if (!isset($info[$key])) {
+                return false;
+            }
             if (null !== $cond && $info[$key] != $cond) {
```
kokx: Why not just add this to the condition of the if statement here?

jszanto: Fixed

module/Application/src/Application/View/Helper/ModuleIsActive.php
```diff
@@ -23,7 +23,7 @@ public function __invoke($condition)
     {
         $info = $this->getRouteInfo();
         foreach ($condition as $key => $cond) {
-            if (null !== $cond && $info[$key] != $cond) {
+            if (!isset($info[$key]) || (null !== $cond && $info[$key] != $cond)) {
```
wsinnema: Please change this yoda condition

PR #588, opened 2016-04-13
Adds a sidebar to company admin interface

### LINKED ISSUE #98: Company: Melding bij verlopen packet

wsinnema:
The sidebar will show packages that will start or expire in the next 7 days
Fixes #98

### DIFF
--- a/module/Company/Module.php
+++ b/module/Company/Module.php
@@ -107,6 +107,7 @@ class Module
                 },
                 'company_admin_edit_job_form' => function ($sm) {
                     $form = new \Company\Form\EditJob(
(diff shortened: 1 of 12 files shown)


jszanto:
I think it might be handier to just add this at the top of the page instead of the sidebar. And also maybe only display it whenever there are packages which are expiring in the next seven days. In this way it won't take up that much empty space on the page

wsinnema:
I would prefer a sidebar, because when there are many changes, you'd have to scroll down to find the table. With a sidebar, you can see both at the same time.

jszanto:
As far as I know there will pretty much never be more than 1 or 2 companies in there

PR #590, opened 2016-04-14
A few important security enhancements

jszanto:
Just a few quick fixes to make login rate limiting and persistent sessions work properly.


kokx:
Right now, I can't completely review this as in-depth as I'd like. Going to do that later this week though. Please don't merge before then.

PR #595, opened 2016-04-15
Send an email on activity creation

stefanmajoor:
Send an email to a predetermined email when an activity is created.

### DIFF
--- a/module/Activity/src/Activity/Service/Activity.php
+++ b/module/Activity/src/Activity/Service/Activity.php
@@ -116,6 +116,8 @@ class Activity extends AbstractAclService implements ServiceManagerAwareInterfac
         $em->persist($activity);
         $em->flush();
 
+        $this->getEmailService()->sendActivityCreationEmail($activity);
+
         return $activity;
     }
 
@@ -313,5 +315,15 @@ class Activity extends AbstractAclService implements ServiceManagerAwareInterfac
     public function getActivityMapper()
     {
         return $this->sm->get('activity_mapper_activity');
+    }
+
+        /**
+     * Get the email service.
+     *
+     * @return EmailService
+     */
+    public function getEmailService()
+    {
+        return $this->sm->get('activity_service_email');
     }
 }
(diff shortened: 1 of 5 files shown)


module/Activity/src/Activity/Service/Email.php
```diff
@@ -0,0 +1,101 @@
...
+
+class Email extends AbstractService
+{
+
+    /**
+     * Send registration email.
+     *
+     * @param NewUserModel $newUser
+     * @param MemberModel $member
+     */
```
jszanto: a registration email?

module/Activity/view/email/activity.phtml
```diff
@@ -0,0 +1,37 @@
+<?php $lang = 'en'; ?>
+<?= $this->translate('Hello,')?><br /><br />
```
jszanto: Won't this result in emails being sent in different languages depending on what language the creator of the activity is using?

stefanmajoor: It indeed would. Can I just make the email in English/Dutch? What has preferences?

jszanto: I think it is fine to just have the email in one language. Which of the two shouldn't really matter, however the person reading it most likely will be dutch.

kokx: However, everyone that is dutch can read english emails. Not the other way around.

We can also put 'English text below' at the beginning of the email. And put an english text below the dutch text.

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -314,4 +316,14 @@ public function getActivityMapper()
     {
         return $this->sm->get('activity_mapper_activity');
     }
+
+        /**
+     * Get the email service.
```
kokx: Seems like some tab/spaces mismatch here.

jszanto:
Merging this now and then I will fix the tab/spaces issue after that.

Issue #594, opened 2016-04-15
Activity: Allow activity updates


mstruijs:
This can be implemented by storing activity proposals, which (after admin approval) can generate a new activity or update an existing one. The latter should have some restrictions related to consistency with existing signups.

jszanto:
I would suggest keeping it simple and only allowing name, description, time, location and costs updates. In this way the existing activity can just be updated without losing the signups.

Issue #510, opened 2016-02-07
Show activities in a compact list view in admin


jszanto:

jszanto:
Has been superseded by #642

PR #596, opened 2016-04-17
Feature/activity updates

mstruijs:
Implements the possiblity to request an update to the basic values of an activity (Issue #594 ). Updating/adding signup fields is not supported (yet?) 
Since there is no 'organiser' role yet, this feature is currently only allowed for admins.

Adds forms to all activity admin actions. (instead of via GET)
Adds a simple view of the description in the admin activity view (Issue #510 ) 
Dissallows subscription to unapproved activities (Issue #455 )


mstruijs:
I have considered the suggestions by Code Climate. I won't do anything with them. This PR is ready for review.

module/Activity/view/activity/admin/queue.phtml
```diff
@@ -1,3 +1,12 @@
+<h1><?= $this->translate('Update Proposals');?> </h1>
+<?= $this->partial('activity/activity/list.phtml', [
+    'activities' => $updatedActivities,
+    'proposals' => $updateProposals,
+    'admin' => true,
+    'MoreLink' => $moreUpdateProposals ? 'This won\'t happen. Really. If it does, just fix these proposals first.': null,
```
jszanto: Can't we just make it null always then?

mstruijs: Sure. I was just using the template of the pagination and added this. Problem is that actually supporting this means another view has to be made, which I rather don't if it will practically never be used.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -56,7 +56,7 @@ public function viewAction()
         $subscriptionDeadLinePassed = $activity->getSubscriptionDeadline() < new \DateTime();
         $result = [
             'activity' => $translatedActivity,
-            'signupOpen' => $activity->getCanSignUp() && !$subscriptionDeadLinePassed,
+            'signupOpen' => $activity->getCanSignUp() && !$subscriptionDeadLinePassed && $activity->getStatus()!==Activity::STATUS_APPROVED,
```
kokx: This line is quite long, also missing proper spacing at some places.

module/Activity/src/Activity/Form/ActivityFieldFieldset.php
```diff
@@ -5,18 +5,20 @@
...
 {
-    public function __construct() {
+    public function __construct(ObjectManager $objectManager) {
 
         parent::__construct('activityfield');
 
-        $this->setHydrator(new ClassMethodsHydrator(false))
-             ->setObject(new ActivityField());
-
+        $this->setHydrator(new DoctrineHydrator($objectManager));
```
kokx: This is not the way to inject a Hydrator. See the `Education` module for a proper example on this.

PR #633, opened 2016-09-12
Add the ability to store sessions directly in memcached

jszanto:
This replaces the default zf2 session storage (using php sessions)  by memcached session storage. This should prevent conflicts with other php applications running on the same server. What's left to figure out is how session cleaning should work. Also the sws compatibility layer will need a few changes.

### DIFF
--- a/config/autoload/global.php
+++ b/config/autoload/global.php
@@ -87,6 +87,14 @@ return [
     'sslcapath' => '/etc/ssl/certs',
 
     /**
+     * Session configuration
+     */
+    'session_config' => [
+        'use_cookies' => true,
+        'hash_function' => 'sha1'
+    ],
+
+    /**
      * Photo's upload directory configuration
      */
     'photo' => [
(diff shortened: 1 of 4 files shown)


kokx:
Note that PHP's default session storage on the server already is memcached session storage. Hence, I'm not sure that you're done with conflicts in this way.

Also, I'm going to have to review this in-depth sometime. Unfortunately I cannot reset github reminders to unread. Hence, when you read this, please mention me @jszanto to remind me to actually review this in-depth.

jszanto:
@kokx PHP's session storage prefixes all memcached keys with `memc.sess.key.` and PHP won't touch any keys other than those prefixed with that. This implementation prefixes all keys with `zfcache:` and therefore there shouldn't be any conflicts. The session id is also no longer stored in the `PHPSESSID` cookie. I just realize however that we might be able to achieve similar behavior by using fpm and setting the session key prefix per pool.

kokx:
@jszanto Then I think it's better to do that via a PHP pool indeed.

jszanto:
Closing this for now after succesfully testing the fpm pool approach

Issue #641, opened 2016-09-27
Page to view past activities


jszanto:
A page to view past activities preferably categorized by year

jszanto:
Fixed in #755

PR #650, opened 2016-10-05
Activity Archive (#641)

stefanmajoor:
Copied and modified the code from the photo page to show an archive of activities.

### DIFF
--- a/module/Activity/config/module.config.php
+++ b/module/Activity/config/module.config.php
@@ -77,7 +77,7 @@ return [
                             ]
                         ]
                     ],
-                    // Route for categorizing albums by association year.
+                    // Route for categorizing activities by association year.
                     'year' => [
                         'type' => 'Segment',
                         'options' => [
--- a/module/Activity/config/module.config.php
+++ b/module/Activity/config/module.config.php
@@ -68,6 +68,28 @@ return [
                             ]
                         ]
                     ],
+                    'archive' => [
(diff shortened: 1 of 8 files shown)


module/Activity/config/module.config.php
```diff
@@ -68,6 +68,28 @@
...
+                    'archive' => [
+                        'type' => 'Literal',
+                        'options' => [
+                            'route' => '/archive',
+                            'defaults' => [
+                                'action' => 'archive'
+                            ]
+                        ]
+                    ],
+                    // Route for categorizing albums by association year.
```
jszanto: Albums? Could you fix this comment or remove it to prevent confusion in the future.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -246,4 +246,41 @@ public function touchAction()
...
+
+
+    /**
+     * Display all the finished activities in a school year
+     *
+     * @param null $year
+     * @return ViewModel
+     */
+    public function archiveAction() {
+        //add any other special behavior which is required for the main photo page here later
```
jszanto: Same here with the comment

jszanto:
@stefanmajoor can you refactor some of this to use the new AssocationYear class which @gafderks made?

jszanto:
Re-opened as #755

PR #652, opened 2016-10-07
Added years to album breadcrumbs.

### LINKED ISSUE #578: Add years to photo breadcrumbs

gafderks:
Fixes #578

### DIFF
--- a/module/Decision/src/Decision/Model/AssociationYear.php
+++ b/module/Decision/src/Decision/Model/AssociationYear.php
@@ -1,0 +1,91 @@
+<?php
+
+
+namespace Decision\Model;
+
+
+class AssociationYear
+{
+    /**
+     * A GEWIS association year starts 01-07
+     */
+    const ASSOCIATION_YEAR_START_MONTH = 7;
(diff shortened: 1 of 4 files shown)


stefanmajoor:
@jszanto Since this code is already duplicated, both in my PR as here, maybe it is time to create a library or something for this. E.g. just a app that regulates everything w.r.t. study years. Especially since this may be needed in the future too.

jszanto:
@stefanmajoor Good idea, perhaps something like this?

```
class AssociationYear {

	public static function fromYear($year); // Returns instance of AssociationYear
	public static function fromDate($dateTime); // Returns instance of AssociationYear

	public function getYear(); // Returns 2015
	public function getYearString(); // Returns 2015-2016
	public function getStartDate(); // Returns DateTime('1-7-2016)
	public function getEndDate(); // Returns DateTime('30-6-2016)

}
```

gafderks:
@jszanto Where would I locate such a class? As a service in the Decision module?

jszanto:
@gafderks I think a model that isn't persisted in the database would be the best option. Since it doesn't really expose any services.

PR #679, opened 2016-12-14
Prevent deleting already accepted polls, just hide them instead.

### LINKED ISSUE #643: Prevent excessive poll deletion
Currently polls are deleted too often which destroys the poll history. This issue may be resolved by some better UX or by introducing soft deletes.

Staartvin:
Instead of deleting an approved poll from the database, it will now set the expiry date to 'now' and hide the poll from the homepage. 

Fixes #643


jszanto:
Just realized a slight issue with this, for expired polls, pressing the delete button will now just change the date. This makes the UI a bit confusing and we don't want information of expired polls to change.

PR #680, opened 2016-12-18
Add API method for getting member by lidnr

Mesoptier:
This PR will allow API users to get information for a member using the following URL: `/api/member/lidnr/1234`.

---

N.B. This PR was a request for the GEPWNAGE LAN website so members can sign up using their lidnr.


Mesoptier:
@jszanto I made the requested changes. I also added a field for the member's birth date and updated the date fields to use ISO8601 (only for the API method).

PR #686, opened 2017-01-04
Added links to fraternity and committee page

### LINKED ISSUE #635: Clickable names of committees / fraternities on pages of members

gafderks:
Fixes #635


jszanto:
Wouldn't it be better to just modify this code https://github.com/GEWIS/gewisweb/blob/master/module/Decision/src/Decision/Service/Member.php#L53 so it includes the organ instead of only the abbreviation?

gafderks:
@jszanto How do you suggest I change the array like `Array ( [BAC] => Array ( [0] => Voorzitter ) )` to include the organ?

jszanto:
not a 100% sure if this will work, but you could try `organ => functions`?

gafderks:
So use an object as a key in the array?

jszanto:
Yes

gafderks:
From the [docs](https://secure.php.net/manual/en/language.types.array.php):
> Arrays and objects can not be used as keys. Doing so will result in a warning: Illegal offset type.

jszanto:
Okay, how about
```
BAC => ['organ' => object, 'functions' => functions],..
```

module/Decision/src/Decision/View/Helper/Organ.php
```diff
@@ -0,0 +1,25 @@
+<?php
```
jszanto: Is this file still needed?

kokx: I can't find any occurence of the `findOrgan` in the code. So I would guess not.

module/Decision/view/decision/member/view.phtml
```diff
@@ -44,8 +44,15 @@ use Decision\Model\Address;
...
-                    <li><?= $organ ?><?php if (!empty($functions)): ?> (<?= implode(', ', $functions) ?>)<?php endif ?></li>
+                <?php foreach ($memberships as $abbr => $install): ?>
+                    <li><a href="<?= $this->url('home/organ', ['type' => $install['organ']->getType(), 'abbr' =>
+                            $abbr ]) ?>"><?= $abbr ?></a>
+                        <?php $functions = $install['functions']; ?>
+                        <?php if (!empty($functions)): ?> (<?=
+                            $functions ?>)
+                        <?php
+                        endif
+                        ?></li>
```
kokx: These three lines should be on one line.

module/Decision/view/decision/organ/show.phtml
```diff
@@ -8,7 +8,10 @@ $this->headTitle($organ->getName()); ?>
             <h2><?= $this->translate('Current members') ?></h2>
             <ul>
                 <?php foreach ($currentMembers as $membership): ?>
-                    <li><?= $membership['member']->getFullName() ?> <?php if (!empty($membership['functions'])): ?> (<?= implode(', ', $membership['functions']) ?>)<?php endif ?></li>
+                    <li><a href="<?= $this->url('member/view', ['lidnr' => $membership['member']->getLidnr()])
+                        ?>"><?= $membership['member']->getFullName() ?></a> <?php if (!empty
+                        ($membership['functions'])):
+                            ?> (<?= implode(', ', $membership['functions']) ?>)<?php endif ?></li>
```
kokx: The indendation here looks off.

module/Decision/src/Decision/View/Helper/Organ.php
gafderks: I have addressed these issues.

module/Decision/config/module.config.php
```diff
@@ -329,5 +329,10 @@
                 ]
             ]
         ]
-    ]
+    ],
+    'view_helpers' => [
+        'invokables' => [
+            'findOrgan' => 'Decision\View\Helper\Organ'
+        ],
+    ],
```
jszanto: I think this should be removed too

gafderks: True

Issue #599, opened 2016-06-03
Empty options can't be deleted when creating a poll


jszanto:
When creating a poll empty options should probably just be discarded.

PR #687, opened 2017-01-04
Allow deleting poll options #599

jszanto:


module/Frontpage/view/frontpage/poll/request.phtml
```diff
@@ -100,11 +101,11 @@ ob_start();
                     <?= $this->formElementErrors($element) ?>
                 </div>
             </div>
-    <?php endforeach ?>
+    <?php $i++; endforeach ?>
```
kokx: This looks a little strange to me. But don't really know a quick way to fix it.

jszanto: I could put the `$i++` on a seperate line if your eyes like that better.

kokx: I guess they would.

jszanto:
@kokx moved it

PR #693, opened 2017-01-05
Member: Moved addresses into left column and put photos in a masonry

### LINKED ISSUE #506: Photos on member page move addresses down too much (admin only)
On the member page, the addresses of a member are moved down too much.

See for example https://gewis.nl/member/5121

gafderks:
Fixes #506 possibly. I was unable to test the masonry as I have no photos in my local website and I cannot upload new photos as ImageMagick does not work on my system. I anyone can test the masonry, that would be nice.


gafderks:
Closed in favor of #694

Issue #498, opened 2016-02-06
Allow users to stay logged in


gafderks:

kokx:
As said in #524, if we implement this, we will need to make sure that sessions are refreshed sometimes. To refresh member information like membership of organs.

jszanto:
Fixed in #696

PR #696, opened 2017-01-05
Persist session in the database (#498)

jszanto:


module/User/src/User/Authentication/Storage/Session.php
```diff
@@ -2,25 +2,166 @@
...
         }
     }
 
     public function forgetMe()
     {
-        $this->session->getManager()->forgetMe();
+        $this->rememberMe = false;
+    }
+
+    public function isEmpty()
```
kokx: Undocumented public methods don't seem like a good idea to me.

module/User/src/User/Authentication/Storage/Session.php
```diff
@@ -2,25 +2,166 @@
...
+        $session = new SessionModel();
+        $session->setIp($this->sm->get('user_remoteaddress'));
+        $user = $this->sm->get('user_service_user')->detachUser($user);
+        $session->setUser($user);
+        $session->setSecret($this->generateSecret());
+        $session->setCreatedAt(new \DateTime());
+        $session->setLastActive(new \DateTime());
+        $mapper->persist($session);
+        $mapper->flush();
+        $this->saveCookie($session);
```
kokx: Maybe include some whitespace here and there for readability? This kinda looks incoherent and just some random lines of code thrown together.

module/User/src/User/Mapper/Session.php
```diff
@@ -63,6 +75,45 @@ public function persist($session)
...
+    /**
+     * Re-attach a user to the entity manager.
+     *
+     * @param entity $entity
+     *
+     * @return entity
+     */
+    public function merge($entity)
+    {
+        return $this->em->merge($entity);
```
kokx: I know that merge has some side-effects to the database. In the sense that I thought it might push local changes back to the database in some cases. Maybe we should insure this shouldn't happen?

jszanto: I actually don't think I'm using merge anymore, so this code can probably be removed safely.

kokx: Then check and remove, of course.

jszanto:
@kokx all fixed now I think

PR #704, opened 2017-01-06
Feature/activity updates by creator

### LINKED ISSUE #594: Activity: Allow activity updates

mstruijs:
Allows users to view and update the activities they created or the organ they are member of. 
This functionality is merged with approval functionality for the admin, which replaces a part of the previous interface.
Resolves #594 and #655


module/Activity/src/Activity/Controller/AdminApprovalController.php
```diff
@@ -0,0 +1,227 @@
...
+
+        if (is_null($proposal)) {
+            return $this->notFoundAction();
+        }
+
+        return [
+            'proposal' => $proposal,
+            'proposalApplyForm' => new RequestForm('proposalApply', 'Apply update'),
+            'proposalRevokeForm' => new RequestForm('proposalRevoke', 'Revoke update')
+            ];
```
jszanto: This formatting doesn't seem quite right.

mstruijs: Curiously, codeclimate does not detect this one. Perhaps because this add is a result of a file rename.

Issue #692, opened 2017-01-05
Activity: Allow editing before admin approval


mstruijs:
This will improve the current workflow of 'see activity'->'spot typo'->'create update-proposal'->'approve update' to 'see activity'->'spot typo'->'edit and approve activity'.

jszanto:
Fixed in #713

PR #713, opened 2017-01-07
A few UI/UX improvements to the activity admin

jszanto:
- Made the activity update view equal to the create view
- Added icons instead of numbers to indicate the approval status of an activity
- Added fancier buttons
- Updates by admins are now automatically approved
- Updates for non-approved activities are now automatically approved


module/Activity/src/Activity/Service/Activity.php
```diff
@@ -132,7 +133,7 @@ public function createUpdateProposal(ActivityModel $oldActivity, array $params,
             $em->remove($oldUpdate);
             $em->flush();
 
-            return $oldProposal;
+            return false;
```
mstruijs: If there exists a previous proposal, this early return means that the current proposal cannot be applied immediately. However, an admin should be able to do this, as a result of this PR.

module/Activity/src/Activity/Service/Activity.php
```diff
@@ -141,7 +142,32 @@ public function createUpdateProposal(ActivityModel $oldActivity, array $params,
...
+     * @param ActivityModel $activity
+     * @return bool indicating whether the update may be applied
+     */
+    protected function canApplyUpdateProposal(ActivityModel $activity)
+    {
+        if (!$this->isAllowed('update', $activity)) {
+            return false;
+        }
+
+        if ($this->isAllowed('approve', 'model')) {
```
mstruijs: This ACL resource and action appear to be undefined. Although this works, I think it would be preferable to use a resource that is actually defined somewhere and that makes sense. Here, `$this->isAllowed('update','activity')` would work and make sense IMO.

module/Activity/view/activity/admin/list.phtml
```diff
@@ -17,6 +17,7 @@
         <th><?= $this->translate('Organ') ?></th>
         <th><?= $this->translate('Submitter') ?></th>
         <th><?= $this->translate('Update status') ?></th>
+        <th><?= $this->translate('Approved') ?></th>
```
mstruijs: There is no reason to display the approval status in an admin list, where this value is already shown by the table in which it appears (note that this table is not used for old activities)

module/Activity/src/Activity/Service/Activity.php
jszanto: I think this only would be an issue if an activities status would change from accepted to rejected, which rarely (never?) occurs.

mstruijs: If an admin edits an approved activity without pending updates, this will return false, so it is not possible to immediately edit and approve that edit, which I think is intended by this PR.

mstruijs:
This PR resolves #692, btw.

Issue #717, opened 2017-01-08
Activity: Validation for subscription of external participants


jszanto:
Currently you can subscribe an external participant without entering a name or email, it might be better to add some validation here sometime.

tomudding:
This was fixed in #722. Therefore, I am closing this issue. If the problem still occurs, please reopen the issue or create a new one.

PR #722, opened 2017-01-10
Feature/external subscriptions

### LINKED ISSUE #124: Activity: Allow outsiders to subscribe
Allow an option for outsiders to subscribe for GEWIS activities

mstruijs:
Adds the option for activities to allow external subscriptions.
A few things to think about:
- Is this CAPTCHA ok? Should some more sophisticated/other tool be used for verification?
- Currently, external subscriptions have the same status as activity owner-created subscriptions. This means in particular that activity owners can remove external subscriptions. Is this the desired behaviour?

Resolves #124.


mstruijs:
Hmm. I guess actual input filtering and validation is fairly important to actually roll out this feature. So I guess I should implement that properly before merging this.

.gitignore
```diff
@@ -18,3 +18,4 @@ public/styleguide/
 autoload_classmap.php
 .mysql
 .vagrant/
+public/img/captcha/
```
jszanto: This folder doesn't exist now. You probably should create it and then commit it and re-add it to the gitignore.

module/Activity/src/Activity/Controller/ActivityController.php
```diff
@@ -167,6 +177,44 @@ public function signupAction()
...
+            $this->redirectActivityRequest($id, false, $error);
+            return;
+        }
+
+        $form = $signupService->getExternalForm($activity->getFields());
+        $form->setData($this->getRequest()->getPost());
+
+        //Assure the form is valid
+        if (!$form->isValid()) {
+            $error = $translator->translate('Invalid form');
```
jszanto: Just showing only this error message isn't very helpful, it would help if additionally the forum errors are shown in the view.

module/Activity/src/Activity/Form/ActivitySignup.php
```diff
@@ -58,6 +65,23 @@ public function initialiseExternalForm($fields)
...
 
+    public function initialiseExternalForm($fields)
+    {
+        $this->add([
+            'name' => 'captcha',
+            'type' => 'Zend\Form\Element\Captcha',
+            'options' => [
+                //'label' => 'Beep-boop I\'m a robot.',
+                'captcha' => new ImageCaptcha([
+                    'font' => getcwd() . '/public/fonts/bitstream-vera/Vera.ttf',
```
jszanto: Why is `getcwd()` needed here? Wouldn't just using `public/fonts/bitstream-vera/Vera.ttf` work?

jszanto: Also, I'm not sure whether this is zend framework issue or not, but some captcha's aren't really readable. 
screenshot from 2017-01-11 16-20-52
as you can see in the image, the last two characters fall outside of the captcha

jszanto:
* Validation indeed is a must, now it is possible to subscribe without entering a name / email, also see #717 
* The Captcha is fine I think, although it is a bit difficult
* I think it is fine that admins can remove externally subscribed members
* Also don't forget escaping for XSS protection in the views.

.gitignore
mstruijs: Well, uh, adding empty folders is not really something you should be doing with git. As the code will create a captcha, before looking for one, this does not lead to problems.

mstruijs:
The captcha is generated by Zend's internal function. I can reduce the amount of lines and dots, but that probably does not make the images more readable. So, unless someone offers an alternative to generate captcha's, I think we can keep it this way.

mstruijs:
I've addressed validation, so this fixes #717 as well.

PR #737, opened 2017-04-04
Fixed dropdown buttons in the main navigation.

### LINKED ISSUE #736: Improvement dropdowns main navigation
The dropdowns can be improved such that no refresh of the page is required for opening them on a mobile phone.

gafderks:
Fixes #736

### DIFF
--- a/module/Application/view/partial/main-nav.phtml
+++ b/module/Application/view/partial/main-nav.phtml
@@ -93,10 +93,21 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
         <div class="collapse navbar-collapse">
             <ul class="nav navbar-nav navbar-left">
                 <li class="dropdown dropdown-hover <?= $this->moduleIsActive(['frontpage', 'page']) ? 'active default' : '' ?>">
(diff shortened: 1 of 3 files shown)


gafderks:
@Mesoptier Can you verify whether this PR conflicts with your redesign of the front page?

gafderks:
@Mesoptier the problem is that we do want the dropdowns to open on hover when the screen is large while we want the dropdowns to open on click on mobile devices. This is not functionality that is in the default navbar. Unfortunately, the solution is not straightforward since the buttons themselves also need to link to a new page on large screens, while on mobile screens this behavior prevents the user from opening the dropdown.

Mesoptier:
@gafderks Ah, that makes sense. Well in that case this PR is fine by me.

armandduijn:
Hmm it's lacking in usability. I would prefer if there was pointy arrow (▼) to communicate that a menu item is a dropdown. It should also point upwards (▲) when the dropdown is expanded.

gafderks:
@armandduijn I've implemented this feature.

PR #754, opened 2017-07-11
Authentication api

kokx:
De authenticatie API.

Basically, de admin is nog niet gebouwd. De rest werkt al wel.

Als iemand de admin wil implementeren mag dat. Wel 1 heel belangrijk dingetje. Implementeer het genereren van secrets op een fatsoenlijke manier met behulp van [random_bytes](https://php.net/random_bytes)

Zie voor informatie van hoe dit geimplementeerd is: https://github.com/GEWIS/gewisweb/wiki/External-authentication

### DIFF
--- a/module/User/src/User/Service/ApiApp.php
+++ b/module/User/src/User/Service/ApiApp.php
@@ -3,11 +3,12 @@
 
 namespace User\Service;
 
-use Application\Service\AbstractAclService;
+use Application\Service\AbstractService;
+use Firebase\JWT\JWT;
 use User\Mapper\ApiApp as ApiAppMapper;
-use User\Model\User;
+use User\Model\User as UserModel;
 
-class ApiApp extends AbstractAclService
+class ApiApp extends AbstractService
 {
 
     /**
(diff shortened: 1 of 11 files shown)


module/User/src/User/Service/ApiApp.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param string $appId
+     * @param UserModel $user
+     * @return string
+     */
+    public function callbackWithToken($appId, UserModel $user)
+    {
+        $app = $this->getMapper()->findByAppId($appId);
+
+        $token = [
+            'iss' => 'https://gewis.nl/',
```
jszanto: What is the `iss` field used for and why isn't it dynamic? (So it also works on test.gewis.nl)

armandduijn: It's arbitrary. According to the [spec](https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#issDef) it can be anything you want because it's up to an application to deal with it. So hardcoding gewis.nl isn't necessarily incorrect but making it dynamic might help other application distinguish if they're receiving a token from a developer or from a real user.

kokx: The `iss` field is the issuer field. As described in [RFC 7519](https://tools.ietf.org/html/rfc7519#section-4.1.1).

It doesn't need to be verified. And will not be verified (thus, it can be used on `test.gewis.nl`). It is just to show that the token was issued by the GEWIS Website.

armandduijn: ^ I approve this message

Issue #733, opened 2017-03-09
Send organ update emails in batch


gafderks:
![image](https://cloud.githubusercontent.com/assets/7315457/23756412/d975e1f8-04e3-11e7-9118-fad159e28a23.png)

armandduijn:
Duplicate. See issue #723.

Issue #723, opened 2017-01-14
Reduce the amount of notification emails


jszanto:
Currently an email is sent on every edit of an activity or organ, this can result in a lot of emails. Also some redundant emails are sent on some admin actions. A possible solution would be to schedule emails sometime in the future and then only send one email if multiple edits are made in a short time period.

Staartvin:
I agree. What should the time period be? 24 hours?

jszanto:
I think 1-3 hours would work, since repetitive edits are usually made quickly after eachother. This issue could probably be combined with #682.

mstruijs:
A side remark, but I'm still not receiving any notifications whatsoever. I know that this is due to forces beyond the website design, but I suppose I'd mention it again.

gafderks:
![image](https://cloud.githubusercontent.com/assets/7315457/23756547/562f1c50-04e4-11e7-8aeb-a064cae08214.png)

PR #761, opened 2017-09-20
Fix unintended, annoying and unnecessary feature

### LINKED ISSUE #682: Use a queue for mail
Currently most emails are sent inside the request which is not a good idea. It leads to reduced performance and mail errors result in the request failing.

armandduijn:
Fixes #682 #723 #733.

**Impact**: Improves the mood of the developers and saves costly bandwidth and storage.

### DIFF
--- a/module/Decision/src/Decision/Service/Organ.php
+++ b/module/Decision/src/Decision/Service/Organ.php
@@ -186,10 +186,6 @@ class Organ extends AbstractAclService
             $this->approveOrganInformation($organInformation);
         }
 
-        $this->getEmailService()->sendEmail('organ_update', 'email/organUpdate',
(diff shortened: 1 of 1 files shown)


armandduijn:
What needs to be checked? Is it something we can do automatically?

jszanto:
The changes need to be reviewed before they are placed on the website. This is why you receive an email prompting you to review them. Perhaps fixing this is a better idea: https://github.com/GEWIS/gewisweb/issues/723

jszanto:
Also disabling the emails whenever an admin does something would already reduce the amount by 50%

armandduijn:
In light of new information, this PR has been deemed unnecessary.

PR #767, opened 2017-11-02
Wrap long words in committee names

### LINKED ISSUE #764: Long Organ names break layout on mobile
screenshot_20171021_185917

raketwissenschaftler:
Fixes #764

### DIFF
--- a/public/scss/components/_headings.scss
+++ b/public/scss/components/_headings.scss
@@ -4,3 +4,7 @@ h1, h2, h3, h4, h5, h6 {
         padding-bottom: 0.5em;
     }
 }
+// Word-wrap the committee names when needed
+h1{
+    word-wrap: break-word;
+}


public/scss/components/_headings.scss
```diff
@@ -4,3 +4,7 @@ h1, h2, h3, h4, h5, h6 {
         padding-bottom: 0.5em;
     }
 }
+// Word-wrap the committee names when needed
+h1{
```
gafderks: Do we want this behavior for _all_ primary headings?

raketwissenschaftler: Good point. I don't see any reasons for not doing it, but it is a bit too broad of a selector, maybe?

armandduijn: Just to be safe, we could add a new class `.break-word` to the primary headings on the organ pages. This makes sure we don't break the styling on other pages.

PR #771, opened 2017-11-02
Fix issue where email addresses are case sensitive

Staartvin:
1) Registering will now lowercase the email address.
2) Signing in with an email address is now also case insensitive.


module/User/src/User/Mapper/User.php
```diff
@@ -60,11 +60,11 @@ public function findByLogin($login)
         if (is_numeric($login)) {
             $qb->where('u.lidnr = ?1');
         } else {
-            $qb->where('m.email = ?1');
+            $qb->where('m.email LIKE ?1');
         }
 
         // set the parameters
-        $qb->setParameter(1, $login);
+        $qb->setParameter(1, $login = str_replace("%", "", $login));
```
jszanto: I think the `$login = ` doesn't quite belong here. Also, single quotes > double quotes.

module/User/src/User/Mapper/User.php
```diff
@@ -60,11 +60,11 @@ public function findByLogin($login)
         if (is_numeric($login)) {
             $qb->where('u.lidnr = ?1');
         } else {
-            $qb->where('m.email = ?1');
+            $qb->where('m.email LIKE ?1');
```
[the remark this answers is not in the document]
jszanto: Good suggestion!

armandduijn:
@jszanto Do we need to lowercase all the existing emails in the database?

Koen1999:
> 
> 
> @jszanto Do we need to lowercase all the existing emails in the database?

I don't think that's neccesary with this code.

I think that we can actually merge this sooner or later

I do have one question for @Staartvin though
What does this piece of code do?
``str_replace('%', '', $login)``

module/User/src/User/Mapper/User.php
```diff
@@ -60,11 +60,11 @@ public function findByLogin($login)
         if (is_numeric($login)) {
             $qb->where('u.lidnr = ?1');
         } else {
-            $qb->where('m.email = ?1');
+            $qb->where('m.email WHERE LOWER(m.email) = ?1');
         }
 
         // set the parameters
-        $qb->setParameter(1, $login);
+        $qb->setParameter(1, str_replace('%', '', $login));
```
jszanto:
```suggestion
        $qb->setParameter(1, strtolower($login));
```

module/User/src/User/Model/NewUser.php
```diff
@@ -119,7 +119,7 @@ public function setLidnr($lidnr)
      */
     public function setEmail($email)
     {
-        $this->email = $email;
+        $this->email = strtolower($email);
```
jszanto: This doesn't seem needed if we just do it on the fly.

Koen1999:
When is case sensitivity of email addresses actually a problem?

Koen1999:
Please refer to #915 from here. It's the continuation of the PR, so we can solve the issue

PR #781, opened 2017-12-10
Alter design of gewis.nl/photo

Staartvin:
See BIG SCREENimage

And

SMALL SCREEN
image


jszanto:
Perhaps we could use this same album style for subalbums too?

armandduijn:
I like the new design! However, it's not immediately clear what the dates under each album name mean. The creation date? Or the last modified date?

jszanto:
The dates are the date when the foto's were taken.

jszanto:
And two small issues at different screen sizes:
screen shot 2017-12-10 at 17 32 34
and:
screen shot 2017-12-10 at 17 33 07

Mesoptier:
I like the design. The code for the styles (classnames for example) needs to be improved before merging. I can look at that next week.

Staartvin:
@jszanto The first issue is fixed, but the second one is little harder. It occurs when titles are too long, right?

jszanto:
@Staartvin Yes, it occurs (on medium sized screens) when the titles don't fit on one line anymore. I suppose it could be fixed by setting a min-height for the header, but that might be ugly.

jszanto:
@Mesoptier @Staartvin Any updates on this?

Staartvin:
@jszanto I've fixed the first issue, but could not resolve the second issue. Perhaps @Mesoptier could look into this?

Kevinjil:
Something like [this](https://gist.github.com/Kevinjil/042214dd617c6190fbdc3e0803b743ac) could be used to fix the column alignment with long titles. It would most likely be a good idea to modify that to create less general selectors.

StephanvandenBerkmortel:
This is a problem with the current photo page design as well. For that design I posed a solution <a href="https://github.com/GEWIS/gewisweb/pull/819">here</a> and can probably be applied to the new design as well.

Issue #511, opened 2016-02-07
Display photos of the week by year


jszanto:

jszanto:
Or provide some other way to see from which year a photo was on the photo of the week page.

gafderks:
I plan on implementing an album of photos of the week with subalbums per year (same approach as for albums of a member, #782), does this offer the wanted functionality?

Kevinjil:
@gafderks I think that would be even better as just displaying them by year.

Koen1999:
This is still something that we can implement

tomudding:
This was implemented with #1424.

PR #782, opened 2017-12-10
Implemented browsable albums for members

### LINKED ISSUE #553: Photo: page through all photos of a member
When clicking on a photo at `/member/:lidnr` make the next and previous buttons go the the next and previous photos of the corresponding member instead of the next and previous photos in the album.

### LINKED ISSUE #681: Sort the photos of a member in some sort of way
There is currently no (logical) sorting applied to the photos at `/member/<member ID>`, which makes them impossible to navigate. Sorting by date (latest first?) would make more sense.

### LINKED ISSUE #695: Make member photo's browseable like an album
When you view the photo's of a member, you should be able to go into an album of all the user's photos. Next and previous should work correctly here as well. If done correctly, this should fix #681 as well.

gafderks:
Fixes #695 
Fixes #553
Fixes #681

Viewing an image from the album of a member (previous and next pointers work as expected):
image

Plan for the future: Also implement browsing pictures of the week in the same way (see #511).


gafderks:
It would be nice to test this feature on a database that has more photos and members than mine. Maybe on test.gewis.nl?

jszanto:
@gafderks I put it up on test.gewis.nl, looks good! I’ll take a look at the code tomorrow.

gafderks:
@jszanto The functionality is done, but I plan on still refactoring some of the code.

gafderks:
I found out that the pagination is not working correctly yet.

module/Photo/src/Photo/Service/Album.php
```diff
@@ -321,6 +336,25 @@ public function getCreateAlbumForm()
         }
         return $this->sm->get('photo_form_album_create');
     }
+    
+    public function getMemberAlbum($lidNr)
+    {
```
jszanto: Perhaps it would be better to use a query to do this. You could join the tags and the photos table and then order by the photo date. This might also make pagination easier?

module/Photo/src/Photo/Model/VirtualAlbum.php
```diff
@@ -0,0 +1,317 @@
...
+
+namespace Photo\Model;
+
+/**
+ * VirtualAlbum.
+ *
+ * @ORM\HasLifecycleCallbacks
+ *
+ */
+class VirtualAlbum extends Album
```
jszanto: All the doctrine comments in here don't seem to be needed.

gafderks:
@jszanto I have refactored the VirtualAlbum and managed to get the pagination working using the query as you suggested. Can you maybe install the updated version on http://test.gewis.nl to try it out?

jszanto:
@gafderks I updated it, looks good!

jszanto:
Before I merge this, could you reformat your code according to the PSR-2 standard? (Should be one click in your IDE)

module/Photo/src/Photo/Model/VirtualAlbum.php
```diff
@@ -0,0 +1,272 @@
...
+    /**
+     * Album ID.
+     *
+     */
+    protected $id;
+    
+    /**
+     * First date of photos in album
+     */
+    protected $startDateTime = null;
```
gafderks: @jszanto Would it be better to remove these variables? They can be inherited from the Album class, but there they have ORM annotations and here the explicitly do not.

gafderks:
@jszanto I have reformatted the code according to the PSR-2 standard.

module/Photo/src/Photo/Model/VirtualAlbum.php
jszanto: The ORM notations won’t be inherited, so it would be better to remove them.

gafderks: I have removed it now.

Issue #634, opened 2016-09-12
Show the GPS location of a photo in the details


gafderks:
This is especially interesting for weekends like the GELIFT weekend.

jszanto:
How would we obtain the location? I think the location will only be available on the very few photos which are taken with a phone.

gafderks:
That is true. We can only display it if it is included in the EXIF data.

kokx:
@jszanto Actually, a lot of the photo's might already contain GPS data. AFAIK, the GEFLITST camera has a GPS function.

PR #785, opened 2017-12-12
Feature/photo gps

Kevinjil:
Implementation of #634.


module/Photo/src/Photo/Service/Metadata.php
```diff
@@ -21,7 +21,7 @@ class Metadata extends AbstractService
      */
     public function populateMetadata($photo, $path)
     {
-        $exif = read_exif_data($path, 'EXIF');
+        $exif = exif_read_data($path, 'EXIF');
```
armandduijn: Interesting. Didn't know that `read_exif_data` has an alias.

Kevinjil: `read_exif_data` is the alias of `exif_read_data`

PR #792, opened 2017-12-22
Redesign member page

armandduijn:
This PR includes a redesign of the [member page](https://gewis.nl/member). The goal of the redesign is to make relevant information easier to find.

Note that this PR is a work in progress. Things might move around. However, if you don't agree with one of the changes, please comment below.

> [Before](https://user-images.githubusercontent.com/706385/34313993-dcbb9c9e-e76f-11e7-81c0-b9174b983965.png) / [After](https://user-images.githubusercontent.com/706385/34313998-ee3009f6-e76f-11e7-88ac-d4712fb34fa3.png)

**Added:**
- _Latest meetings_ - Quickly navigate to recent meeting notes
- _Search form for decisions_ - Removes the need of navigating to another page to enter the search query.
- _Link to manage activities_ - Hopefully this helps members that haven't discovered the power of the admin page
- _Link to manage organ page_ - Same argument as above...

**Removed:**
- _Microsoft Imagine link_ - Does this link need to be here? It's also on the [education page](https://gewis.nl/education).
- _Change password link_ - Again, should this be here? This page is intended as a portal for GEWIS members. Changing one's password is an account specific action.

### DIFF
--- a/module/Decision/src/Decision/Mapper/Meeting.php
+++ b/module/Decision/src/Decision/Mapper/Meeting.php
@@ -112,19 +112,7 @@ class Meeting
      */
     public function findLatestAV()
     {
-        $qb = $this->em->createQueryBuilder();
-
-        $today = new \DateTime();
-        $maxDate = $today->sub(new \DateInterval('P1D'));
-        $qb->select('m')
-            ->from('Decision\Model\Meeting', 'm')
-            ->where('m.type = AV')
-            ->where('m.date >= :date')
-            ->orderBy('m.date', 'DESC')
-            ->setParameter('date', $maxDate)
-            ->setMaxResults(1);
-
-        return $qb->getQuery()->getOneOrNullResult();
+        return $this->findFutureAV('DESC');
     }
 
     /**
@@ -134,20 +122,7 @@ class Meeting
      */
     public function findUpcomingAV()
     {
-        $qb = $this->em->createQueryBuilder();
-
-        $today = new \DateTime();
-        $maxDate = $today->sub(new \DateInterval('P1D'));
-
-        $qb->select('m')
-            ->from('Decision\Model\Meeting', 'm')
-            ->where('m.type = AV')
-            ->where('m.date >= :date')
-            ->orderBy('m.date', 'ASC')
-            ->setParameter('date', $maxDate)
-            ->setMaxResults(1);
-
-        return $qb->getQuery()->getOneOrNullResult();
+        return $this->findFutureAV('ASC');
     }
 
     /**
@@ -223,5 +198,29 @@ class Meeting
     public function getRepository()
     {
         return $this->em->getRepository('Decision\Model\Meeting');
+    }
+
+    /**
+     * Finds an AV planned in the future
+     *
+     * @param string $order Order of the future AV's
+     * @return \Decision\Model\Meeting|null
+     */
+    private function findFutureAV($order)
+    {
+        $qb = $this->em->createQueryBuilder();
+
+        $today = new \DateTime();
+        $maxDate = $today->sub(new \DateInterval('P1D'));
+
+        $qb->select('m')
+            ->from('Decision\Model\Meeting', 'm')
+            ->where('m.type = AV')
+            ->where('m.date >= :date')
+            ->orderBy('m.date', $order)
+            ->setParameter('date', $maxDate)
+            ->setMaxResults(1);
+
+        return $qb->getQuery()->getOneOrNullResult();
     }
 }
--- a/module/Decision/src/Decision/Mapper/Meeting.php
+++ b/module/Decision/src/Decision/Mapper/Meeting.php
@@ -215,9 +215,10 @@ class Meeting
 
         $qb->select('m')
             ->from('Decision\Model\Meeting', 'm')
-            ->where('m.type = AV')
-            ->where('m.date >= :date')
+            ->where('m.type = :type')
+            ->andWhere('m.date >= :date')
             ->orderBy('m.date', $order)
+            ->setParameter('type', 'AV')
             ->setParameter('date', $maxDate)
             ->setMaxResults(1);
 
--- a/module/Decision/src/Decision/Mapper/Meeting.php
+++ b/module/Decision/src/Decision/Mapper/Meeting.php
@@ -75,7 +75,8 @@ class Meeting
      * @param int|null $limit The amount of results, default is all
      * @return array Meetings that have taken place
      */
-    public function findPast($limit = null) {
+    public function findPast($limit = null)
+    {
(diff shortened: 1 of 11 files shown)


jszanto:
Looks good overall, few comments though:
- How does it look for non-active members?
- What about future meetings (e.g. meetings which have been added already so the pieces are made public)
- Organs doesn't really seem very relevant for "managing your committee", also I thought we were planning to get rid of that page anyways.
- Administrator probably isn't really needed as I assume "activities" and "organ page" already link to the website admin.

jszanto:
And the link for authorizations isn't very visible now. Perhaps we could add a button to page of the meeting itself?

gafderks:
@armandduijn I have added a link to the photo's of a member on this page. See https://github.com/GEWIS/gewisweb/commit/bf796e486e9770f24397024175839952890e0f60
Maybe you can incorporate this button in your redesign?

armandduijn:
@gafderks Isn't this already part of the page? The 'My Photo's' button, right?

armandduijn:
I implemented the feedback but it still requires some code clean-up.

> [Updated](https://user-images.githubusercontent.com/706385/34321424-3ff4cd54-e80f-11e7-8a9b-a8425196da3f.png) (w/ upcoming meeting) / [Updated](https://user-images.githubusercontent.com/706385/34321428-4ea52e34-e80f-11e7-9359-bd1b2403e939.png) (w/o upcoming meeting)

**Added:**
- Re-added the authorization link
- Added a notice for the upcoming general meeting. This assumes that at most one general meeting is planned at a time.

**Removed:**
- Removed the organs link

@jszanto I kept the administrator button because it might be relevant for members with "higher" roles (e.g. admin). Furthermore, how can I reproduce a non-active member? Logging in as a non-active member (e.g. member that is not attached to an organ) doesn't make a difference even though the buttons are [wrapped by an ACL](https://github.com/armandduijn/gewisweb/blob/bec128dbf68cbdbc2b54ce3aa2da0079ead949de/module/Decision/view/decision/member/index.phtml#L114).

gafderks:
@armandduijn Nice, I look over it...

BorBorBor:
> This assumes that at most one general meeting is planned at a time.

Although rare, having multiple planned meetings at the same time does happen (once every few years or so). Since it doesn't happen that often I wouldn't go through too much trouble to prepare for this, but it shouldn't break anything.

armandduijn:
My work here is done.
 
The only "bug" is that the PHP function `strftime()` outputs English dates (as seen in the screenshots below). But according to the internet this is because my system, in this case the Vagrant VM, doesn't have the Dutch language files installed. So maybe on test.gewis.nl everything works fine.

- [Desktop](https://user-images.githubusercontent.com/706385/34692928-26a8aad6-f4c2-11e7-802a-5ed1074c1451.png)
- [Tablet](https://user-images.githubusercontent.com/706385/34692929-2c5280a6-f4c2-11e7-9541-76ba647e21b7.png)
- [Mobile](https://user-images.githubusercontent.com/706385/34692935-317b8eba-f4c2-11e7-9dab-5602a2fc2a67.png)

jszanto:
@armandduijn can you provide a screenshot of how it looks for non-active members

module/Decision/src/Decision/Service/Member.php
```diff
@@ -72,6 +72,24 @@ public function getMembershipInfo($lidnr = null)
...
+     * Returns is the member is active
+     *
+     * An active member is a member that is part of at least one organ or is part of
+     * the board.
+     *
+     * @param MemberModel $member
+     * @return bool
+     */
+
+    public function isActiveMember(MemberModel $member)
```
armandduijn: I don't know how correct this method is but it gets the job done for presenting the correct content. We should probably rely on proper ACL's when making changes.

armandduijn: I don't know how correct this method is but it gets the job done for presenting the correct content. We should probably rely on proper ACL's when making changes.

armandduijn:
@jszanto I implemented a new method to determine if a member is active. If he/she isn't, it simply hides the "manage" panel ([screenshot](https://user-images.githubusercontent.com/706385/34821809-2ea8722a-f6c5-11e7-8d8a-89fd3bc04567.png)). If there is a better way to check this, please let me now.

jszanto:
@armandduijn Why not use the same acl check which already was in place earlier? That one worked fine.

armandduijn:
@jszanto Except that it didn't :/. I logged in as member 8001 (no organ membership or board installation) on my local setup and the ACL didn't prevent the panel from showing.

jszanto:
@armandduijn Did that member have an admin role by any chance?

armandduijn:
@jszanto Nope, just a regular member. No defined role.

module/Decision/src/Decision/Service/Member.php
jszanto: There is an `active_member` role which you should use in combination with the existing ACL stuff.

module/Decision/src/Decision/Mapper/Meeting.php
```diff
@@ -87,6 +123,29 @@ public function findLatestAV()
         return $qb->getQuery()->getOneOrNullResult();
     }
 
+    /**
+     * Returns the closest upcoming AV
+     *
+     * @return \Decision\Model\Meeting|null
+     */
+    public function findUpcomingAV()
+    {
```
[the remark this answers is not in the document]
armandduijn: The method `findLatestAV()` returns the AV the furthest in the future. The `findUpcomingAV()` looks for the closest AV in the future. Simply put, it handles the edge case Bor mentioned, where, in some cases, multiple AV's are created in the database.

armandduijn:
Rebased the branch on the latest master branch and updated the page to reflect the features of the current member's page ("Links" and "Public archive" links).

The last couple of commits re-introduced a couple of buttons that were removed initially. Whether they should be there or not is something for future discussion. Right now the goal is to merge this PR with minimal impact on the set of features of the page.

> [Updated](https://user-images.githubusercontent.com/706385/52301438-af9c6a00-298a-11e9-856c-b42a09f24492.png) (w/ upcoming meeting) / [Updated](https://user-images.githubusercontent.com/706385/52301461-bd51ef80-298a-11e9-9cef-7b49df69dbdc.png) (w/o upcoming meeting)

**Added:**
- Re-added the "Organs", "Change password", and "Microsoft Imagine" link
- Split the list of latest meetings in three separate lists

**Changed:**
- Rules (reglementen) are visible to all members of GEWIS (not only active members)
- Used ACL's to determine if the user is allowed to modify an organ instead of the ad-hoc version


Note that this PR does not contain the appropriate translations to prevent merge conflicts. Translations should be done after merging and directly committed to the master branch.

Issue #801, opened 2018-01-19
Language fallback for activities


jszanto:
Provide some sort of easy fallback mechanism when one of the languages for an activity is missing instead of the hacky methods we currently have for this.

raketwissenschaftler:
This should be implemented for job offers as well. Maybe implement it directly into the translate service?

mstruijs:
Indeed, the current system is a mess that makes additions to the fields for activities an painful exercise (updates in about 3 different locations that are not at all obvious are required). This is the part that made me realize that the activity module needs a serious redesign.

PR #802, opened 2018-01-19
Hotfixes

jszanto:


module/Frontpage/view/frontpage/frontpage/home.phtml
```diff
@@ -99,7 +99,7 @@
                         <div class="list-group">
                             <?php foreach ($activities as $activity): ?>
                                 <a class="list-group-item" href="<?= $this->url('activity/view', ['id' => $activity->getId()]) ?>">
-                                    <h4 class="list-group-item-heading"><?= $activity->getName() ?></h4>
+                                    <h4 class="list-group-item-heading"><?= $activity->getNameEn() ?></h4>
```
armandduijn: Why is it forcing the English translation?

jszanto: Because previously it was forcing the Dutch translation which isn't always available. This will be fixed properly once #801 is fixed.

armandduijn: In that case, I approve this hack!

Issue #803, opened 2018-01-20
Organs reachable with invalid URLs


Kevinjil:
Currently, you can visit the organ pages with invalid URLs. For example, it is possible to visit the page of an AVC or regular committee at the URL 'association/fraternity/\<abbr\>'. It would be better to give a HTTP 404.

Related: https://github.com/GEWIS/gewisweb/issues/774#issuecomment-359198832

PR #805, opened 2018-01-21
Fixed issue where organs are reachable with wrong URLs

Kevinjil:
Fix for #803

### DIFF
--- a/module/Frontpage/src/Frontpage/Controller/OrganController.php
+++ b/module/Frontpage/src/Frontpage/Controller/OrganController.php
@@ -32,10 +32,16 @@ class OrganController extends AbstractActionController
 
     public function organAction()
     {
+        $type = $this->params()->fromRoute('type');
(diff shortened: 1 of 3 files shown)


module/Frontpage/src/Frontpage/Controller/OrganController.php
```diff
@@ -32,10 +32,16 @@ public function fraternityListAction()
 
     public function organAction()
     {
+        $type = $this->params()->fromRoute('type');
         $abbr = $this->params()->fromRoute('abbr');
         $organService = $this->getOrganService();
         try {
             $organ = $organService->findOrganByAbbr($abbr);
```
jszanto: Wouldn't it be better to adjust this function to include the type in the query?

Kevinjil: I wanted to keep the query as general as possible, but I think I've found a suitable solution in the next commit.

PR #812, opened 2018-01-27
Add a missing >

jszanto:
Opinions and discussion are welcome on the new design.

PR #817, opened 2018-01-29
Revert "Add a missing >"

jszanto:
Reverts GEWIS/gewisweb#812

### DIFF
--- a/module/Company/view/company/company/job-list.phtml
+++ b/module/Company/view/company/company/job-list.phtml
@@ -96,7 +96,7 @@ $this->headTitle($this->category->getPluralName());
                                     </p>
                                     <?php if ($job->getAttachment() !== ''): ?>
                                         <p class="job-vacancy">
-                                            <a href="<?= $this->fileUrl($job->getAttachment()); ?>">
+                                            <a href="<?= $this->fileUrl($job->getAttachment()); ?>"
                                                 <?= $this->translate('View'). ' ' . $job->getCategory()->getName(); ?>
                                             </a>
                                         </p>


BorBorBor:
@raketwissenschaftler thanks 4 assigning this to me, I have found the missing part: 

# >

jszanto:
I suggest we create forks of Gecko, WebKit and Blink to support missing >. This would save us the effort of convincing W3C from implemting this in the specification.

BorBorBor:
The old website did not have this problem, and editing it live via notepad on the GEWIS PCs did not yield this problem. I suggest rolling back the current version.

raketwissenschaftler:
@BorBorBor I appreciate the suggestion, however, I think that important issues as this really need some more care. I would prefer if important issues such as these could be handled a bit more carefully and seriously in the future. Please also take into account what @jszanto said.

@Kevinjil: Can you please take a look at this issue?

BorBorBor:
@kokx maybe we should consider asking Coppelmans to start drafting a proposal for a Senate Inequality.

kokx:
@BorBorBor And of course, this has to be discussed at length at the Senaatsavond, before the Senate can make a decision about this important topic.

raketwissenschaftler:
I have discussed the proposed changes with the board. They did not have a clear opinion about this, we will need to discuss it on the next GMM.

BorBorBor:
I disagree about bothering the whole GMM with this. My preference would be to make this a key point in the long-term vision of GEWIS.

kokx:
I agree that this needs to be part of long-term plans. However, I think this is much, much bigger than GEWIS. It should be obvious that this needs a judgement from the WISO.

jszanto:
Closing this for now until we reach a consensus about this issue.

PR #819, opened 2018-02-07
Fixed text wrap of photos

StephanvandenBerkmortel:
Multi-line captions of photo collections are now shortened to one line with ellipsis styling to avoid other tile collections not being appended properly

### DIFF
--- a/public/scss/modules/_photo.scss
+++ b/public/scss/modules/_photo.scss
@@ -37,6 +37,9 @@ Photo
 }
 
 .thumbnail .caption {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
   padding: 9px 0 0;
   min-height: 50px;
 }


public/scss/modules/_photo.scss
```diff
@@ -37,6 +37,9 @@ Photo
 }
 
 .thumbnail .caption {
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
raketwissenschaftler: Do we want the album names to be shortened with an ellipsis, or do we want to make the row distance variable?

jszanto: well it stimulates people to come up with short album names

raketwissenschaftler: Agreed. Merge it I'd say. Can you do this @jszanto ?

PR #820, opened 2018-02-15
Feature/797 filebrowser 

### LINKED ISSUE #797: Search through public dir from members panel
Allow logged in members to look through the public directory(containing logos and templates) from the GEWIS-website.

mstruijs:
Implements a file browser for, among others, files stored in /public , fixes #797 .

Bare-bones works as intended, but has important UI improvements pending:
- [x] Navigating up on a dir
- [x] 'Breadcrumbs'  for navigating up to any containing dirs
- [x] Place all folders on top, then the files.

Further UI suggestions are welcome.


mstruijs:
'Codeclimate: ListDir is te complex' _Really_? Een early return voor defensive programming en een foreach die een if heeft om een paar cases te skippen is te complex? Laat me raden, als ik in plaats van een continue een if statement pak is het wel ok? Wat een onzin. #wont-fix

mstruijs:
Apart from features already mentioned, filenames are now html escaped and filenames starting with '.' aren't listed.

mstruijs:
Ok, this is pretty much done now. 

UI might not be exactly the correct style, can a frontend specialist verify this?

@kokx is in the review list, as he wants to pentest some parts here.

module/Decision/src/Decision/Controller/FileBrowser/DummyReader.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
jszanto: Is this file still required?

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
```diff
@@ -0,0 +1,83 @@
...
+        $fullPath = $this->root . $path;
+        if (!is_file($fullPath)) {
+            return null;
+        }
+        $response = new \Zend\Http\Response\Stream();
+        $response->setStream(fopen('file://' . $fullPath, 'r'));
+        $response->setStatusCode(200);
+        $headers = new \Zend\Http\Headers();
+        $headers->addHeaderLine('Content-Type', 'application/octet-stream')
+                ->addHeaderLine('Content-Disposition', 'attachment; filename="' . end(explode('/', $fullPath)) . '"')
```
jszanto: If you leave out `attachment`, then the browser will just display the file rather than downloading it immediately.

jszanto:
Regarding the security of this, there is a slight risk of a path traversal vulnerability in here. At the moment it \appears\ to be safe. If you input an url such as `/decision/files/../../../../../../etc/passwd`, the webserver will save you by rewriting that to `/`. However it may not be a good idea to rely on the webserver for this.

For instance if someone later decides to replace `fromRoute` by `fromQuery` over here: https://github.com/GEWIS/gewisweb/pull/820/files#diff-035f387c5ac15025451bfe31c46567c9R144, then this code immediately becomes vulnerable to path traversal.

module/Decision/Module.php
```diff
@@ -112,6 +112,10 @@ public function getServiceConfig()
                         $sm->get('decision_doctrine_em')
                     );
                 },
+                'decision_fileReader' => function ($sm) {
+                    //NB: The returned object should implement the FileReader Interface.
+                    return new \Decision\Controller\FileBrowser\LocalFileReader(getcwd() . '/public/webfiles/');
```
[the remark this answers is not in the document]
mstruijs: Do you mean only the path, or the class as well? I agree with the path, not sure whether class is a good idea (creates a dependency on internal classes in the config)

module/Decision/src/Decision/Controller/FileBrowser/DummyReader.php
mstruijs: Not necessarily. Could be useful for local UI testing in case the given dir is empty.

module/Decision/Module.php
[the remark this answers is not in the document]
jszanto: Only the path

module/Decision/src/Decision/Controller/DecisionController.php
```diff
@@ -129,6 +130,46 @@ public function authorizationsAction()
...
+        }
+        //$fileReader =  new LocalFileReader(getcwd() . '/public/webfiles/');
+        $fileReader = $this->getServiceLocator()->get('decision_fileReader');
+        if ($fileReader->isDir($path)) {
+            //display the contents of a dir
+            $folder = $fileReader->listDir($path);
+            if ($folder===null) {
+                return $this->notFoundAction();
+            }
+            $trailingSlash = (strlen($path)>0 && $path[strlen($path)-1]==='/');
```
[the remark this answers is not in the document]
mstruijs: Not picked up by Codeclimate, for some reason. Will fix.

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
mstruijs: Ah, I just copied this from somewhere. I should do that.

module/Decision/src/Decision/Controller/FileBrowser/DummyReader.php
jszanto: For testing, perhaps set the default path in the config to some harmless path, e.g. getcwd() . /public/

mstruijs:
@jszanto 
Hmm, more fragile than I thought. Tbh, I didn't quite understand why the up traversal stopped at the root of the domain. Filtering the string `/../` (or maybe `/..` at the end of the path as well? ) should do the trick, right?

jszanto:
@mstruijs just filtering `/../` still would allow tricks such as `./../.`. I think a safer approach would be to use something like `realpath()` to resolve the path and then check if the configured directory is in that path.

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
```diff
@@ -0,0 +1,83 @@
...
+
+    public function listDir($path)
+    {
+        //remove the trailing slash from the dir
+        $fullPath = $this->root . $path;
+        if (!is_dir($fullPath)) {
+            return null;
+        }
+        //We can insert an additional /, except when when $path is the root
+        $delimiter = $path !== '' ? '/': '';
```
Kevinjil: When you fix the stuff mentioned by @jszanto, lease also add a space before the `:` ;)

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
```diff
@@ -0,0 +1,83 @@
...
+                $kind = 'dir';
+            } elseif (is_file($fullPath . '/' . $dircontent)) {
+                $kind = 'file';
+            } else {
+                //Ignore all strange filesystem thingies like symlinks and such
+                continue;
+            }
+            $files[] = new FileNode(
+                $kind,
+                htmlspecialchars($path . $delimiter . $dircontent),
```
Kevinjil: Wouldn't it be more 'correct' to store the valid paths in the FileNode object? This way, the nodes still contain the actual valid paths, like the comments added in the class state and you can make the call to htmlspecialchars in the views.
Otherwise I would suggest making a small change in the documentation of the FileNode class.

mstruijs:
By filtering, I mean not matching the regex at the router level. I don't
think your trick would work then. I guess an extra check would be a good
idea, but I think this should be detected independent of the actual file
system as well (for usability reasons, at least) and I think the routing
level is fine.


2018-02-18 23:58 GMT+01:00 jszanto <notifications@github.com>:

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
mstruijs: @jszanto Why do we do the codeclimate thing again? That ghastly checker nags on and on about so-called 'cognitive complexity' (the mere assumption something with that name could be detected automatically by an entity that has no cognition is a logic fallacy). But it ignores simply stylistic stuff like here!

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
mstruijs: I'd prefer to invoke htmlspecialchars sooner rather than later. If there are truly filepaths that have html chars in them, how on earth would I be able to access them? In other words, I'm working on the reasonable assumption that given that a path is valid, htmlspecialchars does nothing.

I think the documentation in FileNode is correct. It represents a node in a 'filesystem', where 'filesystem' here means 'whatever it is FileReader browses through'. The 'path' is valid in the sense that it is the location of the node as far as the FileReader is concerned. (hmtlspecialchars doesn't change this, as I already mentioned)

module/Decision/src/Decision/Controller/FileBrowser/LocalFileReader.php
mstruijs: Btw, attachment isn't enough (on Firefox, at least). Dropping 'application' as well seems to do the trick, though.

mstruijs:
Alright, @jszanto , @Kevinjil , I think I adressed or fixed all of your comments and merged into testing.

mstruijs:
@kokx Content header moet idd gefixt worden. Had eerst application/octet-stream voor alles, waardoor een download geforced wordt. Is dit prima, security-wise? 

Meer escapen is idd nodig, duidelijk.

Symlinks zijn expres genegeerd. @jszanto moet dit anders?

Dotfiles worden niet getoond by design. (een reden is ._DS_Store en dergelijke)

`;` wordt geexclude door de regex `(?<path>[^?*:;{}\\\]*)`. Moet dit in files toegestaan worden? Zijn er andere characters in deze regex die in files toegestaan mogen worden?

module/Decision/view/decision/member/index.phtml
```diff
@@ -19,6 +19,7 @@ $sections[$this->translate('Meetings')] = [
     [$this->translate('Decision Search'), 'search', $this->url('decision/search')],
     [$this->translate('Authorizations'), 'pencil', $this->url('decision/authorizations')],
     [$this->translate('Organs'), 'heart', $this->url('organ')],
+    [$this->translate('Files'), 'download-alt', $this->url('decision/files', ['path' => ''])]
```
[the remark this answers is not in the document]
mstruijs: Ja, is wss beter.

mstruijs:
Hmm, maybe I just use the same regex at the router to filter nodes. That makes things consistent, at least.

kokx:
Afaik kun je alle characters in een filename zetten, behalve een `/`. Ik kan iniedergeval de volgende file aanmaken: `!@#$%^&*();[]:'"{},.\\|<>?`. Maar zodra een `/` er in zit, wil dat per definitie de scheiding van een dir zijn.

Qua content-type, gebruik een functie als [`mime_get_type`](https://secure.php.net/manual/en/function.mime-content-type.php) om een content type te bepalen, en vervang gevaarlijke content types als `text/html` en `text/javascript` (let op, kunnen meer types zijn, even uitzoeken!) door `text/plain`. Dan zou HTML als het goed is niet uitgevoerd moeten worden.

kokx:
Overigens is het naar mijn idee prima om bepaalde chars in filenames te disallowen. Zolang dit maar consistent gebeurd.

jszanto:
Symlinks hoeven niet genegeerd te worden, dode symlinks wel. Voor de content types, ik zou `text/html` niet als gevaarlijke content type beschouwen in deze context. Het is een nice to have als de html als text getoond word, maar het kan geen kwaad als het uitgevoerd wordt.

module/Decision/Module.php
```diff
@@ -145,14 +155,21 @@ public function getServiceConfig()
...
                 // and aliases don't work with abstract factories
                 'decision_doctrine_em' => function ($sm) {
                     return $sm->get('doctrine.entitymanager.orm_default');
                 }
-            ]
+            ],
+            /*
+             * Regex pattern matching filenames viewable in the browser
+             */
+            'filebrowser_valid_file' => '[^?*:;{}\\\]*'
```
mstruijs: I wasn't sure if this is the appropriate location for a config constant that should be used in both `Module.php` and `module.config.php` of the same module.

PR #826, opened 2018-03-20
Fix for #813: jobs cannot be made inactive

### LINKED ISSUE #813: Jobs can't be made inactive

Kevinjil:
This fixes #813.

### DIFF
--- a/module/Company/src/Company/Form/FixedKeyDictionaryCollection.php
+++ b/module/Company/src/Company/Form/FixedKeyDictionaryCollection.php
@@ -25,11 +25,6 @@ class FixedKeyDictionaryCollection extends Collection
     {
         $collection = [];
         foreach ($values as $name => $value) {
-            if (array_key_exists("active", $value)) {
-                if ($value["active"] == "0") {
-                    continue;
-                }
-            }
             $element = $this->get($name);
             $collection[$name] = $element instanceof FieldsetInterface ? $element->bindValues($value) : $value;
         }
--- a/module/Company/src/Company/Service/Company.php
+++ b/module/Company/src/Company/Service/Company.php
@@ -498,9 +498,6 @@ class Company extends AbstractACLService
         $id = -1;
 
         foreach ($jobs as $lang => $job) {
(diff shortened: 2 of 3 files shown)


module/Company/src/Company/Service/Company.php
```diff
@@ -482,9 +482,6 @@ public function saveJobData($languageNeutralId, $jobs, $data, $files)
         $id = -1;
 
         foreach ($jobs as $lang => $job) {
-            if ($job->getActive() !== '1') {
-                continue;
-            }
```
wsinnema: Did you check that you can also save a job without English or Dutch translation?

module/Company/src/Company/Form/FixedKeyDictionaryCollection.php
```diff
@@ -25,11 +25,6 @@ public function bindValues(array $values = array())
     {
         $collection = [];
         foreach ($values as $name => $value) {
-            if (array_key_exists("active", $value)) {
-                if ($value["active"] == "0") {
-                    continue;
-                }
-            }
```
wsinnema: Doesn't this break the validation of jobs?

wsinnema:
I also saw that it is possible that both translations are empty and inactive, maybe we should forbid that too.

jszanto:
@Kevinjil Is there an update on this?

Koen1999:
@Kevinjil, if you have time I would love to see this PR finished.

tomudding:
@Koen1999 I have tested this and it works as is (on the test branch). However, when only one language is active for a job, it will state that no job is active. This is because only one of the two jobs is checked. See [edit-package.phtml#L49-L63](https://github.com/Kevinjil/gewisweb/blob/0a8fd99cdd23bd80d3dde0fa88f04200cd530e1f/module/Company/view/company/admin/edit-package.phtml#L49-L63).

tomudding:
@Koen1999 This is now on the testing branch. On its own it works perfectly. However, as you can see on the testing branch, in combination with the two other PRs something goes wrong when creating new jobs.

PR #827, opened 2018-03-27
Removed \r from translations

wsinnema:


kokx:
What problem does this fix?

PR #828, opened 2018-04-05
Improve image quality

kokx:
Currently, image quality on the website is a little off. This is mostly because we convert stuff to PNG. This commit fixes this.

Note that this commit was written while drunk and was not tested at the time of writing. Will test as soon as I think about this.

Why we want to do this? Look at https://gewis.nl/photo/view/121342, and note the MWM logo, you see some kind of gradient in the logo here. Now look at the source image https://gewis.nl/photo/download/121342 where we do not see such gradient.

Also, note that serving files as PNG while they are originally JPEG, will actually transfer a lot of data which we do not need. PNG is a horrible format when it comes to serving photos, in which case lossy compression is nice (PNG is great however for things where we need lossless compression.).

### DIFF
--- a/module/Photo/src/Photo/Service/Admin.php
+++ b/module/Photo/src/Photo/Service/Admin.php
@@ -90,9 +90,10 @@ class Admin extends AbstractAclService
     {
         $image = new Imagick($path);
         $image->thumbnailImage($width, $height, true);
-        $image->setimageformat("png");
+        $image->setimageformat("jpeg");
+        $image->setImageCompressionQuality(90);
         //Tempfile is used to generate sha1, not sure this is the best method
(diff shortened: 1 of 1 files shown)


jszanto:
I think it would be better if we wouldn't resize the images during upload and do it on the fly instead. This simplifies things, allows more flexibility in the design and improves the quality of previously uploaded images as well. I can recommend using [Glide](https://github.com/thephpleague/glide) to achieve this.

kokx:
In that case, I think that it might be better to use a dedicated image server, like thumbor. Which also improves images on the fly, but also caches them. Which saves some server load.

jszanto:
Glide also has caching built in, so the performance should be about the same as Thumbor. I think Glide would be easier to implement in our current setup.

jszanto:
Closing this for now, since I believe we decided we will solve this in another way.

PR #830, opened 2018-04-13
Send email address along with JWT token

sqlr:


kokx:
I just checked this, and third parties cannot obtain email addresses from this. It is important, however, that the email addresses and membership numbers are handled with care by the external parties that receive them.

PR #831, opened 2018-04-26
List jobs and categories only once in the admin interface

### LINKED ISSUE #815: When a job has multiple languages it is shown twice in the company admin

wsinnema:
Fixes #815

### DIFF
--- a/module/Company/src/Company/Service/Company.php
+++ b/module/Company/src/Company/Service/Company.php
@@ -151,6 +151,19 @@ class Company extends AbstractACLService
         }
         return $nonemptyCategories;
     }
+
+    private function getUniqueInArray($array, $callback) {
+        $tempResults = [];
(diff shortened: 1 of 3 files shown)


module/Company/src/Company/Service/Company.php
```diff
@@ -160,7 +173,10 @@ public function getCategoryList($visible)
                     $translator->translate('You are not allowed to access the admin interface')
                 );
             }
-            return $this->getCategoryMapper()->findAll();
+            $results = $this->getCategoryMapper()->findAll();
```
jszanto: Why not just add a `distinct` clause to the query here? Wouldn't that solve all of this in 1 line of code?

kokx:
This looks good and is approved. Why isn't this merged yet @wsinnema @Kevinjil ?

Issue #800, opened 2018-01-19
Exception logging


jszanto:
Log exceptions properly somewhere in a log file.

Koen1999:
Can we close this issue? #832 is already merged.

PR #832, opened 2018-05-14
Implemented error logging

Kevinjil:
Added Monolog and implemented event callback to log errors to a file.
This would implement #800.


Kevinjil:
There is some redundant logging (like insufficient permission for an action) which is not very useful for us, cleanup in https://github.com/Kevinjil/gewisweb/commit/55f41796f97a3fdb794a5aeb985f6b422eb460af to make the logs more readable

PR #845, opened 2018-11-06
added <hr> above pagination

Koen1999:


module/Frontpage/view/frontpage/poll/history.phtml
```diff
@@ -24,7 +24,8 @@ $this->headTitle($this->translate('Poll history')); ?>
                     'viewLink' => true
                 ]) ?>
             <?php endforeach; ?>
-        <?php endif; ?>
+        <?php endif; ?>0
```
jszanto:
```suggestion
        <?php endif; ?>
```

jszanto:
@Koen1999 If you remove that zero, then we can merge this, since it has been open for a while now

PR #849, opened 2018-11-30
Additions for nginx authentication

wsinnema:
Added route that can check whether the user is logged in, and added a
redirect query parameter to the login route. These are needed to enable proper authentication on the new web server.

### DIFF
--- a/module/User/config/module.config.php
+++ b/module/User/config/module.config.php
@@ -137,7 +137,7 @@ return [
                 ],
                 'priority' => 100
             ],
-            'test_login' => [
+            'validate_login' => [
                 'type' => 'segment',
                 'options' => [
                     'route' => '/api/validateLogin',
--- a/module/User/config/module.config.php
+++ b/module/User/config/module.config.php
@@ -136,12 +136,24 @@ return [
                     ]
                 ],
                 'priority' => 100
-            ]
+            ],
+            'test_login' => [
+                'type' => 'segment',
+                'options' => [
+                    'route' => '/api/validateLogin',
+                    'defaults' => [
+                        'controller' => '\User\Controller\Api',
+                        'action' => 'validate'
+                    ],
+                ],
+                'priority' => 100,
+            ],
         ],
     ],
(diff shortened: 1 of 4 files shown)


wsinnema:
This PR adds two features. 
1. It adds a route /api/validateLogin that returns a 401 if the user is not logged in, and a 200 otherwise. 
2. It adds a  GET parameter to the login page so the user can be redirected to another page after login

The final flow will be as follows:
1. The visitor visits a protected page (which is not part of the GEWIS site)
2. Nginx calls the /api/validateLogin url 
3. If it returns a 401, the user is redirected to the login page
4. If it returns a 200, the correct page is shown
5. After login, the user is redirected to the protected page

kokx:
The real question still stands: Why is this needed exactly?

wsinnema:
This is needed to use the GEWIS website authentication to protect other pages (for example, committee pages)

kokx:
How exactly will that be done? And is it just authentication, or will some information be attached as well? (due to GDPR, we need to be careful about sharing with committees)

Koen1999:
Okay. So, I started testing this as I see the need for this, as convinced by CBC.
I don't see any sensitive information being handed over.

Yet, this PR introduces a new bug.
When you visit a restricted page, you are forwarded to the login page. After logging in, you are redirected to gewis.nl/user instead of the requested page. This results in status code 500.

kokx:
Okay, so I'm getting really bothered by what the real need would be for this. Please DO tell us what it is, instead of hinting that it is important somehow.

jszanto:
@Koen1999, I can't seem to reproduce this, the redirect from the restricted page should include a redirect uri such that it redirects properly. Have you configured your web server correctly for testing?

Koen1999:
> 
> 
> Okay, so I'm getting really bothered by what the real need would be for this. Please DO tell us what it is, instead of hinting that it is important somehow.

The idea is that when migrating to the new server, a central point is used for authenticating.
This is basically the start of that.

PR #853, opened 2019-02-03
Added profile picture

Koen1999:
I have added a profile picture to public and self profile of members. Determination of the profile picture is done automatically based on the number of hits, number of tags and age of a picture the member is tagged in.

This is an example of a public profile with a beautiful picture.
image

This picture was selected with the following formula: 
Base rating = Amount of hits divided by the number of tags
Final Rating = Base Rating + Base rating / age
Age is determined in days

The idea is that the newer pictures are preferred over older ones; that profiles of a single or a few persons are preferred over group pictures; and that good profile pictures are obviously viewed often.

### DIFF
--- a/module/Photo/src/Photo/Model/Photo.php
+++ b/module/Photo/src/Photo/Model/Photo.php
@@ -320,6 +320,14 @@ class Photo implements ResourceInterface
     }
 
     /**
+     * @return \Photo\Model\Hit
+     */
+    public function getHits()
+    {
+        return $this->hits;
+    }
+
+    /**
      * @return \Photo\Model\WeeklyPhoto|null
      */
     public function getWeeklyPhoto()
--- a/module/Photo/src/Photo/Model/Photo.php
+++ b/module/Photo/src/Photo/Model/Photo.php
@@ -320,11 +320,19 @@ class Photo implements ResourceInterface
     }
 
     /**
-     * @return \Photo\Model\Hit
+     * @return int
      */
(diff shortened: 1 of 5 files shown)


module/Photo/src/Photo/Model/Photo.php
```diff
@@ -319,6 +319,14 @@ public function getTags()
         return $this->tags;
     }
 
+    /**
+     * @return \Photo\Model\Hit
+     */
+    public function getHits()
```
jszanto: If you're only interested in the hitcount you can turn this in to `getHitCount()` and return `$this->hits->count()`. If I recall correctly that is a lot faster than retrieving all hits.

jszanto:
Codeclimate has some valid points about your code, please check those

Koen1999:
> 
> 
> Interesting feature, perhaps it would be nice if (at some point) you could override the generated profile image with one which you can choose yourself. Then we can use these profile images on other pages such as the committee page.

This was really just an easy, but nice extension for the website we have. What you suggest however, would also be a nice addition, but would need slightly more effort.

Koen1999:
I have fixed the codeclimate issues, however, I haven't tested the changes yet. 

I must also note that we might need to change the formula for determining the profile picture, but we can do that anytime, even after release.

Koen1999:
Changes are tested, it works as expected.
I suggest we merge this feature after it's approved.
Further additions or extensions can always be done at a later moment.

Issue #293, opened 2015-12-09
Polls: Don't limit answers to 75 characters


jszanto:

kokx:
Do limit it to something sensible

Koen1999:
What would be considered a sensible length?

jszanto:
Somewhere around 128 characters seems sensible for a poll answer

PR #856, opened 2019-02-03
Changed length validators for polls

Koen1999:
This is a fix for issue #293


module/Frontpage/src/Frontpage/Form/PollComment.php
```diff
@@ -55,8 +55,8 @@ public function getInputFilterSpecification()
                     [
                         'name' => 'string_length',
                         'options' => [
-                            'min' => 3,
-                            'max' => 128
+                            'min' => 2,
+                            'max' => 32
                         ]
```
jszanto: 32 characters seems very short for a poll comment?

Koen1999: This is the author name. I don't think we need long fields for that

PR #859, opened 2019-02-03
My activities

### LINKED ISSUE #725: Display activities which the user is signed up for

Koen1999:
This resolves #725 .

Added a page for viewing member activities.
These include: activities signed up for, activities created by self, activities created by organs member is part of
The latter two include activities that are not open for subscription yet.

It looks as follows: 
image

image

### DIFF
--- a/module/Activity/src/Activity/Mapper/Activity.php
+++ b/module/Activity/src/Activity/Mapper/Activity.php
@@ -129,11 +129,11 @@ class Activity
     }
 
     /**
-     * Get a query to retrieve upcoming activities sorted by date that a user is subscribed to
+     * Get upcoming activities sorted by date that a user is subscribed to
      *
      * @param \User\Model\User $user Option user that should relate to activity
      *
-     * @return \Zend\Db\Sql
+     * @return array
      */
     public function getUpcomingActivitiesSubscribedBy($user) {
         $qb = $this->em->createQueryBuilder();
@@ -147,16 +147,16 @@ class Activity
             ->andWhere('a = b.activity')
             ->andWhere('b.user = :user')
             ->setParameter('user', $user);
-        $query = $qb->getQuery();
-        return $query;
+        $result = $qb->getQuery()->getResult();
+        return $result;
     }
 
     /**
-     * Get a query to retrieve upcoming activities sorted by date that a user created
+     * Get upcoming activities sorted by date that a user created
      *
      * @param \User\Model\User $user Option user that should relate to activity
      *
-     * @return \Zend\Db\Sql
+     * @return array
      */
     public function getUpcomingActivitiesCreatedBy($user) {
         $qb = $this->em->createQueryBuilder();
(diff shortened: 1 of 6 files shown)


Koen1999:
It's all tested and well, I think that the remaining codeclimate issues may be ignored.

module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -218,7 +221,11 @@ public function getUpcomingActivities($category = null)
         }
 
         $activityMapper = $this->getServiceManager()->get('activity_mapper_activity');
-        $activity = $activityMapper->getUpcomingActivities(null, null, $category);
+        if ($category === 'my') {
+            $activity = $activityMapper->getUpcomingActivitiesForMember($user);
+        } else {
+            $activity = $activityMapper->getUpcomingActivities(null, null, $category);
+        }
```
jszanto:
```suggestion
if ($category === 'my') {
            return $activityMapper->getUpcomingActivitiesForMember($user);
        }
        return $activityMapper->getUpcomingActivities(null, null, $category);
```

module/Activity/src/Activity/Mapper/Activity.php
```diff
@@ -94,6 +94,118 @@ public function getUpcomingActivities($count = null, $organ = null, $category =
...
+        // Get created by member (including non-approved)
+        $result = array_merge($result, $this->getUpcomingActivitiesCreatedBy($user));
+
+        // Get associated with organs (including non-approved)
+        foreach ($user->getMember()->getCurrentOrganInstallations() as $organ) {
+            $result = array_merge($result, $this->getUpcomingActivitiesByOrgan($organ));
+        }
+
+        // Do sorting based on start time
+        $result = $this->sortMultiArrayByKey($result, 'getBeginTime');
```
jszanto: No need to write a custom sort function, you can use `usort`, something like this should work:

```suggestion
        $result = usort($result, function($a, $b) {
            return $a->getBeginTime() < $b->getBeginTime() ? -1 : 1;
        });
```

Koen1999: Thanks for the tip!

Koen1999:
This pull request is now ready to be merged.

module/Activity/src/Activity/Mapper/Activity.php
```diff
@@ -94,6 +94,98 @@ public function getUpcomingActivities($count = null, $organ = null, $category =
...
+        foreach ($user->getMember()->getCurrentOrganInstallations() as $organ) {
+            $result = array_merge($result, $this->getUpcomingActivitiesByOrgan($organ));
+        }
+
+        // Do sorting based on start time
+        usort($result, function($a, $b) {
+            return $a->getBeginTime() < $b->getBeginTime() ? -1 : 1;
+        });
+
+        return $result;
```
Kevinjil: Result can contain duplicate activities if e.g. you subscribed to an activity you have created yourself.

module/Activity/src/Activity/Mapper/Activity.php
```diff
@@ -113,6 +113,8 @@ public function getUpcomingActivitiesForMember($user)
         foreach ($user->getMember()->getCurrentOrganInstallations() as $organ) {
             $result = array_merge($result, $this->getUpcomingActivitiesByOrgan($organ));
         }
+        
+        $result = array_unique($result);
```
Kevinjil: Unfortunately, this will throw a fatal _"Object of class Activity\Model\Activity could not be converted to string"_ error ([ref](http://php.net/manual/en/function.array-unique.php))

Koen1999: I have resolved the issue, but haven't tested it yet.

Koen1999: Tested as well now

Issue #834, opened 2018-06-11
Sort GM pieces by agenda item number


raketwissenschaftler:
On the request of Wout:
Now, the pices for the GM are sorted randomly, see attachement. Ordering them by agenda item would be nicer.
![image](https://user-images.githubusercontent.com/6400064/41216798-1cf42b66-6d56-11e8-9340-d09734b30d9a.png)

Kevinjil:
Solved by #861

PR #861, opened 2019-02-05
Ensured documents for meetings are sorted by number

dvanleijenhorst:
Solves issue #834. 

Unnumbered items are currently sorted alphabetically underneath the numbered items, and numbered items are sorted as expected.

### DIFF
--- a/module/Decision/src/Decision/Model/Meeting.php
+++ b/module/Decision/src/Decision/Model/Meeting.php
@@ -225,15 +225,12 @@ class Meeting
         usort($temp, function ($a, $b) {
             $aa = preg_split("/(\.|\s)/", $a->getName());
             $bb = preg_split("/(\.|\s)/", $b->getName());
-            if (!is_numeric($aa[0]) || !is_numeric($bb[0])) {
-                return strcmp($aa[0], $bb[0]);
-            }
             for ($i = 0; $i < min(count($aa), count($bb)); $i++) {
                 if (!is_numeric($aa[$i])) {
                     return -1;
-                } else if (!is_numeric($bb[$i])) {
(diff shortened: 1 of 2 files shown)


dvanleijenhorst:
As a side-effect of trying to fix codeclimate issues, unnumbered items now appear sorted above the numbered items, and numbered items are as normal

Koen1999:
This resolves #834

module/Decision/src/Decision/Model/Meeting.php
```diff
@@ -214,4 +214,29 @@ public function addDocuments($documents)
...
+            $aa = preg_split("/(\.|\s)/", $a->getName());
+            $bb = preg_split("/(\.|\s)/", $b->getName());
+            for ($i = 0; $i < min(count($aa), count($bb)); $i++) {
+                if (!is_numeric($aa[$i])) {
+                    return -1;
+                } elseif (!is_numeric($bb[$i])) {
+                    return 1;
+                } elseif ($aa[$i] != $bb[$i]) {
+                    return $aa[$i] < $bb[$i] ? -1 : 1;
+                }
```
kokx: Using `else` generally decreases readability in methods. Instead, I would rewrite it into:

```php
if (!is_numeric($aa[$i])) {
    return -1;
}
if (!is_numeric($bb[$i])) {
    return 1;
}
if ($aa[i] != $bb[i]) {
    return $aa[$i] < $bb[i]$ ? -1 : 1;
}
```

Also. This function seems like it's basically lexicographic sort on document names. Would sort on `$doc->getName()` not be good enough?

jszanto: That would break when there are more than 9 documents
1
10
11
12
2
3
4
5
6

PR #862, opened 2019-02-05
Implemented check for amount of authorizations before authorizing

Koen1999:
This resolves an issue that is not on Github


Koen1999:
The feature has already been well-tested.

PR #865, opened 2019-02-13
Fix Masonry Position Calculation in Firefox

### LINKED ISSUE #864: Incorrect photo-grid-item position calculations in Firefox
The positions of `.photo-grid-item`s are not correctly calculated in Firefox (tested in >FF64). This results in a single column of pictures instead of multiple columns.

An example of this is visible on any page including multiple images (e.g. a member's photo page):
gewis ff masonry incorrect calculations

The DOM (after going through Masonry):
gewis ff masonry incorrect calculations code

tomudding:
Fixes #864.

Adds an additional `left`-property to `.grid-sizer`. This allows Masonry to correctly calculate the width for `.photo-grid-item`s. This additional property is needed because Firefox recently (~7 months ago) changed a method which Masonry used to calculate the widths (represented in the `left`-property) and there has not been any response from Masonry regarding this issue (except for use this property).

Images are now correctly aligned in a grid:
gewis ff masonry incorrect calculations fixed

### DIFF
--- a/public/scss/modules/_photo.scss
+++ b/public/scss/modules/_photo.scss
@@ -11,6 +11,10 @@ Photo
   clear: both;
 }
 
+.grid-sizer {
+  float: left;
+}
+
 .gutter-sizer {
   width: 0.5%;
 }


tomudding:
This should not be merged yet as it appears to break down on devices with small screens (such as mobile phones). An additional fix to address this will (hopefully) be implemented today.

kokx:
Would an update of masonry not be a better fix? Though that is based on the assumption that newer versions of masonry have fixed this.

tomudding:
@kokx Yes, updating Masonry would indeed be the best solution. However, as far as I can tell Masonry has not received an update that fixes this issue. The latest official release of Masonry was on July 4, 2018 (v4.2.2) and the first time the issue was reported (see [Masonry/1057](https://github.com/desandro/masonry/issues/1057)) was on July 18, 2018. In that issue the only fix mentioned is to add the additional `left`-property.

An update of Masonry is still reasonable because we currently use v4.0.0, which is a bit more than 3 years old. But there are more libraries that need to be updated as some point, so maybe bundle them together and update them all at once (if possible).

PR #870, opened 2019-04-02
Made profile picture selection possible + caching

### LINKED ISSUE #868: Make selection of profile pictures possible

### LINKED ISSUE #869: Cache generated profile pictures

Koen1999:
Profile pictures are now cached:
- 1 day for active members
- 5 days for non-active members

Furthermore, members can select a profile photo of their own, if they are tagged in it. Of course they can also always revert this to automatic selection or choose another picture.

This resolves #868 and  this resolves #869.

### DIFF
--- a/module/Frontpage/src/Frontpage/Form/Page.php
+++ b/module/Frontpage/src/Frontpage/Form/Page.php
@@ -3,7 +3,6 @@
 namespace Frontpage\Form;
 
 use Zend\Form\Form;
-use Zend\InputFilter\InputFilter;
 use Zend\InputFilter\InputFilterProviderInterface;
 use Zend\I18n\Translator\TranslatorInterface as Translator;
 
@@ -110,7 +109,7 @@ class Page extends Form implements InputFilterProviderInterface
                     [
                         'name' => 'string_length',
                         'options' => [
-                            'min' => 3,
+                            'min' => 2,
                             'max' => 25
                         ]
                     ],


module/Decision/src/Decision/Model/Member.php
```diff
@@ -559,6 +559,18 @@ public function getCurrentOrganInstallations()
...
 
+    /**
+     * Returns whether the member is currently part of any organs
+     *
+     * @return bool
+     */
+    public function isActive()
+    {
+        $installations = $this->getCurrentOrganInstallations();
+        $bool = !empty($installations);
```
jszanto: why not just `return !empty($installation);` 😜

module/Decision/view/decision/member/self.phtml
```diff
@@ -147,6 +147,12 @@ $this->headTitle($member->getFullName());
                               </a>
                             </div>
                     </div>
+                    <?php if ($isExplicitProfilePhoto): ?>
+                        <a type="button" class="btn btn-primary" href=<?= $this->url('photo/removeprofilephoto', ['photo_id' => 0]); ?>>
```
jszanto: `btn-danger` would be more suitable perhaps

module/Photo/config/module.config.php
```diff
@@ -145,6 +145,32 @@
                             ],
                         ],
                     ],
+                    'setprofilephoto' => [
```
jszanto: I believe `set_profile_photo` would be a better name for this if you want it to match with our existing naming scheme.

module/Photo/config/module.config.php
```diff
@@ -145,6 +145,32 @@
...
+                            'defaults' => [
+                                'controller' => 'Photo',
+                                'action' => 'setProfilePhoto',
+                            ],
+                        ],
+                    ],
+                    'removeprofilephoto' => [
+                        'type' => 'Segment',
+                        'options' => [
+                            'route' => '/removeprofilephoto/:photo_id',
```
jszanto: You can make photo_id an optional route param here for nicer routes. Then you don't need to set it to `0`.

jszanto: I believe by changing this to `'/removeprofilephoto[/:photo_id]'`, but not sure, since I haven't touched zend framework in a while.

PR #872, opened 2019-04-30
Added ability to add documents for VV's

kokx:


module/Decision/src/Decision/Mapper/Meeting.php
```diff
@@ -112,17 +112,17 @@ public function findPast($limit = null, $type = null)
      */
     public function findLatestAV()
```
Kevinjil: Shouldn't this function name also be updated for consistency?

Issue #884, opened 2019-05-08
Hide internal activities on activity overviews if not logged in


Koen1999:

mstruijs:
This is mostly caused by the fact that non-logged in clients have permission for all activities, see issue #885

Koen1999:
I will check with the board what exactly is the desired behavior.

Koen1999:
The current behaviour is fine, so no changes are required

PR #887, opened 2019-05-08
Made activity homepage visible again for guests

### LINKED ISSUE #878: Remove login wall activities page
https://gewis.nl/activity is currently begind a login wall, this has to be removed in order to allow for external participants.

### LINKED ISSUE #885: Non-user cannot view activity even when allowed
PR #859 introduced a requirement in this [commit](https://github.com/GEWIS/gewisweb/commit/2d92f4a950a566f229dbd8e247b2a346a6353a3e#diff-98000bbef9461386cb74746e7b66c51b) for viewing an activity to be done by a user.

While most activities should be invisible (or should they?), this should be handled properly using the permissions via in the acl, as defined in [the module file](https://github.com/GEWIS/gewisweb/blob/master/module/Activity/Module.php). Related is #884, which may be automatically resolved by this issue.

Koen1999:
This resolves #885 and resolves #878.
The identity of the user will now only requested when trying to view the My Activities page.
All activities can now be viewed by external members, just not the subscription lists.

I will await issue #884 before making additional changes.
This PR may be merged.

### DIFF
--- a/module/Activity/src/Activity/Service/ActivityQuery.php
+++ b/module/Activity/src/Activity/Service/ActivityQuery.php
@@ -207,11 +207,10 @@ class ActivityQuery extends AbstractAclService implements ServiceManagerAwareInt
      * Get all activities that are approved by the board and which occur in the future
      *
      * @param String $category Type of activities requested
-     * @param Decision/Model/Member $user User for which activities are selected if category === 'my'
      *
      * @return array Array of activities
      */
-    public function getUpcomingActivities($category = null, $user = null)
+    public function getUpcomingActivities($category = null)
     {
         if (!$this->isAllowed('view', 'activity')) {
             $translator = $this->getTranslator();
@@ -222,6 +221,7 @@ class ActivityQuery extends AbstractAclService implements ServiceManagerAwareInt
 
         $activityMapper = $this->getServiceManager()->get('activity_mapper_activity');
         if ($category === 'my') {
+            $user = $this->getServiceLocator()->get('user_service_user')->getIdentity();
             return $activityMapper->getUpcomingActivitiesForMember($user);
         }
         return $activityMapper->getUpcomingActivities(null, null, $category);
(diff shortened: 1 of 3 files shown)


module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -222,6 +221,7 @@ public function getUpcomingActivities($category = null, $user = null)
 
         $activityMapper = $this->getServiceManager()->get('activity_mapper_activity');
         if ($category === 'my') {
```
mstruijs: I think it is better to add some explicit acl checks somewhere such that `getIdentity()` is not called for non-users. The behaviour of `getIdentity()` isn't exactly obvious in that case, so more explicit checks can prevent future mistakes.

module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -222,6 +221,7 @@ public function getUpcomingActivities($category = null, $user = null)
...
+            $user = $this->getServiceLocator()->get('user_service_user')->getIdentity();
             return $activityMapper->getUpcomingActivitiesForMember($user);
```
mstruijs: It would make more sense to directly call `getUpcomingActivitiesForMember` from the controller, especially since getting the user identity is usually done at the controller level.

module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -222,6 +221,13 @@ public function getUpcomingActivities($category = null, $user = null)
 
         $activityMapper = $this->getServiceManager()->get('activity_mapper_activity');
         if ($category === 'my') {
+            if (!$this->getRole() === 'guest') {
```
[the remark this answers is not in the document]
Koen1999: Fixed this

module/Activity/src/Activity/Service/ActivityQuery.php
Koen1999: I would instead argue that it should be part of the Service level.

Issue #251, opened 2015-11-22
Photo: Save button in generating album cover doesn't actually save


kokx:
The save button when generating album covers doesn't actually save. When generating a new cover, the new cover is already applied on the user side. And neither is there a way to cancel and go back to the old cover. Thus, this button should either be named somethink like 'OK'. Or actual save functionality should be in place.

Koen1999:
Just change text of save to 'Close'

PR #900, opened 2019-06-05
Changed the album cover save to close

kokx:
Note: translation still needs to be fixed

Should be a fix for issue #251


Koen1999:
This indeed fixes #251

PR #905, opened 2019-06-18
Hide external subscribers from public view on archived activities

### LINKED ISSUE #904: Put subscription lists of past activities behind login wall
Currently, past activities that allowed for external subscriptions show their subscription list to anyone (including non-members and search engines).
This is undesired behavior.

tomudding:
Hides external subscribers from the subscription list of an archived activity when viewed by non-members and other external viewers.

Fixes #904.


Koen1999:
afbeelding
Ik denk dat je het if-statement incorrect hebt geplaatst

PR #908, opened 2019-06-20
Implemented HTTP Status Code 418 according to RFC2324

Koen1999:
afbeelding

### DIFF
--- a/module/Application/config/module.config.php
+++ b/module/Application/config/module.config.php
@@ -23,6 +23,18 @@ return [
                 ],
                 'priority' => 100
             ],
+            'teapot' => [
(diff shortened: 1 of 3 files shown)


module/Application/config/module.config.php
```diff
@@ -23,6 +23,18 @@
                 ],
                 'priority' => 100
             ],
+            'teapot' => [
+                'type' => 'Literal',
+                'options' => [
+                    'route' => '/teapot',
```
jszanto: The correct route for this would be `/coffee` I believe

Koen1999: You're entirely right

PR #915, opened 2019-08-24
Fix issue where email addresses are case sensitive

### LINKED ISSUE #758: Case-sensitive email addresses during registration and activation
Some guy had a problem with activating his account during the Boekenborrel. It seems that he has registered his email address with a capital letter in it, During activation he forgot that he did that and this caused problems.

Koen1999:
This is a continuation of #771 
This fixes #758 by the way


Koen1999:
This fixes #758 by the way

PR #921, opened 2019-09-10
Made adding options on behalf of non-organ possible

### LINKED ISSUE #911: Allow adding options organised by the board (and initiatives)

Koen1999:
This fixes #911 

Still have to test it, will do so later this week

### DIFF
--- a/module/Activity/src/Activity/Service/ActivityCalendar.php
+++ b/module/Activity/src/Activity/Service/ActivityCalendar.php
@@ -148,7 +148,13 @@ class ActivityCalendar extends AbstractAclService
         $proposal->setName($name);
         $description = $validatedData['description'];
         $proposal->setDescription($description);
-        $proposal->setOrgan($this->sm->get('decision_service_organ')->getOrgan($organ));
+        if ($organ > -1) {
+            $proposal->setOrgan($this->sm->get('decision_service_organ')->getOrgan($organ));
+        } elseif ($organ == -1) {
+            $proposal->setOrganAlt("Board");
+        } elseif ($organ == -2) {
+            $proposal->setOrganAlt("Other");
+        }
         $em->persist($proposal);
         $em->flush();
 
@@ -380,7 +386,7 @@ class ActivityCalendar extends AbstractAclService
             return true;
         }
 
-        if ($this->getOrganService()->canEditOrgan($option->getProposal()->getOrgan())) {
+        if ($option->getProposal()->getOrgan() !== null && $this->getOrganService()->canEditOrgan($option->getProposal()->getOrgan())) {
             return true;
         }
 
--- a/module/Activity/src/Activity/Service/ActivityCalendar.php
+++ b/module/Activity/src/Activity/Service/ActivityCalendar.php
@@ -148,6 +148,7 @@ class ActivityCalendar extends AbstractAclService
         $proposal->setName($name);
         $description = $validatedData['description'];
         $proposal->setDescription($description);
+//        See /Activity/Form/ActivityCalendarProposal for more details on the definition of these options
         if ($organ > -1) {
             $proposal->setOrgan($this->sm->get('decision_service_organ')->getOrgan($organ));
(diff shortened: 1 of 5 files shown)


Koen1999:
This has been tested and is currently deployed on test.gewis.nl

Will merge after an approving review

module/Activity/view/email/options-overdue.phtml
```diff
@@ -14,7 +14,16 @@ Er zijn een aantal opties ouder dan drie weken,<br />
...
-            <td><?= $option->getProposal()->getOrgan() === null ? $option->getProposal()->getCreator()->getMember()->getFullName() : $option->getProposal()->getOrgan()->getAbbr() ?></td>
+            <?php
+            if ($option->getProposal()->getOrgan() !== null) {
+                $creator = $option->getProposal()->getOrgan()->getAbbr();
+            } elseif ($option->getProposal()->getOrganAlt() !== null) {
+                $creator = $option->getProposal()->getOrganAlt();
+            } else {
+                $creator = $option->getProposal()->getCreator()->getMember()->getFullName();
+            }
+            ?>
```
armandduijn: Consider refactoring this block of conditions. This block is also used a few lines down on line 52-59. A possible solution is to define a new method in the `ActivityOptionProposal` model which returns the creator.

```php
public function getCreator(): string {
  // Logic to determine the proposal was created by an organ, alt, or a member
  // Whom ever the creator is, always return a string (predictable)
}
```

Comment also applicable to the file `module/Activity/view/activity/activity-calendar/index.phtml`

module/Activity/src/Activity/Service/ActivityCalendar.php
```diff
@@ -148,7 +148,13 @@ public function createProposal($data)
...
         $description = $validatedData['description'];
         $proposal->setDescription($description);
-        $proposal->setOrgan($this->sm->get('decision_service_organ')->getOrgan($organ));
+        if ($organ > -1) {
+            $proposal->setOrgan($this->sm->get('decision_service_organ')->getOrgan($organ));
+        } elseif ($organ == -1) {
+            $proposal->setOrganAlt("Board");
+        } elseif ($organ == -2) {
+            $proposal->setOrganAlt("Other");
+        }
```
armandduijn: Please add a comment about the meaning or reason for the magic numbers -1 and -2. I checked and I think they're first defined in `module/Activity/src/Activity/Form/ActivityCalendarProposal.php` on line 33-34. Add this so future maintainers know where they appeared.

armandduijn:
Note that I haven't properly tested it on test.gewis.nl. I don't have board permissions and I prefer not to fiddle around with the database. Futhermore, I don't see a way to add a member initiative, i.e. without specifying an organ (not sure if that's part of this PR though).

module/Activity/view/email/options-overdue.phtml
armandduijn: Alternatively, you can put the refactored code in a service. A possible candidate is the ` ActivityCalendar`. Putting the code in a service instead of the model might be better practice in order to keep your models slim (no business logic). Not entirely sure which is best because I'm not very familiar with Zend framework's architecture.

module/Activity/src/Activity/Service/ActivityCalendar.php
Koen1999: Well, all organs have positive numbers for their IDs. I figured using some negative numbers would be useful for the non-organ options.

module/Activity/view/email/options-overdue.phtml
Koen1999: As the functionality really only relates to the model, I added a function to the model.

Koen1999:
> 
> 
> Note that I haven't properly tested it on test.gewis.nl. I don't have board permissions and I prefer not to fiddle around with the database. Futhermore, I don't see a way to add a member initiative, i.e. without specifying an organ (not sure if that's part of this PR though).

Only board members and other admins can use the alternative options.

PR #931, opened 2019-10-03
Improved user registration

### LINKED ISSUE #922: Improve user registration
Currently if during registration on the website, the wrong email address is entered. It's not possible to register again as there already is an entry in the NewUser table.
This results in a PDOException at the moment.

A good solution that would allow registering with the correct email address again, without intervention of the web committee would be as follows:
Check if entry in NewUser exists for given membership ID, if so remove and recreate this entry correctly.
Resend registration email.

Koen1999:
This fixes #922

### DIFF
--- a/module/User/src/User/Form/Register.php
+++ b/module/User/src/User/Form/Register.php
@@ -69,6 +69,13 @@ class Register extends Form
                 ]
             ]);
             break;
+        case self::ERROR_ALREADY_REGISTERED:
+            $this->setMessages([
+                'lidnr' => [
+                    $this->translate->translate("You already attempted to register, please check your email or try again after 24 hours.")
+                ]
+            ]);
+            break;
         case self::ERROR_USER_ALREADY_EXISTS:
             $this->setMessages([
                 'lidnr' => [
--- a/module/User/src/User/Model/NewUser.php
+++ b/module/User/src/User/Model/NewUser.php
@@ -2,10 +2,10 @@
 
 namespace User\Model;
 
+use DateTime;
 use Decision\Model\Member;
-
-use Doctrine\ORM\Mapping as ORM;
 use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\ORM\Mapping as ORM;
 
 /**
  * User model.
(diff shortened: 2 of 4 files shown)


module/User/src/User/Model/NewUser.php
```diff
@@ -72,6 +79,16 @@ public function getLidnr()
...
+    /**
+     * Set the user's membership number.
+     *
+     * @param int $lidnr
+     */
+    public function setLidnr($lidnr)
+    {
+        $this->lidnr = $lidnr;
+    }
+
```
kokx: Why all this useless rearrangement of methods? It completely obscures whatever you're actually changing.

Koen1999:
I will attempt to undo the automatic code cleanup later this week so you'll be able to review it.

Koen1999:
> 
> 
> I will attempt to undo the automatic code cleanup later this week so you'll be able to review it.

This became slightly later than expected, but I undid the reformatting of the code.

module/User/src/User/Service/User.php
```diff
@@ -98,9 +98,19 @@ public function register($data)
             return null;
         }
 
+        $newUser = $this->getNewUserMapper()->getByLidnr($data['lidnr']);
+        if (is_null($newUser)) {
+            $time = $newUser->getTime();
+            $requiredInterval = (new DateTime())->sub(new DateInterval('P1D'));
+            if ($time > $requiredInterval) {
+                $form->setError(RegisterForm::)
```
kokx: This seems a bit weird? And with syntax errors even?

Koen1999: Apparently I forgot to fill it in after creating the error.

module/User/src/User/Form/Register.php
```diff
@@ -69,6 +69,13 @@ public function setError($error)
                 ]
             ]);
             break;
+        case self::ERROR_ALREADY_REGISTERED:
```
Koen1999: I know indentation is incorrect, but at least it's consistent with the other cases.

PR #932, opened 2019-10-03
Automatic code cleanup attempt

Koen1999:


Koen1999:
Is this something that we might like to do?

module/Application/Module.php
```diff
@@ -9,6 +9,12 @@
 
 namespace Application;
 
+use Application\View\Helper\Acl;
+use Application\View\Helper\FileUrl;
+use Application\View\Helper\Infima;
+use Application\View\Helper\JobCategories;
+use Application\View\Helper\ModuleIsActive;
+use Application\View\Helper\ScriptUrl;
```
jszanto: I'm not really a fan of adding all these imports when each class is only used once here. In my opinion it even makes the code slightly less readable.

jszanto:
This is a good idea, however this will probably cause a lot of merge conflicts with the other currently open PR's. Furthermore if we do such a thing we probably should fix some more of the issues at the same time and make sure that everything is formatted according to PSR-12

module/Application/Module.php
Koen1999: I personally think it makes it more readable. But I guess it's rather subjective.

PR #934, opened 2019-10-04
Separate album children from album photos

### LINKED ISSUE #926: Sub-albums cause overlap with other sub-albums or album photos
Sub-albums tend to randomly overlap each other or photos overlap sub-albums. This is caused by an incorrectly calculated `top`-property on sub-albums and pictures.

**An example where photos overlap sub-albums (sub-album in red and photo in blue):**
GEWIS-WEB Sub-album overlap

tomudding:
Separates album children from album photos. This is required to prevent children and photos overlapping children.

Fixes #926.

**Old Viewer (incorrect):**
GEWIS-WEB Sub-album overlap old viewer

**New Viewer (incorect):**
GEWIS-WEB Sub-album overlap new viewer

**New Viewer (correct):**
GEWIS-WEB Sub-album no overlap

### DIFF
--- a/module/Photo/view/photo/album/index-new.phtml
+++ b/module/Photo/view/photo/album/index-new.phtml
@@ -99,21 +99,23 @@ $this->scriptUrl()->requireUrl('member/search')
             $path = $this->url('photo/album', ['album_id' => $album->getId()]);
         }
         ?>
-
-        <div class="photo-grid">
-            <div class="grid-sizer"></div>
-            <div class="gutter-sizer"></div>
-            <?php foreach ($album->getChildren() as $item): ?>
-                    <!-- TODO: fix this -->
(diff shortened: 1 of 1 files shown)


tomudding:
I have chosen to completely separate sub-albums from photos. Masonry has issues with calculating the correct height of the row of sub-albums. This is partially due to the text (caption) under the image. Furthermore, it breaks again during image load (why? I don't know).

I have tried some "pure" Masonry fixes, but this meant that the layout would only be applied after loading all images (for example with `imageLoaded`). I think you can imagine what this looked like.

Because Masonry does no longer calculate the width and gutter for sub-albums, in some cases the size of the sub-album cover image does not match the size of the photos. I have tried to fix this by altering the (previously) used Bootstrap classes to match the Masonry calculations.

Issue #933, opened 2019-10-04
Metadata icon not visible in Chrome


armandduijn:
The metadata icon (ⓘ) in the topbar of the new photo viewer is not visible in Chrome v77. It does show-up in Safari on macOS and iOS, though.

<img width="720" alt="Schermafbeelding 2019-10-04 om 13 07 12" src="https://user-images.githubusercontent.com/706385/66203363-7ac2ed00-e6a8-11e9-9eb3-a0336a077eed.png">

jszanto:
I'm unable to reproduce this, is it perhaps caused by a chrome extension which you are using?

armandduijn:
Indeed, you're correct. I narrowed it down to AdBlock. So it's not our fault, but a issue nonetheless. I'm guessing that a lot of our users have this extension installed.

Probably AdBlock is triggered by the word "share" in the class name.

armandduijn:
Confirmed

<img width="1037" alt="Schermafbeelding 2019-10-18 om 14 14 25" src="https://user-images.githubusercontent.com/706385/67093502-1fb5ed80-f1b2-11e9-8993-f16002a9616d.png">

PR #942, opened 2019-10-18
Change class name of button (fix #993)

### LINKED ISSUE #993: Organ mutations aren't updated
I don't think it is related to #925 but I noticed 'new' organ mutations aren't shown on the committee pages. 
See e.g. [Kas Kontrole Kommissie 2020](https://www.gewis.nl/organ/show/273).
[The mutations from AV172](https://www.gewis.nl/decision/AV/172) are not shown.

armandduijn:
See #933

### DIFF
--- a/public/js/photoswipe-ui-default.min.js
+++ b/public/js/photoswipe-ui-default.min.js
@@ -501,7 +501,7 @@
                     name: 'button--info',
                     option: 'infoEl',
                     onTap: function() {
-                        console.log(_captionContainer, 'caption--empty', pswp.currItem.title);
+                        _toggleShareModal();
                     }
                 },
                 {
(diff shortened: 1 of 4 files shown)


jszanto:
Did you test this? Afaik changing the name of that class only there will break the button.

armandduijn:
Nope. I only find and replaced occurrences of the class name. I’ll look into it.

armandduijn:
The functionality has been fixed and tested. The button is now displayed and toggles the metadata modal. See commit body and changes for full details.

public/js/photoswipe-ui-default.min.js
```diff
@@ -501,7 +501,7 @@
                     name: 'button--info',
                     option: 'infoEl',
                     onTap: function() {
-                        console.log(_captionContainer, 'caption--empty', pswp.currItem.title);
+                        _toggleShareModal();
```
kokx: Usually, the leading `_` means that the function used is private, or somehow not part of the public API.  Or that it hides functionality that should not be relied upon as a user of the API. Is that the case here?

armandduijn: Yes, it's a private function but it's being called from within the library. I edited the library's code. Unfortuantly, there isn't a way to toggle the modal from the public API.

Right now the metadata modal piggy-backs on the share modal provided by the library. If modifying the library's code to toggle the share modal is undesirable, we would have to implement a custom modal.

armandduijn: @kokx Do you have any comments?

kokx: Yeah. That I hate that something like this is necessary. But if it absolutely is necessary, its okay I guess.

armandduijn:
@Koen1999 No more blocking comments. Ready to merge.

PR #946, opened 2019-10-24
Automatically fill end date when start date is filled

### LINKED ISSUE #945: Make filling in dates for options easier
afbeelding

When filling in the start date, it is known that the end date is after or on the start date.
To make this process easier, we can automatically set the end date to the start date once it has been filled in. Often these two dates are the same.

This was a request that came up during the CM.

Koen1999:
This fixes #945 

I tested it, and it works

### DIFF
--- a/module/Activity/view/partial/option.phtml
+++ b/module/Activity/view/partial/option.phtml
@@ -51,4 +51,8 @@
 
     $("#begin-time-<?= $this->index ?>").datetimepicker(options);
     $("#end-time-<?= $this->index ?>").datetimepicker(options);
+
+    $("#begin-time-<?= $this->index ?>").change(function() {
+        $("#end-time-<?= $this->index ?>").val($("#begin-time-<?= $this->index ?>").val());
+    });
 </script>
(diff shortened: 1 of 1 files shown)


kokx:
Seems like a good fix. Maybe take my comment at #945 in account.

armandduijn:
I would like to suggest an extra check. Only auto-update the end date if the end date input hasn't been touched by the user.

Entering a start date, then an end-date, then updating the start date overwrites the previously entered end date. This scenario may occur when adding options for activities spanning multiple days where the user initially enters a wrong start date.

This could be implemented by storing a variable `isTouched = false` and setting it to `true` when the end date input has been changed by the user. Not sure if this can be done with an additional jQuery's `change()` function because setting a value via Javascript as on line 56 may also fire this event handler.

Koen1999:
> 
> 
> I would like to suggest an extra check. Only auto-update the end date if the end date input hasn't been touched by the user.
> 
> Entering a start date, then an end-date, then updating the start date overwrites the previously entered end date. This scenario may occur when adding options for activities spanning multiple days where the user initially enters a wrong start date.
> 
> This could be implemented by storing a variable `isTouched = false` and setting it to `true` when the end date input has been changed by the user. Not sure if this can be done with an additional jQuery's `change()` function because setting a value via Javascript as on [line 56](https://github.com/GEWIS/gewisweb/pull/946/files#diff-415b33ecc3ea0c59d25adae92cbcd8dbR56) may also fire this event handler.

I can also adjust it to only change the value when the end date field is empty. I think this has the same effect.

Koen1999:
I have made the changes.

PR #953, opened 2019-11-12
Allow adding labels for jobs

### LINKED ISSUE #949: Implement labels for jobs
The idea is that jobs/internships/etc can be given multiple subcategories/labels.
A job may for example receive labels "Data Science" and "Artificial Intelligence"

It's probably useful to implement the possible categories as a model so new labels can be added easily.

Request from Susan.

Koen1999:
Still some bugs present, WIP

Should fix #949


Koen1999:
This pull request can be merged. This part of the labelling system is fully functional.
Labels can be created and changed in the admin interface, similar to categories.
We still need to add functionality to add these labels to jobs.

To be continued....in another PR

module/Company/Module.php
```diff
@@ -146,8 +160,9 @@ private function getOtherFactories()
                 $acl->allow('guest', 'company', 'list');
                 $acl->allow('guest', 'company', 'view');
                 $acl->allow('guest', 'company', 'listVisibleCategories');
+                $acl->allow('guest', 'company', 'listVisibleLabels');
                 $acl->allow('guest', 'company', 'showBanner');
-                $acl->allow('company_admin', 'company', ['insert', 'edit', 'delete', 'listall', 'listAllCategories']);
+                $acl->allow('company_admin', 'company', ['insert', 'edit', 'delete', 'listall', 'listAllCategories', 'listAllLabels']);
```
kokx: Maybe split this into multiple lines?

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
     }
+
+    /**
+     * Action that displays a form for editing a label
+     *
+     *
+     */
+    public function editLabelAction()
+    {
+        // Get useful stuff
```
kokx: If your comment is like this, you can just as well leave it out. Either the comment should be improved, or the code is trivially self-explanatory.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
+        $companyService = $this->getCompanyService();
+        $labelForm = $companyService->getLabelForm();
+
+        // Get parameter
```
kokx: Idem.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
+        $labelId = $this->params('labelID');
+
+        // Get the specified company
```
kokx: And again.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
+        $labelForm->bind($labelsDict);
+        $labelForm->setAttribute(
+            'action',
+            $this->url()->fromRoute(
+                'admin_company/editLabel',
+                [
+                    'labelID' => $labelId,
+                ]
+            )
+        );
```
kokx: This looks like work that should be partially done in the Form and partially in the View. Not in the controller.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
+            }
+        }
+
+        $vm = new ViewModel([
+            'labels' => $labels,
+            'form' => $labelForm,
+            'languages' => $this->getLanguageDescriptions(),
+        ]);
+
+        return $vm;
```
kokx: Just do `return new Viewmodel(...)`.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -256,6 +257,70 @@ public function editCategoryAction()
...
+                    'labelID' => $labelId,
+                ]
+            )
+        );
+        // Handle incoming form data
+        $request = $this->getRequest();
+        if ($request->isPost()) {
+            if ($companyService->saveLabel(
+                $request->getPost()
+            )){
```
kokx: I don't think this would be that long as a single line?

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -508,6 +573,49 @@ public function addCategoryAction()
         ]);
     }
 
+    public function addLabelAction()
+    {
+        // Get useful stuff
```
kokx: > If your comment is like this, you can just as well leave it out. Either the comment should be improved, or the code is trivially self-explanatory.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -508,6 +573,49 @@ public function addCategoryAction()
...
+        // The form was not valid, or we did not get data back
+
+        // Initialize the form
+        $labelForm->setAttribute(
+            'action',
+            $this->url()->fromRoute(
+                'admin_company/default',
+                ['action' => 'addLabel']
+            )
+        );
```
kokx: Again, looks like work that should not be done here.

module/Company/src/Company/Form/EditLabel.php
```diff
@@ -0,0 +1,65 @@
...
+                'name' => 'labels',
+                'hydrator' => $this->getHydrator(),
+                'options' => [
+                    'use_as_base_fieldset' => true,
+                    'count' => count($languages),
+                    'target_element' => new LabelFieldset($translate, $this->getHydrator()),
+                    'items' => $languages,
+                ]
+            ]
+        );
```
kokx:
```
$this->add([
    .....
]);
```

Would look a lot less verbose and more clean.

module/Company/src/Company/Form/EditLabel.php
```diff
@@ -0,0 +1,65 @@
...
+
+        $this->setHydrator($hydrator);
+
+        $this->setAttribute('method', 'post');
+
+        $this->setLanguages($languages);
+
+        $this->add(
+            [
+                'type' => '\Company\Form\FixedKeyDictionaryCollection',
```
kokx: Use the `::class` constant here.

module/Company/src/Company/Form/LabelFieldset.php
```diff
@@ -0,0 +1,69 @@
...
+                'use_hidden_element' => true,
+            ],
+        ]);
+        // Hidden language element, because it will only be set at initialization.
+        $this->add([
+            'name' => 'language',
+            'attributes' => [
+                'type' => 'hidden',
+            ],
+        ]);
```
kokx: Could use some newlines.

module/Company/src/Company/Mapper/Label.php
```diff
@@ -0,0 +1,155 @@
...
+        $label->setHidden(false);
+        $this->em->persist($label);
+        $this->em->flush();
+        if ($id == -1) {
+            $id = $label->getId();
+        }
+        $label->setLanguageNeutralId($id);
+
+        return $label;
+    }
```
kokx: Unless you have very good reasons not to, this is usually a job left for the service. You would create a label in the service, and then send it over to the mapper's `persist` method.

module/Company/src/Company/Mapper/Label.php
```diff
@@ -0,0 +1,155 @@
...
+    public function delete($labelID)
+    {
+        $label = $this->findEditableLabel($labelID);
+        if (is_null($label)) {
+            return;
+        }
+
+        $this->em->remove($label);
+        $this->em->flush();
+    }
```
kokx: `delete` or `remove` methods should only have an argument by object. You could create a `removeById` method, but it should simply be a wrapper around the normal `remove` method.

module/Company/src/Company/Mapper/Label.php
```diff
@@ -0,0 +1,155 @@
...
+    public function siblingLabel($label, $lang)
+    {
+        $objectRepository = $this->getRepository(); // From clause is integrated in this statement
+        $qb = $objectRepository->createQueryBuilder('c');
+        $qb->select('c')->where('c.languageNeutralId=:labelID')->andWhere('c.language=:language');
+        $qb->setParameter('labelID', $label->getLanguageNeutralId());
+        $qb->setParameter('language', $lang);
+        $labels = $qb->getQuery()->getResult();
+        return $labels[0];
+    }
```
kokx: These two methods seem like they would belong in a service.

module/Company/src/Company/Service/Company.php
```diff
@@ -199,6 +199,41 @@ public function getCategoryList($visible)
...
+            return $this->getUniqueInArray($results, function ($a) {
+                return $a->getLanguageNeutralId();
+            });
+        }
+        if (!$this->isAllowed('listVisibleLabels')) {
+            throw new \User\Permissions\NotAllowedException(
+                $translator->translate('You are not allowed to list all labels')
+            );
+        }
+        if ($visible) {
```
kokx: This is true here by definition. Since you always either throw an exception from the earlier `if (!$visible)` or return from it.

module/Company/src/Company/Service/Company.php
```diff
@@ -215,6 +250,24 @@ public function insertCategory($lang, $id, $cat = null)
...
+     */
+    public function insertLabel($lang, $id, $cat = null)
+    {
+        if (!$this->isAllowed('insert')) {
+            $translator = $this->getTranslator();
+            throw new \User\Permissions\NotAllowedException(
+                $translator->translate('You are not allowed to insert a job')
+            );
+        }
+        $result = $this->getLabelMapper()->insert($lang, $id, $cat);
```
kokx: Just before this line would be the perfect place to create a label model, and send it to the mapper.

module/Company/src/Company/Service/Company.php
```diff
@@ -246,6 +299,37 @@ public function insertCategoryByData($data, $files)
...
+            $id = -1;
+            foreach ($newLabels as $lang => $label) {
+                $arr[$lang] = $this->insertLabel($lang, $id, $label);
+                if ($id == -1) {
+                    $id = current($arr)->getId();
+                }
+            }
+            return $arr;
+        }
+        return null;
```
kokx: Why not something like:

```
if (!$form->isValid()) {
    return null;
}
// do other things here
```

?

module/Company/src/Company/Service/Company.php
```diff
@@ -264,6 +348,26 @@ public function saveCategory($data = null)
...
+        if ($data != null) {
+            $labelForm = $this->getLabelForm();
+            $labelForm->setData($data);
+            if ($labelForm->isValid()) {
+                $this->saveLabel();
+                return true;
+            }
+            return;
+        }
+        $this->getLabelMapper()->save();
```
kokx: Always use `!==` or `===` when comparing non-object types.

Does the `save` method need to be called when there was no data to be inserted?

If it still needs to be called, you could do something like:

```
if ($data !== null) {
    $this->getLabelMapper()->save();
    return;
}

// other things
```

module/Company/src/Company/Service/Company.php
```diff
@@ -627,6 +731,29 @@ public function getAllCategoriesById($categoryID)
...
+        if (!$this->isAllowed('edit')) {
+            $translator = $this->getTranslator();
+            throw new \User\Permissions\NotAllowedException(
+                $translator->translate('You are not allowed to edit packages')
+            );
+        }
+        if (is_null($labelID)) {
+            throw new \Exception('Invalid argument');
+        }
+        $package = $this->getCategoryMapper()->findAllCategoriesById($labelID);
```
kokx: This function says it finds categories, but the method itself says you want to get all labels. Doesn't seem right?

module/Company/src/Company/Service/Company.php
```diff
@@ -627,6 +731,29 @@ public function getAllCategoriesById($categoryID)
...
+
+        return $package;
```
kokx: You can simply return this immediately, no need to store it in a variable in between.

module/Company/src/Company/Controller/AdminController.php
Koen1999: The code, and related comments are mainly adjusted from the Category methods/objects. The code is self-explanatory enough.

module/Company/Module.php
Koen1999: Fixed this

module/Company/src/Company/Controller/AdminController.php
Koen1999: I agree on this, fixed this

module/Company/src/Company/Controller/AdminController.php
Koen1999: I tried to be consistent with already existing code in this module.

module/Company/src/Company/Form/EditLabel.php
Koen1999: I agree, fixed it

module/Company/src/Company/Mapper/Label.php
Koen1999: This was consistent with the existing code within the module, but I strongly agree with your comment. I have changed it for the existing code as well.

module/Company/src/Company/Service/Company.php
Koen1999: Good point

module/Company/src/Company/Service/Company.php
Koen1999: Yep, did so

module/Company/src/Company/Service/Company.php
Koen1999: This seems more logical, I also adjusted the existing code

module/Company/src/Company/Service/Company.php
Koen1999: It does not appear to be neccesary to me, so I removed it, also for the existing code

module/Company/src/Company/Service/Company.php
Koen1999: Appearantly I forgot to change that

module/Company/src/Company/Service/Company.php
Koen1999: Agreed

module/Company/src/Company/Form/LabelFieldset.php
Koen1999: I added some

module/Company/src/Company/Controller/AdminController.php
Koen1999: I changed it in most views

module/Company/src/Company/Mapper/Label.php
Koen1999: I see your point, but this requires some slightly larger changes to the module. To be consistent with the remaining code within the module I chose not to change this.

module/Company/src/Company/Controller/AdminController.php
kokx: IMO, you should always leave code better than you found it. Never add more trash, especially if the reason is that it was done this way elsewere.

module/Company/src/Company/Mapper/Label.php
kokx: I would prefer that you fix these mistakes here at least. And possibly add an issue to refactor the rest of the module. Instead of making the refactoring an even more herculean task to be done later.

Koen1999: While making these changes, I actually noticed this code is used in no place.

module/Company/src/Company/Controller/AdminController.php
Koen1999: I looked into this, and there actually is a reason for this. The same form is being used for two different actions. addLabel and editLabel. The same holds for the Category Form. Moving this piece of code to the form would either result in duplication of code or the Controller still managing this indirectly.

PR #956, opened 2019-11-14
Display jobs as a grid

### LINKED ISSUE #948: Make jobs page more compact
The jobs page currently only displays 1 or 2 jobs because they are displayed including a description and other stuff, taking up a lot of space. The idea is to make the page more compact, possibly using cards so it gives a better overview of the available jobs. Full details on the job are displayed on the job page itself.

Request from Susan.

armandduijn:
Changes the layout of the jobs page by displaying the jobs in a grid (fix #948).

## Changes
As requested, the main jobs page is made compacter by removing some details. These details are still viewable when navigating to a specific job.

I've added the date when the job was posted to the website. This may be useful as older job posts may be irrelevant, e.g. deadline submission passed. Furthermore, each job has tags that communicate whom to the target audience is (backend implemented by @Koen1999).

The page doesn't throw a 404 when no jobs/internships/... are available. Instead, it simply shows a "No Results" message.

## Screenshots
- [Phone](https://user-images.githubusercontent.com/706385/68847967-d03ee080-06cf-11ea-9db3-351c8d61f939.png)
- [Tablet](https://user-images.githubusercontent.com/706385/68848039-e77dce00-06cf-11ea-8e1d-aebe89a963df.png)
- [Desktop](https://user-images.githubusercontent.com/706385/68848074-f5335380-06cf-11ea-80ee-b28fa645ab34.png) (with hover-over tag)

## Todo

- [ ] ~Retrieve job tags from database (currently hardcoded)~
- [x] Update styling of company-specific job page
- [x] Move job list shuffling logic to controller or service
- [ ] Test layout in different browsers


armandduijn:
Depends on #953

jszanto:
A few suggestions, some of these could be separate issues perhaps:

- [ ] Remove the "view job" link, make it such that you can click the entire card
- [ ] Add the location of the job to the card
- [ ] On top add buttons to filter by labels
- [ ] On top add buttons to change the sorting of the jobs
- [ ] Show the date as "X days/weeks/months ago"

Does this also fix the inconsistently sized company logos?

armandduijn:
1. The entire card is already clickable. The link at the bottom is just to communicate that there's more information available on a separate page.
2. What do you mean with location? Do you mean the company's address?
3. I'll implement filtering-by-label (and removing the hard-coded labels) once PR #953 is closer to completion.
4. Adding sorting is a good suggestion, but we should tackle that in a separate PR. Also, how would we implement sorting? Server-side using GET parameters or client-side using Javascript (similar to the search).
5. I'll have to investigate how to convert the date to "X ago". I'm aware of a Javascript library called [Moment.js](https://momentjs.com) that can do this, but it might be overkill to introduce a new dependency for such a small feature. Do you know if it's possible to convert the date with Zend?

I've only tested the layout with a couple of logos currently on gewis.nl. But it should work fine for other weirdly shaped logos. The logo's container has a fixed height and the logo inside can grow/shrink proportionally to fill the height of its container.

jszanto:
1. In my opinion it is obvious that the card is clickable, optionally you could add a `transform: scale(1.01);` on hover to the card to make it more obvious. But if it is not clear enough I guess the link can stay.
2. Just the city of the company/job, not all of them are in Eindhoven.
4. I'd do the sorting client-side, since there are probably <100 jobs at any time.
5. You could use [Carbon](https://carbon.nesbot.com/docs/#api-humandiff) for this, although this one really is a nice-to-have, not a must.

armandduijn:
(1), (3), and (4) are noted.

Do we store the city of a job? The job table doesn't have a city field, neither does the company table. Showing the company's address might be an option, but judging from the production database it's kind of a mess. Full addresses without a consistent syntax mixed-in with HTML. Futhermore, is the address of a company also the address of a job? For example, the Rabobank has multiple offices but its headquarter is in Utrecht.

armandduijn:
@Koen1999 This PR is nearly complete. I'm only waiting on your PR #953 to merge so that I can remove the hardcoded labels.

Koen1999:
Is it perhaps an idea to comment out the labels in the view and deploy this PR as is?
I think the changes look good so far.

armandduijn:
Sure, @Koen1999. I've commented-out the HTML so we can re-enable it later on.

Can we test this branch with production data? I'm not sure how to push to the test server. Note that I've added a property to the `Job` model so an ORM update is needed.

For your convenience, I've already translated the strings so you can easily copy them when eventually merging to master.

- [Translations.txt](https://github.com/GEWIS/gewisweb/files/3892362/Translations.txt)

Koen1999:
I will test and deploy it later today.

Koen1999:
I have tested it and I found one issue.
The company images are typically of high resolution and they don't fit within the cards, thus disturbing the layout.

Koen1999:
I made some mistakes in merging the PR. The changes are succesfully committed to master after 01a379fdbd5210f1997c8054fa9545e8346ea3d1

PR #959, opened 2019-11-20
Fix #903: Add English regulations

### LINKED ISSUE #903: Update Regulations links to english on members page
afbeelding

located on https://gewis.nl/member

jszanto:
This is achieved by linking them to the public archive


Koen1999:
I don't think is ideal behavior, but dependencies are reduced, which is good.
I will merge this PR, but I think that we can also consider removing the links to the regulations on the member page as by this PR, it's just as easy to click public archive instead.

PR #960, opened 2019-11-25
Refactored sorting for documents such that it's not stored in the db

Koen1999:
Persisting the order in the database was not working anyway, so why even make that attempt.

### DIFF
--- a/module/Decision/src/Decision/Controller/DecisionController.php
+++ b/module/Decision/src/Decision/Controller/DecisionController.php
@@ -69,7 +69,6 @@ class DecisionController extends AbstractActionController
 
         try {
             $meeting = $service->getMeeting($type, $number);
-            $meeting->sortDocuments();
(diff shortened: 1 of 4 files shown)


armandduijn:
I'm planning implementing sorting properly. I've already written the frontend in [a branch on my fork](https://github.com/armandduijn/gewisweb/tree/feature/reorder-documents). Maybe we can skip this PR and I'll instead remove the code in my PR.

Koen1999:
If you wish to implement this properly, please do so.
This is more like a hotfix thing for Bouke's email. I think this should
work.
When I looked at the code I was annoyed, so I though I would only slightly
improve it 🙈

Op ma 25 nov. 2019 11:14 schreef Armand Duijn <notifications@github.com>:

Koen1999:
This was a fix for #961

PR #962, opened 2019-11-25
Allow reordering meeting documents

### LINKED ISSUE #961: Documents aren't corretly sorted
This is an issue for Bouke's mail from 25 November.

armandduijn:
Allows admins to reorder meeting documents (fix #961).


module/Decision/src/Decision/Service/Decision.php
```diff
@@ -257,6 +264,58 @@ public function deleteDocument($post)
...
+
+    /**
+     * Changes a document's position in the ordering
+     *
+     * @param int $id Document ID
+     * @param bool $moveDown If the document should be moved down in the ordering, defaults to TRUE
+     * @return void
+     * @throws \InvalidArgumentException If the document doesn't exist
+     */
+    public function changePositionDocument($id, $moveDown = true)
```
armandduijn: The basic flow is (1) retrieve documents, (2) swap document positions, then (3) persit position. Unfortuantly, I have to update the positions of all documents related to a meeting because of legacy. Old documents don't have a positon yet, so they are set to 0 by default. In the future, when documents have display positions, we can simplify this by only mutating two rows.

module/Decision/src/Decision/Controller/AdminController.php
```diff
@@ -66,6 +66,42 @@ public function deleteDocumentAction()
         return $this->redirect()->toRoute('admin_decision/document');
     }
 
+    public function changePositionDocumentAction()
```
armandduijn: Do I need to check permissions here?

module/Decision/src/Decision/Controller/AdminController.php
```diff
@@ -66,6 +66,42 @@ public function deleteDocumentAction()
...
+    {
+        if (!$this->getRequest()->isPost()) {
+            // TODO: Throw 405 Method Not Allowed error
```
armandduijn: Could someone provide guidance on how to create a new error page? Creating such a page might be benefical for other routes too.

Koen1999: I think that what you want to throw is a 403. You can do so like this: 
```
throw new NotAllowedException(
    $translator->translate('You are not allowed to ...')
);
```

module/Decision/src/Decision/Controller/AdminController.php
Koen1999: Yes, you should. I think you can use something like `$this->isAllowed('delete_document', 'meeting')`

armandduijn:
The implementation is done.

I'm using Ajax requests to send the data to the backend. Because of this Ajax-only approach, I decided to remove the `redirect()` that was previously in the controller action. Instead, the controller action returns simple API-like responses. The responses only contain the applicable status code and any errors.

The layout works in all major browsers (latest). It might look awkward on mobile, but early on we decided not to optimize the admin dashboard for mobile.

## Screenshots
- [Screenshot of document list](https://user-images.githubusercontent.com/706385/69810505-5db02380-11ec-11ea-8652-b4be4f90f6ef.png)
- [Screenshot of document list during Ajax request (icons disabled)](https://user-images.githubusercontent.com/706385/69810524-67398b80-11ec-11ea-8c7a-3dcffaff2217.png)

module/Decision/src/Decision/Service/Decision.php
```diff
@@ -523,4 +598,23 @@ public function isAllowedToBrowseFiles()
...
+
+    /**
+     * Checks the user's permission
+     *
+     * @param string $operation
+     * @param string $resource
+     * @param string $errorMessage English error message
+     * @throws NotAllowedException If the user doesn't have permission
+     */
+    private function isAllowedOrFail($operation, $resource, $errorMessage)
```
Koen1999: I kind of like the idea of having a method for this, as this is a very regularly used block of code.

armandduijn: Maybe refactoring it in a [trait](https://www.php.net/manual/en/language.oop5.traits.php)? Other services or controllers can then make use of this shared functionality.

Koen1999: Well, if you would like other services to make use of it, you would have to refactor that code as well, which is a lot of work. I'm not quite sure what a good place would be to put it. But for this PR, it's fine like this.

Koen1999:
The point I'm confused about is the following.
Previously we sorted documents on the fly. You removed that code.
How are the meeting documents of old meetings displayed now? Should they all be sorted manually?

I think having the old sorting mechanism as a fallback is preferred.

armandduijn:
Documents are sorted automatically by the attribute `displayPosition` when they're retrieved in the context of a meeting. This is implemented using the Doctrine annotation [OrderBy](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/annotations-reference.html#orderby) ([link to code fragment](https://github.com/GEWIS/gewisweb/pull/962/files#diff-8e8d864fdc79ad35d45b06b8696df23bR51-R57)).

Koen1999:
Alright, I see. But the displayPosition attribute is not present in the older versions and therefore unset for all previous meetings.

armandduijn:
I prefer to remove the old sorting algorithm because keeping it would make the code unnecessarily complicated.

I propose to: run the previous sorting algorithm on all meeting documents (627 total) offline, calculate the display positions, and update them in the database. This ensures that older documents are still sorted correctly after merging this PR and removes the necessity for legacy code.

Do you agree? If so, I'm thinking of supplying a SQL script that you can execute.

Koen1999:
> 
> 
> I prefer to remove the old sorting algorithm because keeping it would make the code unnecessarily complicated.
> 
> I propose to: run the previous sorting algorithm on all meeting documents (627 total) offline, calculate the display positions, and update them in the database. This ensures that older documents are still sorted correctly after merging this PR and removes the necessity for legacy code.
> 
> Do you agree? If so, I'm thinking of supplying a SQL script that you can execute.

This sounds like a good solution to make it backwards compatible.

armandduijn:
I wrote a custom command that will sort the meeting documents using the legacy sorting algorithm and store their positions. This was previously calculated on-the-fly.

First, run the following command in the base directory of the website:
```
php public/index.php orm:schema-tool:update --dump-sql
```
This should only return one SQL query that adds the `displayPosition` attribute.

Then, run the command to update the database:
```
php public/index.php orm:schema-tool:update --force
```

Finally, run the custom command I wrote:
```
php public/index.php documents sort-legacy
```

Tested locally and completed within 5 seconds.

**Note**: I suggest making a backup of the database before running the command. Even though I don't forsee any problems (only inserts values in a new field), it's better to have a quick way to revert to a previous state.

Koen1999:
Looks very good! Codeclimate issues seem to be not problematic. I will proceed to make a backup, merge the PR and deploy using the script.

Koen1999:
It works good, but I noticed that the sorting algorithm didn't sort the documents without numbering very well. I slightly modified the sorting algorithm in 8da71d627f1d3cec92fb67f41d254a3ca09ca67a to perform better on these documents

PR #964, opened 2019-11-28
Allow for adding labels to jobs

### LINKED ISSUE #949: Implement labels for jobs

### LINKED ISSUE #1014: Editing of labels and categories is not working
The code of these two models is similar, hence the grouped issue.

Koen1999:
This is a new PR, coupled to another branch because of some merge/revert issues I had.

Follow up of #953, thus this fixes #949. It also fixes #1014.


Koen1999:
Current issues and progress: 

- [ ] Labels cannot be changed after creation
- [] Probably some issues when assigning or removing labels from jobs (to be verified)

Koen1999:
@tomudding @kokx @jszanto @armandduijn 
It is working except for the issues mentioned in the comment above. Do you think we could merge this or should these bugs be resolved first?
I tested the functionality, and it should be usable for the external affairs officer.

A code review would by either of you be welcome anyway :)

Here are some images of the adjusted front-end: 
afbeelding
afbeelding

module/Company/view/company/admin/index.phtml
```diff
@@ -24,110 +24,141 @@
...
-                            </th>
+                            <td><?= $label->getName(); ?> (<a href=<?= $editUrl; ?>><?= $this->translate('edit')?></a>)</td>
                         </tr>
+                    <?php endforeach; ?>
+                    </tbody>
+                    <tfoot>
+                    <tr>
+                        <th><?= $this->translate('Categories') ?></th>
+                    </tr>
+                    </tfoot>
```
tomudding, line 82: This should be `Labels` and not categories. However, is there what is the reason we should have an additional table footer when the name is already in the table header?

The same applies to the table footer for the `Categories` section.

tomudding:
Just to ensure that this is intended behaviour (or not):

If a Job is only available in one language, but we display the Job in multiple languages, should the label only be visible for that one language?

An example if this is Job 859 (see test), it is only in English (where the labels are shown). However, the Job is still visible while the locale is Dutch, but there is no slug (not even the English one).

module/Company/view/company/admin/index.phtml
```diff
@@ -24,110 +24,141 @@
         </div>
         <hr>
         <div class="row">
-            <div class="col-md-2">
-                <a class="btn btn-success" href='<?= $this->url('admin_company/default', ['action' => 'addCategory'])?>'> <i class="fas fa-plus"></i><?=$this->translate('Add Category');?>
- </a>
+            <div class="col-sm-6">
+                <a class="btn btn-success" href='<?= $this->url('admin_company/default', ['action' => 'addCategory'])?>'> <span class="glyphicon glyphicon-plus"></span><?=$this->translate('Add Category');?>
```
Koen1999, line 28: We switched to FontAwesome right? Shoudl probably adjust this

module/Company/view/company/admin/index.phtml
```diff
@@ -24,110 +24,141 @@
...
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-md-8">
+                <table class="table table-striped ">
+                    <thead>
+                    <tr>
+                        <th>
+                            <?= $this->translate('Name') ?>
+                            <a onclick="Company.Admin.sortCompaniesByName(1)"><span class="glyphicon glyphicon-arrow-up"></span></a>
```
Koen1999, line 93: Same Glyphicons

module/Company/view/company/admin/index.phtml
Koen1999: No, there is no reason. Just being consistent with previous designs

module/Company/view/company/admin/index.phtml
tomudding: Yes, they are already the FA equivalent on master, so they have to be changed "back" (to FA).

armandduijn:
> @tomudding @kokx @jszanto @armandduijn
> It is working except for the issues mentioned in the comment above. Do you think we could merge this or should these bugs be resolved first?

This is a pretty large PR. Is there maybe a specific part you want me to review?

Koen1999:
> 
> 
> > @tomudding @kokx @jszanto @armandduijn
> > It is working except for the issues mentioned in the comment above. Do you think we could merge this or should these bugs be resolved first?
> 
> This is a pretty large PR. Is there maybe a specific part you want me to review?

Well, if possible the whole PR. But I know you're more into front-end, so if you could do that that would be fine too.
The PR is currently on test.gewis.nl 

The only persisting issue that labels (and categories) cannot be edited. I solved the other one. This shouldn't be a big issue as this has to be done rarely if at all.

Koen1999:
> 
> 
> Just to ensure that this is intended behaviour (or not):
> 
> If a Job is only available in one language, but we display the Job in multiple languages, should the label only be visible for that one language?
> 
> An example if this is Job 859 (see test), it is only in English (where the labels are shown). However, the Job is still visible while the locale is Dutch, but there is no slug (not even the English one).

I suppose that is intended behaviour. But the scenario is unlikely to occur anyway since all labels will probably have both an english and a dutch description.

I also saw you did some tests with long labels. Would you consider it a problematic issue?

tomudding:
> 
> 
> > Just to ensure that this is intended behaviour (or not):
> > If a Job is only available in one language, but we display the Job in multiple languages, should the label only be visible for that one language?
> > An example if this is Job 859 (see test), it is only in English (where the labels are shown). However, the Job is still visible while the locale is Dutch, but there is no slug (not even the English one).
> 
> I suppose that is intended behaviour. But the scenario is unlikely to occur anyway since all labels will probably have both an english and a dutch description.

[I'll edit in my response to this later, as there are some other issues that prevent me from testing this.]

> I also saw you did some tests with long labels. Would you consider it a problematic issue?

I do not think long labels will be problematic, as I do not see a use for very long labels. If it becomes an issue I suggest we simply edit the CSS and truncate the labels.

Koen1999:
> I might have misinterpreted the intended behaviour, but I would argue that the display name of a label should be shown to users and not the slug name.

The slug is just some abbreviation of the actual label which is shown when hovering over the tag.

tomudding:
> 
> 
> > I might have misinterpreted the intended behaviour, but I would argue that the display name of a label should be shown to users and not the slug name.
> 
> The slug is just some abbreviation of the actual label which is shown when hovering over the tag.

Then it my misunderstanding of what meant what.

module/Company/src/Company/Controller/AdminController.php
```diff
@@ -432,6 +483,12 @@ public function editJobAction()
         }
         $languages = array_keys($jobDict);
         $jobForm->setLanguages($languages);
+        $labels = $jobDict[$companyService->getTranslator()->getLocale()]->getLabels();
```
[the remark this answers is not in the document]
tomudding: @Koen1999 This is the most important one.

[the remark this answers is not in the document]
Koen1999: Good spot! I fixed it

Koen1999:
> 
> 
> Single language jobs are once again editable, however, it appears that the labels are no longer selected when editing a job.
> 
> This is not a problem for jobs that have two languages.

I suppose we've had all edge cases now :sweat_smile:

module/Company/view/company/admin/index.phtml
```diff
@@ -24,110 +24,141 @@
...
+                    <tfoot>
+                    <tr>
+                        <th><?= $this->translate('Labels') ?></th>
+                    </tr>
+                    </tfoot>
+                </table>
+            </div>
+        </div>
+        <div class="row">
+            <div class="col-md-8">
```
tomudding, line 87: I think that `col-md-12` looks better

module/Company/src/Company/Service/Company.php
```diff
@@ -521,11 +664,71 @@ public function saveJobData($languageNeutralId, $jobs, $data, $files)
             $id = $this->setLanguageNeutralId($id, $job, $languageNeutralId);
             $this->getJobMapper()->persist($job);
             $this->saveJob();
+
+            $mapper = $this->getLabelMapper();
+            $lang = $job->getLanguage();
+            // Contains language specific labels
+            $labelsLangBased = [];
+            foreach ($labelIDs as $id) {
```
tomudding, line 672: You are overwriting another `$id` here.

tomudding:
@Koen1999 I fixed an issue which caused two separate empty languages to be created for a job (or rather two different jobs), however, currently when only one language is used (either active or inactive at creation) with labels the other language is empty.

This is because the `MultiCheckbox` is global in the form and thus both languages are processed despite one being empty.

module/Company/src/Company/Service/Company.php
```diff
@@ -850,7 +1115,7 @@ public function getAcl()
     }
 
     /**
-     * Get the default resource ID.
+     * Get the default resource Id.
```
armandduijn, line 1118: We should write a section in the Wiki on how we want to capitalize acronyms in comments/method names. I understand that everybody has a personal preference, but we risk undoing each other’s changes. Additionally, it pollutes the PR with unrelated code changes.

module/Company/src/Company/Service/Company.php
```diff
@@ -183,87 +207,193 @@ public function getCategoryList($visible)
...
+            return $el->getSlug() == "jobs";
+        })) ;
+        if (count($jobsWithoutCategory) > 0 && $noVacancyCategory  == 0) {
+            $filteredCategories[] = $this->getCategoryMapper()
+                ->createNullCategory($translator->getLocale(), $translator);
+        }
+        return $filteredCategories;
+    }
+
+    public function getLabelList($visible)
```
armandduijn, line 223: The parameter name is a little confusing. I would expect `$visible = false` to return all labels that aren't visible. Instead, it returns all labels (if the user has admin rights). Consider renaming the parameter to something more descriptive such as `$onlyVisible`.

module/Company/src/Company/Service/Company.php
```diff
@@ -183,87 +207,193 @@ public function getCategoryList($visible)
...
+        $labelForm->bind($labels);
+        $labelForm->setData($data);
+        
+        if (!$labelForm->isValid()) {
+            return false;
+        }
+
+        $id = -1;
+        foreach ($labels as $label) {
+            $id = $this->setLanguageNeutralLabelId($id, $label, $languageNeutralId);
```
armandduijn, line 361: Why is the ID set to -1 and then overwritten multiple times in the foreach loop?

module/Company/src/Company/Service/Company.php
tomudding: @armandduijn I totally agree with you, normally I would not change this. But in this case, in the changes originally made, both capitalizations were used. To directly streamline it (in this module) I opted to change everything in one go.

As PSR-1 explicitly does not mention how to capitalize properties, I used the naming convention for (albeit) classes from the PHP. In this case `Id` is really used for `Identifier` whereas `ID` is used for `Identity Document`.

We should add a section to the Wiki to resolve this in the future.

module/Company/src/Company/Service/Company.php
tomudding: This code is a modified version of the `saveJobData` function, which is from (at least) 4 years ago. We might have to refactor the label and category functions as they are virtually the same and combine them. But until we switch to PHP 7 this will be messy.

If required we can change the behaviour, but we also have to change it for the aforementioned job function.

tomudding: To answer your question, on label/category creation, there is no identifier for the item. As the item is created, the `$languageNeutralId` is empty (perhaps `null` is better). The first identifier is created and will become the `$languageNeutralId`, after which the `$id` is used to update the `$languageNeutralId` for the second item (i.e. language).

module/Company/src/Company/Controller/CompanyController.php
```diff
@@ -16,7 +16,7 @@ public function listAction()
     {
         $companyService = $this->getCompanyService();
         $featuredPackage = $companyService->getFeaturedPackage();
-        if ($featuredPackage == null) {
+        if ($featuredPackage === null) {
```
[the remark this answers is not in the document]
tomudding: Functionally they are exactly the same, so there should be no differences which could cause "bugs".

I'd argue this is very similar to the `ID`/`Id`-situation. We should standardize on what to use (`is_null`, even though I also used `===`) and then apply it all at once (e.g., in another PR).

module/Company/src/Company/Service/Company.php
```diff
@@ -245,161 +245,153 @@ public function getLabelList($visible)
...
+     * Creates a new JobCategory.
      *
-     * @param mixed $packageID
+     * @param  array $data Category data from the EditCategory form.
+     * @return bool|int Returns false on failure, and the languageNeutralId on success
      */
-    public function insertCategory($lang, $id, $category = null)
+    public function createCategory($data)
     {
-        if (!$this->isAllowed('insert')) {
```
Koen1999: Why did we remove this assertion?

module/Company/src/Company/Service/Company.php
```diff
@@ -245,161 +245,153 @@ public function getLabelList($visible)
...
+     * @param int|string $languageNeutralId Identifier of the JobCategories to save
+     * @param array $categories The JobCategories to save
+     * @param array $data The (new) data to save
+     *
+     * @return bool|int Returns false on failure, and the languageNeutralId on success
      */
-    public function insertLabel($lang, $id, $label = null)
+    public function saveCategoryData($languageNeutralId, $categories, $data)
     {
-        if (!$this->isAllowed('insert')) {
```
Koen1999, line 209: Same

module/Company/view/partial/company/admin/editors/label.phtml
```diff
@@ -0,0 +1,43 @@
...
+    <?php foreach( $form->getLanguages() as $lang):?>
+    <?php $langDesc = $languages[$lang]?>
+        <div class="col-md-6">
+            <h1><?= $this->translate($langDesc)?></h1>
+            <?php $fieldset = $this->form->get("labels")->get($lang); ?>
+            <?php
+                $element = $fieldset->get('name');
+                $element->setAttribute('class', 'form-control');
+            ?>
+            <div class="form-group<?= count($element->getMessages()) > 0 ? ' has-error' : '' ?>">
```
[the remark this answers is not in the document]
Koen1999: Good tip, I didn't know we had this helper

module/Company/src/Company/Controller/CompanyController.php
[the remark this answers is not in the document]
Koen1999: The functionality is indeed the same. Since we are using different methods throughout the code anyway, and never agreed on a common method we can leave it as is for now. If we want to swap, we should do it together with other code quality improvements I guess.

module/Company/src/Company/Form/LabelFieldset.php
```diff
@@ -0,0 +1,71 @@
...
+            'attributes' => [
+                'type' => 'text',
+                'required' => 'required',
+            ],
+            'options' => [
+                'label' => $translate->translate('Display name'),
+                'required' => 'required',
+            ],
+        ]);
+        $this->add([
```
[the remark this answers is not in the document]
Koen1999: Mmhm, interesting point. I copied this from the categories. The one from categories is never used either. Should I remove them both?

module/Company/view/company/admin/index.phtml
```diff
@@ -24,110 +24,141 @@
...
-                    </tbody>
-                    <tfoot>
+                    <?php
+                    foreach ($categoryList as $category):
+                        $editUrl = $this->url('admin_company/editCategory',
+                            ['languageNeutralCategoryId' => $category->getLanguageNeutralId()]);
+                        ?>
                         <tr>
-                            <th><?= $this->translate('Categories') ?></th>
+                            <td><?= $category->getName(); ?> (<a href=<?= $editUrl; ?>><?= $this->translate('edit')?></a>)</td>
```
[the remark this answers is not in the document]
Koen1999: Good spot

module/Company/view/company/admin/edit-job.phtml
```diff
@@ -174,6 +174,19 @@ $element->setAttribute('class', 'form-control');
...
 
+<?php
+$element = $form->get('labels');
+?>
+<div class="form-group<?= count($element->getMessages()) > 0 ? ' has-error' : '' ?>">
+    <label for="<?= $element->getName() ?>" class="control-label col-sm-2">
+        <?= $element->getOption('label') ?>
+    </label>
+    <div class="col-sm-10">
+        <?= $this->formMultiCheckbox($element) ?>
```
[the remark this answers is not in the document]
Koen1999: I agree that more user-friendly options exist. Yet Zend does not allow me to easily modify the way this element is rendered. I also do not see how to use a decorator with this specific element.

module/Company/src/Company/Service/Company.php
Koen1999: I actually think that visible is more descriptive than onlyVisible. I added some PHPDoc

module/Company/src/Company/Form/LabelFieldset.php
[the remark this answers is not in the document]
armandduijn: For labels, yes, if it serves no purpose. For categories, I would leave it for now. I'm not sure if some part of the website relies on it.

module/Company/view/company/admin/edit-job.phtml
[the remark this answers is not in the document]
armandduijn: @wsinnema, you've worked with Zend Forms in the past for the company module. Do you know if it's possible to add breaks in a `MultiCheckbox` element? See code [here](https://github.com/GEWIS/gewisweb/pull/964/files/86ca2566a94d376937e245ea4b3736837fe1d6ac#diff-037cb76f3f5ebdbff79ef16fef76a3d2R42-R49).

module/Company/src/Company/Service/Company.php
tomudding: I have missed that when re-ordering the code. I will add it back later today.

module/Company/src/Company/Service/Company.php
tomudding: See above.

module/Company/src/Company/Service/Company.php
armandduijn: Adding documentation is already a good improvement 👍

module/Company/view/company/admin/edit-job.phtml
```diff
@@ -174,6 +174,19 @@ $element->setAttribute('class', 'form-control');
 <?php endforeach;?>
 </div>
 
+<?php
+$element = $form->get('labels');
+?>
+<div class="form-group<?= count($element->getMessages()) > 0 ? ' has-error' : '' ?>">
+    <label for="<?= $element->getName() ?>" class="control-label col-sm-2">
```
tomudding, line 181: This should be replaced with the `BootstrapElementError` Helper similar to the other (already changed) instances.

module/Company/view/partial/company/admin/editors/label.phtml
```diff
@@ -0,0 +1,43 @@
...
+    <?php $langDesc = $languages[$lang]?>
+        <div class="col-md-6">
+            <h1><?= $this->translate($langDesc)?></h1>
+            <?php $fieldset = $this->form->get("labels")->get($lang); ?>
+            <?php
+                $element = $fieldset->get('name');
+                $element->setAttribute('class', 'form-control');
+            ?>
+            <div class="form-group<?= $this->bootstrapElementError($element) ?>">
+                <label for="<?= $element->getName() ?>" class="control-label col-sm-4">
```
tomudding, line 12: Unlike the old `count` method for the Bootstrap errors, `BootstrapElementError` does not have a space before the `has-error`. Thus there must be a space between `form-group` and `<?=`.

This issue appears 3 times in this document.

module/Company/view/company/admin/edit-job.phtml
Koen1999: I think it's nice to have it already in place. Even if there's no real form validation at this point in time.

module/Company/view/company/admin/edit-job.phtml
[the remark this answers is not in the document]
tomudding: I think this can be done by using `setLabelAttributes()`. Just add a class with `margin-right` (and `bottom`?) to give the elements a bit more space. I'll try to see if this works.

module/Company/Module.php
```diff
@@ -146,8 +166,10 @@ private function getOtherFactories()
                 $acl->allow('guest', 'company', 'list');
                 $acl->allow('guest', 'company', 'view');
                 $acl->allow('guest', 'company', 'listVisibleCategories');
+                $acl->allow('guest', 'company', 'listVisibleLabels');
```
tomudding, line 169: @Koen1999 Is this privilege still used now that you removed the hidden property?

Koen1999: Yes it is. Invisible labels are defined as "labels that are not assigned to any active job" in this context

Koen1999:
@armandduijn @tomudding I think this PR is now ready to be merged.
With some help from @wsinnema I managed to fix the layout of the checkboxes.
afbeelding

PR #965, opened 2019-11-29
Updates the layout of the meeting page

armandduijn:
This PR updates the layout of the meeting page (user-faced) and adds the ability to directly link to specific decisions. Furthermore, I've fixed an issue which caused the meeting date to only show in Dutch.

## Screenshots
- [Meeting page with data](https://user-images.githubusercontent.com/706385/69879538-fa93bf00-12c7-11ea-88d6-cb1a364d1eeb.png)
- [Meeting page with data (hover over decision)](https://user-images.githubusercontent.com/706385/69879559-067f8100-12c8-11ea-849a-50fc0dc81ca9.png)
- [Meeting page without data](https://user-images.githubusercontent.com/706385/69879588-18612400-12c8-11ea-94b8-9974ac600117.png)

## Translations

- [Translations.txt](https://github.com/GEWIS/gewisweb/files/3905827/Translations.txt)

### DIFF
--- a/module/Application/Module.php
+++ b/module/Application/Module.php
@@ -32,6 +32,7 @@ class Module
         $translator->setlocale($locale);
 
         Carbon::setLocale($locale);
+        \Locale::setDefault($locale);
 
         $eventManager->attach(MvcEvent::EVENT_DISPATCH_ERROR, [$this, 'logError']);
         $eventManager->attach(MvCEvent::EVENT_RENDER_ERROR, [$this, 'logError']);
--- a/module/Decision/view/decision/decision/view.phtml
+++ b/module/Decision/view/decision/decision/view.phtml
@@ -1,54 +1,72 @@
 <?php
-// set title
-$this->headTitle(sprintf('%s %d', $meeting->getType(), $meeting->getNumber())); ?>
+$meetingName = sprintf('%s %d', $meeting->getType(), $meeting->getNumber());
+
+$this->headTitle($meetingName);
+?>
 <section class="section">
     <div class="container">
-        <h1><?= $meeting->getType() ?> <?= $meeting->getNumber() ?>
-            <small><?= $this->dateFormat(
-                    $meeting->getDate(),
-                    \IntlDateFormatter::FULL,
-                    \IntlDateFormatter::NONE,
-                    'nl_NL'
-                ) ?></small>
-        </h1>
+        <h1><?= $meetingName ?></h1>
+        <h2>
+            <?= $this->dateFormat($meeting->getDate(), \IntlDateFormatter::FULL, \IntlDateFormatter::NONE) ?>
+        </h2>
         <div class="row">
             <div class="col-md-6">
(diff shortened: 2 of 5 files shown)


module/Application/Module.php
```diff
@@ -32,6 +32,7 @@ public function onBootstrap(MvcEvent $e)
         $translator->setlocale($locale);
 
         Carbon::setLocale($locale);
+        \Locale::setDefault($locale);
```
armandduijn: Will setting the locale globally break anything? I need to set it here so that the meeting date is displayed correctly.

kokx: Probably not. But maybe this should be rolled out on test first and tested for a bit. Especially looking for dates in various areas (such as with photos and activities).

Koen1999:
I am wondering. What does the hyperlink at the decision point to?
afbeelding

armandduijn:
When you click it, it updates the URL to point directly to the decision (e.g. `/../AV/1#AV-1.1.1`). You can now share this link. Works similarily to how GitHub allows you to share a link to a code line: click line number and see how the URL is updated.

module/Application/Module.php
Koen1999: Has this been tested already?

armandduijn: No, not yet. I still have to look into it. My intuition also says that line 34 might be unnecssary because Carbon uses the global locale. I'll have to check first.

armandduijn: I did some investigation and I've concluded that (1) setting the locale globally will not break anything, and (2) setting the Caron locale separately is needed.

Regarding (1), I'm using `\Locale::setDefault()` so that the view helper `dateFormat()` correctly outputs formats the date on the meeting page ([line 9](https://github.com/GEWIS/gewisweb/pull/965/files#diff-4ee0de3a261ae71a9f007aae25b88f8b)).

The `dateFormat()` view helper is used 18 times throughout our code base. Every call to this view helper passes a fourth parameter specifying the session's locale ([example](https://github.com/GEWIS/gewisweb/blob/a2216e34885737f28f25ae805501a1c206816d8b/module/Activity/view/email/activity.phtml#L15)). In fact, I came across a couple of occurrences where the locale was hardcoded ([example](https://github.com/GEWIS/gewisweb/blob/a2216e34885737f28f25ae805501a1c206816d8b/module/Decision/view/decision/member/birthdays.phtml#L26-L30)). So, setting the Locale globally will not affect these calls as the fourth parameter overwrites the global default. Furthermore, Locale only affects functions that use the value returned by `Locale::getDefault()` (i.e. has no system side effects). Zend calls `getDefault()` a couple of times internally, but it should be fine.

Regarding (2), it tried removing line 34 and Carbon defaulted to English, ignoring the locale.

armandduijn: If you agree, you can resolve this conversation and proceed to merge this PR.

kokx: Looks good. Nice work!

PR #966, opened 2019-12-09
WIP: Delete models cascade

Koen1999:
In order to delete user information when a membership ends, it's likely a good idea to implement a cascading delete.

Currently, I only did it for the activity module, but feedback is welcome.

### DIFF
--- a/module/Activity/src/Activity/Model/Activity.php
+++ b/module/Activity/src/Activity/Model/Activity.php
@@ -123,7 +123,7 @@ class Activity implements OrganResourceInterface, CreatorResourceInterface
     /**
      * Who did approve this activity.
      *
-     * @ORM\ManyToOne(targetEntity="User\Model\User")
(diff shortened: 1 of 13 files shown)


kokx:
This will very likely not work nicely. Especially since changes in the database are not synced in a way that these cascades will get triggered.

IMO, this would only introduce possible bugs, rather than do anything useful.

Koen1999:
What synchronization do you mean? This is not related to the
decision/member synchronization from gewisdb.
I propose cascading when you trigger a delete on a User model. The user
model would be deleted using for example a cronjob triggering a API from
gewisweb on all users that do not have a member model.

Op ma 9 dec. 2019 23:03 schreef Pieter Kokx <notifications@github.com>:

kokx:
Then I'd say this is putting the cart before the horse. First implement the cronjob, then implement the cascade. Don't have the cascade there silently waiting for unintended side-effects.

PR #967, opened 2019-12-12
Minor changes: Automatically sent mails

rinkp:
Originally to insert missing line break, but while you're at it, you might as well just change it a little to improve it jus that little bit

screenshot current email without linebreak

### DIFF
--- a/module/Decision/view/email/authorization_received.phtml
+++ b/module/Decision/view/email/authorization_received.phtml
@@ -1,7 +1,7 @@
 <?php $lang = 'en'; ?>
-For English, see below.
+For English, please see below.<br /><br />
 
-L.s.<br /><br />
+L.S.,<br /><br />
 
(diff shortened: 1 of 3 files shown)


module/Decision/view/email/authorization_received.phtml
```diff
@@ -1,7 +1,7 @@
 <?php $lang = 'en'; ?>
-For English, see below.
+For English, please see below.<br /><br />
 
-L.s.<br /><br />
+L.S.,<br /><br />
```
Koen1999: Shouldn't the line break be just `<br>` ?

rinkp: > Shouldn't the line break be just `<br>` ?

That would have my preference, but due to the existing use of `<br />` I opted for that one. Possibly it was once decided that `<br />` looked better or was more compatible with some mail clients.
I think consistency is even more important than preferences, so if we were to change it, it should be changed everywhere.

Koen1999: Yeah, I guess we can do it in some other PR at some point.

PR #970, opened 2019-12-28
Rename a few auth headers and add some more

jszanto:
This is something the CBC would like to use for future authentication integrations

### DIFF
--- a/module/Decision/src/Decision/Service/Member.php
+++ b/module/Decision/src/Decision/Service/Member.php
@@ -45,20 +45,7 @@ class Member extends AbstractAclService
             return null;
         }
 
-        $memberships = [];
-        foreach ($member->getOrganInstallations() as $install) {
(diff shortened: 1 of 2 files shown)


Koen1999:
I'm not completely familiar with the way these headers work, but I'd like some clarification.
Aren't headers visible to anyone sniffing around on the network?

wsinnema:
The routes that have the headers will be blocked by the reverse proxy. It is
therefore only visible for other webapps running on the same server, and the
reverse proxy itself.

Koen1999:
Alright, good to know. Then we can merge this PR once it is tested and the reverse proxy is properly configured.

jszanto:
The headers are only visible if you are logged in, so if these routes are exposed (which they currently are) then you can only see the headers for yourself. And for people sniffing around on the network, TLS prevents that.

This also has been tested already btw.

Issue #780, opened 2017-12-09
Use Dutch date notation in agenda item badges


gafderks:
Make the notation depend on the current language.

English (as it currently is): 
![image](https://user-images.githubusercontent.com/7315457/33799508-94515d28-dd2d-11e7-9b84-30bc7119004c.png)

When the website is switched to Dutch:
![image](https://user-images.githubusercontent.com/7315457/33799518-b153f08e-dd2d-11e7-8f44-aecf7fc720e2.png)

jszanto:
Related: #526

PR #974, opened 2020-01-29
Remove locale overrides in `dateFormat()` view helper

armandduijn:
This PR removes the unnecessary locale parameter in calls to the `dateFormat()` view helper. The view helper should rely on the default locale set in the bootstrap code of the Application module.

We shouldn't rely on the default locale in _activity emails_. These contain Dutch and English texts so the dates should be formatted accordingly.

**Note**: Relies on code in #965 that [sets the global locale](https://github.com/GEWIS/gewisweb/pull/965/files#diff-627ee89ffa9e0e61c9dab22d2c8bae21R35).


armandduijn:
I checked if #780 could also be fixed with this PR but alas. The view helper only supports a limited number of patterns and I couldn't find a way to extend it with a custom pattern. The only solution seems an (ugly) if/else statement.

Koen1999:
Looks good. I tested it, it works as expected.

PR #982, opened 2020-02-09
Fix NonUniqueResultException when retrieving an organ based on its abbreviation

### LINKED ISSUE #978: Committee page not accesible due to duplicate DB entries
Some committee pages, like C4, are not accesible because they were created more than once.
We should just take the most recent committee in the mapper to solve this issue.

tomudding:
This fixes #978. 

Previously Doctrine's ORM would throw an `NonUniqueResultException` when two (or more) organs shared the same abbreviation. This limits the selection of an organ based on its abbreviation to the latest record. In other words, now the organ with the most recent `foundationDate` is retrieved.

### DIFF
--- a/module/Decision/src/Decision/Mapper/Organ.php
+++ b/module/Decision/src/Decision/Mapper/Organ.php
@@ -121,6 +121,8 @@ class Organ
             $qb->andWhere('o.type = :type')
                 ->setParameter('type', $type);
         }
+        $qb->orderBy('o.foundationDate', 'DESC')
+            ->setMaxResults(1);
 
         $qb->setParameter('abbr', $abbr);
 
--- a/module/Decision/src/Decision/Mapper/Organ.php
+++ b/module/Decision/src/Decision/Mapper/Organ.php
@@ -104,12 +104,18 @@ class Organ
     /**
      * Find an organ by its abbreviation
      *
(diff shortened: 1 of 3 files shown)


module/Decision/src/Decision/Mapper/Organ.php
```diff
@@ -121,6 +121,8 @@ public function findByAbbr($abbr, $type = null)
             $qb->andWhere('o.type = :type')
                 ->setParameter('type', $type);
         }
+        $qb->orderBy('o.foundationDate', 'DESC')
+            ->setMaxResults(1);
```
kokx: I'm note sure about the side-effects here. The `findByAbbr` method might be used in places where we don't use this assumption.

At the very least you should document the assumption you're using here in the docblock. And you should check if it is not (indirectly) used anywhere that this might cause issues. If that is the case, fix those issues or add a parameter to make this behavior opt-in.

tomudding: @kokx `findByAbbr` is only used by `findOrganByAbbr`, which is again only used once (in `OrganController` for the `organAction`). So it should not be a problem.

However, I will add an additional parameter to the method to make it opt-in and explain its functionality in the docblock.

PR #983, opened 2020-02-09
Remove addresses and add contact link in footer

### LINKED ISSUE #971: Replace adres in footer by link to contact page
This prevents confusion about the address since there are 3 addresses and the footer only contains 2

tomudding:
This fixes #971.

**Old:**
GEWIS-WEB #971 Old

**New:**
GEWIS-WEB #971 New

It might even be worth it to remove the phone number and e-mail address (it looks a bit better), however, I like the idea of having that information on every page (instead of having to visit another page).


rinkp:
If I may give my personal opinion on this, I'd like to say that I don't think it's really clear that it is a hyperlink to the contact page. 

Perhaps it can be changed to something like "Addresses" if it's only for that information. I'd combine that with for example the [map-marked-alt](https://fontawesome.com/icons/map-marked-alt?style=solid) icon. 

I haven't yet visualized how that'd look like, so maybe that isn't the best option either. 


I do think it is important to make all addresses known by changing the footer to have a link to all of them.

Koen1999:
I agree that removing the phone number and mail address is a good idea as it is also included on the contacts page. But we can wait for other opinions

jszanto:
I think removing the phone number and email address also will make it more clear that it is a hyperlink to the contact page

tomudding:
After removing the telephone number and e-mail address it looks like this:
GEWIS-WEB #971 Newer

I have tried things like `Contact us` / `Neem contact op` to replace the `Contact`, however, I found that this does not make it better. Furthermore, looking at some popular Dutch websites they all use 'Contact'.

I will update the translation files (not that the translation is different) when I have `gettext` working.

PR #985, opened 2020-02-10
Fix missing members when getting 'latest' organ

### LINKED ISSUE #984: Organs do not show all members

tomudding:
This fixes #984.

I introduced a bug in #982, which let to only one (or zero) members being shown when viewing organs. This was due to a misunderstanding of the query used, which I then forcefully limited to one result (i.e. one member).

I have "fixed" this by getting all results and only return the first - most recent - instance of an organ.


kokx:
I think it could also have been done using a `HAVING` clause, maybe that is worth exploring since that might make it simpler. Though for now I guess this is OK.

PR #986, opened 2020-02-11
Removed Grunt and added node-sass

kokx:
When creating a self-contained docker container for the website (needed
for migration to PHP 7 and keycloak), we had some issues to compile the
CSS, because ruby-sass is relatively complicated to install.

Since the entire mess was quite complicated anyways, having to install
npm, grunt and ruby-sass, we think it is better to move to node-sass
instead. This required renaming the main.scss to gewis-theme.scss, which
isn't really a big change.

So, the new procedure for compiling the scss is as follows:

- Have npm installed
- Execute `npm install`
- Execute `npm run scss`
- ???
- Profit!

For ease during development, we added the command `npm run watch`, which
is equivalent to the `grunt watch` command.


armandduijn:
Are there any downsides to switching to node-sass? I've only every used ruby-sass for compiling sass.

kokx:
Well, [ruby-sass is officially deprecated](https://twitter.com/SassCSS/status/978367843564695552). node-sass uses libsass, [which is the reference/official sass library](http://sassbreak.com/ruby-sass-libsass-differences/). So I'd say that using node-sass is actually the way to go forward (or whatever libsass frontend is handy for your project).

So I guess there is mainly an upside: removing a deprecated dependency.

armandduijn:
Sounds good. I'll test it on my local setup today/tomorrow and give my okay if it works.

PR #988, opened 2020-02-13
[WIP] Docker production image

Kevinjil:
This image can be run in a production environment. An example docker-compose is included, but we intent to not use this in our production setup. A.o. the MySQL database will not run inside Docker.

This depends on, and should be rebased after, #986.

The `./web orm:schema-tool:update` is deliberately not put inside the image / Dockerfile, as we do not want unattended Databaes updates.


Koen1999:
Do we still need to rebase or will git take care of that automatically?

Koen1999:
Is config/autoload/doctrine.global.php intended for production?
If so, we should disable automatically generating proxies in there.

jszanto:
I assume it is not, since you wouldn't run your database in docker in production and you would not hardcode your database credentials in that file.

Koen1999:
After resolving conflicts, I tested it, but it's not working for me. I also verified wtih the version from commit 784307d406416b5037e6578f8eca692cb3935f26

Just to verify. To deploy, you just go clone, go into docker/prod and run docker-compose up -d. Then you create the database schema and it should work, right?

I get the following error message:
```Fatal error: Uncaught exception 'Exception' with message 'DateTime::__construct(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone.' in /code/vendor/monolog/monolog/src/Monolog/Handler/RotatingFileHandler.php:50 Stack trace: #0 /code/vendor/monolog/monolog/src/Monolog/Handler/RotatingFileHandler.php(50): DateTime->__construct('tomorrow') #1 /code/module/Application/Module.php(122): Monolog\Handler\RotatingFileHandler->__construct('data/logs/gewis...', 10, 'INFO') #2 [internal function]: Application\Module->Application\{closure}(Object(Zend\ServiceManager\ServiceManager), 'logger', 'logger') #3 /code/vendor/zendframework/zendframework/library/Zend/ServiceManager/ServiceManager.php(939): call_user_fu in /code/vendor/zendframework/zendframework/library/Zend/ServiceManager/ServiceManager.php on line 946```

PR #990, opened 2020-02-18
Updated frontpage panorama

Koen1999:
This is an update requested by the board. See email.


Koen1999:
It looks as follows on desktop: 
afbeelding


And it looks as follows on movile:
afbeelding

tomudding:
It might also be worth it to check out [img srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images). That way we can have multiple files of different sizes (compressed) to make sure that different screen sizes still get a good image.

You would get something like this (the widths for the images need to be adjusted):
```
<img src="img/panorama_small.jpeg" sizes="100vw" srcset="
    img/panorama_small.jpeg 512w,
    img/panorama_medium.jpeg 1024w,
    img/panorama_large.jpeg 2048w,
    img/panorama_huge.jpeg 4096w,
    img/panorama_mega.jpeg 5120w">
```

Koen1999:
> 
> 
> Looks good, but the image should be compressed a bit. 11mb is a little large for a header image I think.

I have compressed the 11 MB to 500 KB.

> 
> 
> It might also be worth it to check out [img srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images). That way we can have multiple files of different sizes (compressed) to make sure that different screen sizes still get a good image.
> 
> You would get something like this (the widths for the images need to be adjusted):
> 
> ```
>      img/panorama_small.jpeg 512w,
>     img/panorama_medium.jpeg 1024w,
>     img/panorama_large.jpeg 2048w,
>     img/panorama_huge.jpeg 4096w,
>     img/panorama_mega.jpeg 5120w">
> ```

I didn't feel like working it out this nicely, but it's a good solution too.

PR #1006, opened 2020-06-13
[WIP] Add day and night mode

### LINKED ISSUE #1004: Add a night mode/theme for the website
This will introduce SCSS themes which can be changed by clicking a button/link on the website. The setting is saved in local-storage to retain the setting through multiple sessions (on the same device). To make it retain its current setting across multiple devices we would have to save the state to the database, but this is currently not needed.

By introducing the themes I will check compatibility with the WCAG. We should be able to be level AA compliant, but probably not level AAA (due to the additional contrast restrictions).

To give you an impression of what I think it is going to look like:
GEWIS-WEB - Dark Mode Mock-Up EDITED

tomudding:
Adds a button to the main navigation bar which is used to switch themes. It also shows what the current theme is.

There are several ways the theme of the website can be changed. By default the day theme is selected. If the OS of the user prefers a night theme, the night theme is applied. Finally, if the user has a preferred theme this will overwrite the two previous themes.

This closes #1004.

### DIFF
--- a/module/Application/view/layout/layout.phtml
+++ b/module/Application/view/layout/layout.phtml
@@ -65,19 +65,21 @@
 
                 if (!(typeof systemPreferredTheme === 'undefined' || systemPreferredTheme === null)) {
                     if (systemPreferredTheme.matches) {
+                        document.body.className = 'theme-night';
                         document.getElementById("theme-icon").className = "fas fa-moon";
                     } else {
+                        document.body.className = 'theme-day';
                         document.getElementById("theme-icon").className = "fas fa-sun";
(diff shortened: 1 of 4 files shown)


tomudding:
I think this should wait till Bootstrap 5 has been released. We could upgrade to 4 (released in 2018), but either way we have to rewrite a lot of things as 3 does not properly support themes. The best support theming will probably be released in Bootstrap 6, but that will take a few years to be released.

tomudding:
Our old versions do not support this, for the time being it is better to close this till we have better support.

PR #1009, opened 2020-06-21
implement job deletion

jszanto:


Koen1999:
Is it really desired to remove jobs? I think making them inactive is preferred.

jszanto:
Yes, because leaving them there clutters up the job list in the admin interface. And there is no need to store the job description and other information.

tomudding:
Looks good (and works as expected)! I assume you still have to do this, but there still a few strings that need to be translated.

Koen1999:
Then why did we implement inactive in the first place? @jszanto

tomudding:
@Koen1999 I expect the rationale for implementing the ability to deactivate a job is to ensure that jobs can be entered into the system beforehand and then be activated when needed, or to temporarily disable (read: hide) a job instead of completely removing it.

And now that have been removing (i.e. deactivating) jobs for the board for some time (which are never activated again) the effects are visible of not deleting them. They are still visible in the company packages, they are still in the database.

If a company would like to have the same job again, I would prefer creating a new version with the same slug. Otherwise, you have to make the slug longer or you have to add digits to differentiate the jobs from other (in)active jobs.

And an additional things which I think is also nice by implementing this, we will get fewer e-mails from the board!

Koen1999:
> 
> 
> @Koen1999 I expect the rationale for implementing the ability to deactivate a job is to ensure that jobs can be entered into the system beforehand and then be activated when needed, or to temporarily disable (read: hide) a job instead of completely removing it.
> 
> And now that have been removing (i.e. deactivating) jobs for the board for some time (which are never activated again) the effects are visible of not deleting them. They are still visible in the company packages, they are still in the database.
> 
> If a company would like to have the same job again, I would prefer creating a new version with the same slug. Otherwise, you have to make the slug longer or you have to add digits to differentiate the jobs from other (in)active jobs.
> 
> And an additional things which I think is also nice by implementing this, we will get fewer e-mails from the board!

The last point you mention is definitely why we want this, indeed.
I agree that it would make the DB a bit less cluttered.
I'll test and approve if it works.

Koen1999:
It probably used to work before the label adding PR. I'll see if I can modify the constraints
`XXXXX.ERROR: exception 'PDOException' with message 'SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`XXXXX`.`JobLabelAssignment`, CONSTRAINT `FK_42CB55B5BE04EA9` FOREIGN KEY (`job_id`) REFERENCES `Job` (`id`))'`

Koen1999:
@jszanto It is working for active jobs, but not for inactive jobs.

tomudding:
@Koen1999 @jszanto It works now in conjunction with the two other PRs.

Koen1999:
> 
> 
> @Koen1999 @jszanto It works now in conjunction with the two other PRs.

Wait, nothing has changes to this PR? How is it working for inactive jobs now then?

tomudding:
> 
> 
> > @Koen1999 @jszanto It works now in conjunction with the two other PRs.
> 
> Wait, nothing has changes to this PR? How is it working for inactive jobs now then?

Because when you tested this, the correct code for inactive jobs (the PR) was not present. And perhaps that was not even the issue, but I can no longer reproduce any errors relating to deleting jobs. By including this code, this PR works. So the order of merging the PRs is somewhat important (inactive jobs -> deleting jobs -> labels for jobs).

PR #1010, opened 2020-06-22
Sort photos based on date

### LINKED ISSUE #936: Photos are not sorted in chronological order
In the new photo viewer

tomudding:
This fixes #936.

### DIFF
--- a/module/Photo/src/Photo/Model/Album.php
+++ b/module/Photo/src/Photo/Model/Album.php
@@ -14,7 +14,7 @@ use Zend\Permissions\Acl\Resource\ResourceInterface;
  */
 class Album implements ResourceInterface
 {
-
+    
     /**
      * Album ID.
      *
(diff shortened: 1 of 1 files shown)


Koen1999:
I'm pretty sure this will work. But wouldn't it be more logical to have the newest photos up front? (i.e. reverse the ordering)
What do you think?

tomudding:
@Koen1999 If I remember correctly we got remarks from people who tested the new photo viewer that there was no order at all. This resolved that, but we can easily change it do `DESC` if needed.

In addition to this, we should also move forward with making the "new" photo viewer the default photo viewer. As long as we do not force members to use the new photo viewer we will also not know what issues arise when using it.

jszanto:
If you are viewing an album of an activity I think it makes more sense to have them in ascending order. In that way you can chronologically view the photos of the activity. Don’t see why you would want to start at the end.

Koen1999:
> 
> 
> If you are viewing an album of an activity I think it makes more sense to have them in ascending order. In that way you can chronologically view the photos of the activity. Don’t see why you would want to start at the end.

Well. Albums are sorted in the same way. Essentially, people are mostly interested in the most recent thing. But I agree that for activities it may make more sense to begin at the beginning :/

Koen1999:
> @Koen1999 If I remember correctly we got remarks from people who tested the new photo viewer that there was no order at all. This resolved that, but we can easily change it do `DESC` if needed.
> 
> In addition to this, we should also move forward with making the "new" photo viewer the default photo viewer. As long as we do not force members to use the new photo viewer we will also not know what issues arise when using it.


I agree. Let me quickly change it and approve.
We should however also fix making the photo viewer work on the photos of a member. Probably adding a route or something

(sorry for editing instead of quoting...)

Anyway, perhaps ascending is better anyway.

PR #1017, opened 2020-07-16
Enable new photo viewer on Member's pages

### LINKED ISSUE #935: New photoviewer doesn't work on member page
When on a page https://gewis.nl/photo/member/<member_id>, there is a link to the new photo viewer. However, clicking on that link results in a 404 error

tomudding:
This enables the photo viewer on Member's pages (and thus fixes #935) and fixes the "try this out"-link.

After this we should (as discussed today) make the new photo viewer the default and have the old photo viewer as a backup (separate PR).

### DIFF
--- a/module/Photo/view/photo/album/member-new.phtml
+++ b/module/Photo/view/photo/album/member-new.phtml
@@ -1,0 +1,211 @@
+<?php
+// set title
+$this->headTitle($album->getName());
+$currentAlbum = $album;
+while(!is_null($currentAlbum->getParent())) {
+    $this->headTitle($currentAlbum->getParent()->getName());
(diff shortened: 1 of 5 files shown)


module/Photo/view/photo/album/member-new.phtml
```diff
@@ -0,0 +1,211 @@
+<?php
```
jszanto, line 1: This seems like a lot of code duplication, can't we re-use the existing view?

tomudding: We can certainly do that.

jszanto (approved):
Looks good, assuming it was all tested and working.

PR #1019, opened 2020-07-16
Show URL for jobs when set

tomudding:
Once again, URLs are shown on Job pages when properly set.

The board / company admins should be made aware that affiliate links are not supposed to be placed here and that the link is an alternative for the job attachment.

### DIFF
--- a/module/Company/view/company/company/jobs.phtml
+++ b/module/Company/view/company/company/jobs.phtml
@@ -62,6 +62,9 @@ $this->headTitle($category->getPluralName());
                             <p class="job-email">
                                 <a href="mailto:<?= rawurlencode($job->getEmail()) ?>"><?= $job->getEmail() ?></a>
                             </p>
+                            <?php if ($job->getWebsite() !== ""): ?>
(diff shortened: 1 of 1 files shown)


module/Company/view/company/company/jobs.phtml
```diff
@@ -62,6 +62,13 @@ $this->headTitle($category->getPluralName());
                             <p class="job-email">
                                 <a href="mailto:<?= rawurlencode($job->getEmail()) ?>"><?= $job->getEmail() ?></a>
                             </p>
+                            <?php if ($job->getWebsite() !== ""): ?>
+                                <p class="job-website">
+                                    <a href="<?= $escaper->escapeHtmlAttr($job->getWebsite()) ?>" rel="noreferrer">
```
Koen1999, line 67: I don't think escape HTML is neccesary here, but it's not blocking.

tomudding: @Koen1999 Yes, escaping is necessary, otherwise stored XSS is possible. I know this is highly unlikely, but I do not want the website to break because some company has a funny URL or one of the admins goes rogue.

PR #1028, opened 2020-09-01
Reduce size of Athenaeum images

tomudding:
Reduce the size of the Athenaeum images as requested by Wouter.

The AB logo is reduced by 79% in size and the banner's size is reduced by 97%.


Koen1999 (approved):
Looks good!

PR #1032, opened 2020-09-18
Added a disclaimer to the activity view

Koen1999:
As per request of Wouter.

### DIFF
--- a/module/Activity/src/Activity/Model/ActivityTranslation.php
+++ b/module/Activity/src/Activity/Model/ActivityTranslation.php
@@ -101,7 +101,6 @@ class ActivityTranslation
      */
     protected $fields;
 
-    // TODO -> where can i find member organ?
     protected $organ;
 
     protected $isMyFuture;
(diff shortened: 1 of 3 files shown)


tomudding:
This is more or less a stylistic remark, but I would think that putting it in the box is better. Or at least make it a bit smaller.

I understand that Wouter wants it this way, but I would only show it when you are subscribed. Because it is already in the Activity Policy (minus the e-mail addresses), and the agreement message disappears when you are subscribed. And the board should inform users about this.

Kevinjil (approved):

Koen1999:
> 
> 
> This is more or less a stylistic remark, but I would think that putting it in the box is better. Or at least make it a bit smaller.
> 
> I understand that Wouter wants it this way, but I would only show it when you are subscribed. Because it is already in the Activity Policy (minus the e-mail addresses), and the agreement message disappears when you are subscribed. And the board should inform users about this.

It also informs user of the point of contact in case of other questions about the activity they may have before subscribing.

PR #1053, opened 2021-01-18
Added CSRF token to login form

Koen1999:


tomudding (approved):
I mentioned that I wanted to change this to get rid of the `echo`. However, this ironically makes it longer when using the short echo tags.

PR #1054, opened 2021-01-18
Attempt at creating a good docker compose file for production and development

Koen1999:
Inspired by #988

Still not quite perfect...


Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+    && docker-php-ext-configure intl \
+    && docker-php-ext-install \
+        calendar \
+        exif \
+        gd \
+        intl \
+        mbstring \
+        pdo \
+        pdo_mysql \
+        soap \
```
jszanto, line 32: No longer needed

Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM node:14-alpine as node-build
+WORKDIR /code
+
+ADD package.json package-lock.json ./
+RUN npm install
+
+ADD public/ ./public/
+RUN npm run scss
+
+FROM php:5.6-fpm as php-target
```
jszanto, line 10: Consider using the alpine php image, it is a lot smaller

Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+        libicu-dev \
+        libjpeg62-turbo-dev \
+        libmagickwand-dev \
+        libpng-dev \
+        libxml2-dev \
+        unzip \
+        zlib1g-dev \
+    && docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
+    && docker-php-ext-configure intl \
+    && docker-php-ext-install \
```
jszanto, line 24: Don't forget to add something like opcache for production

Dockerfile
```diff
@@ -0,0 +1,59 @@
+FROM node:14-alpine as node-build
```
jszanto, line 1: I would recommend splitting this in two docker files. Or use stages such that you have a development stage which still includes all the tools such as npm and composer.

build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
jszanto, line 1: This can be done in the dockerfile, doesn't need to be a seperate bash script

jszanto (changes requested):
General remarks:
- Consider putting many configuration options, such as database credentials in env variables for production
- Why did the permissions on so many files change? Can you revert that?

Dockerfile
Koen1999: I think that the idea of the current setup is that is should be faster. I was thinking about making a docker-compose override file for development. In this I could just re-add some stuff that is useful.

Dockerfile
Koen1999: Which of the four?

PR #1056, opened 2021-01-27
Created production image for docker to be deployed with Portainer

Koen1999:
Still WIP


Koen1999:
A current blocking issue is that the environment variables are substituted into the config file (config/autoload/doctrine.local.php) at build time instead of at the time of deployment. I am not sure how to fix this.

Koen1999:
TODO, take #1064 into account

Koen1999:
@kokx @Kevinjil 
Does one of you know why we make use of docker to install php extensions while we also make use of composer to manage these including the versions?
I have been thinking about this and I think it makes more sense to stick with only composer.

tomudding:
> Does one of you know why we make use of docker to install php extensions while we also make use of composer to manage these including the versions?

@Koen1999 Packagist does not have all (or any) of the PHP extensions. It does have [platform packages](https://getcomposer.org/doc/01-basic-usage.md#platform-packages) for the extension. These are used to compare versions, however, this can be very inconsistent. This is the reason we have the following two lines in `composer.json`:

    "ext-gd": "^0.0.0",
    "ext-exif": "^0.0.0",

We cannot install them via Composer, but we can make sure they are present.

PR #1059, opened 2021-02-07
Company fix

### LINKED ISSUE #1058: English vacancy page is broken
The page for vacancies is broken on the English version of the website: https://gewis.nl/career/vacancies

The Dutch version still works and even the English versions of the internship, graduation page, etc are also still working. However on this page there are no more GEWIS style attributes and the company logos are all over the place..

Koen1999:
This should help fix #1058

### DIFF
--- a/module/Company/src/Company/Model/Company.php
+++ b/module/Company/src/Company/Model/Company.php
@@ -42,7 +42,7 @@ class Company // implements ArrayHydrator (for zend2 form)
      * @ORM\Column(type="string")
      */
     protected $slugName;
-    
+
     /**
      * The company's contact's name.
      *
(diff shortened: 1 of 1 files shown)


Koen1999:
Or it can help prevent instances like this in the future

tomudding (commented):
I do not think we should call this a _fix_. This simply allows us to be "notified" about the underlying problem. However, I do concede that changing the logging is a good idea, but that is not really a solution to the problem.

Perhaps it is wise to solicit advice from the others before applying the changes you have made here. I would argue that we (perhaps in addition to changing the logging) should prevent jobs from being created in a language that is not supported by the company.

PR #1063, opened 2021-02-24
Performed more codeclimate improvements to some of the other modules

Koen1999:
The Activity and Photo module have been left untouched to prevent merge conflicts with some of the open PRs.


module/Application/view/partial/main-nav.phtml
```diff
@@ -1,36 +1,6 @@
...
-            }
-            @keyframes AnimationName {
-                0%{background-position:0% 50%}
-                50%{background-position:100% 50%}
-                100%{background-position:0% 50%}
-            }
-        </style>
-    <?php endif ?>
-<?php endif ?>
 <nav class="navbar navbar-gewis navbar-static-top rainbow" role="navigation">
```
rinkp, line 4:
```suggestion
<nav class="navbar navbar-gewis navbar-static-top" role="navigation">
```

Unless we want to let users reintroduce the rainbow (with e.g. [Stylus](https://addons.mozilla.org/en-US/firefox/addon/styl-us/)), this one can be removed safely as well I think

rinkp (commented):
(disclaimer: I was just wondering what the code style was going to be when I saw this minor thingy; I haven't checked the code)

module/Application/view/partial/main-nav.phtml
Koen1999: Yes, this is indeed true. It's still WIP, so hold up on your reviews.

Koen1999:
Obviously the code still isn't perfect. But is has improved.

module/Company/src/Company/Model/Company.php
```diff
@@ -360,7 +360,7 @@ public function getNumberOfActiveJobs($category = null)
     public function getNumberOfExpiredPackages()
     {
         return count(array_filter($this->getPackages()->toArray(), function ($package) {
-            return $package->isExpired(new DateTime);
+            return $package->isExpired(new DateTime());
```
Koen1999, line 363: Does this change have the same behaviour? The old code looked like a bug to make anyway

Koen1999 (commented):
I think we lost quite some lines of code due to all this whitespace XD

module/Company/src/Company/Model/Company.php
tomudding: There is no difference in behaviour between the two versions. Hence it is a choice for the developer whether to use them or not.

However, looking at our codebase we use parentheses for our own models. Furthermore, if we want to keep adhering to PSR, we must take into account that PSR-12 dictates that parentheses must be present when instantiating a new class. Additionally, all official PHP sources also use parentheses (iirc).

Koen1999: With parentheses makes way more sense to me anyway

module/Frontpage/src/Frontpage/Controller/PollController.php
```diff
@@ -51,7 +51,7 @@ public function obtainPoll()
      */
     public function voteAction()
     {
-        $pollId = (int) $this->params('poll_id');
+        $pollId = (int)$this->params('poll_id');
```
tomudding, line 54: Keeping the space here improves readability in my opinion.

module/Frontpage/src/Frontpage/Model/PollComment.php
```diff
@@ -124,7 +123,7 @@ public function getAuthor()
     /**
      * Set the author.
      *
-     * @param string $author;
+     * @param string $author ;
```
tomudding, line 126: Semi-colon can be removed (and thus no space required).

module/User/Module.php
```diff
@@ -39,10 +39,10 @@ public function onBootstrap(MvcEvent $e)
 
         // this event listener will turn the request into '403 Forbidden' when
         // there is a NotAllowedException
-        $em->attach(MvcEvent::EVENT_DISPATCH_ERROR, function($e) {
+        $em->attach(MvcEvent::EVENT_DISPATCH_ERROR, function ($e) {
             if (($e->getError() == 'error-exception') &&
-                    ($e->getParam('exception', null) != null) &&
-                    ($e->getParam('exception') instanceof NotAllowedException)) {
+                ($e->getParam('exception', null) != null) &&
+                ($e->getParam('exception') instanceof NotAllowedException)) {
```
tomudding, line 45: Perhaps we should do this as described in the second example of [PSR-12 5.1](https://www.php-fig.org/psr/psr-12/#51-if-elseif-else). Improves the readability.

module/User/src/User/Model/User.php
```diff
@@ -140,13 +139,13 @@ public function getMember()
...
+        $names = [];
 
         foreach ($this->getRoles() as $role) {
-            $roles[] = $role->getRole();
+            $names[] = $role->getRole();
         }
 
-        return $roles;
+        return $names;
     }
```
tomudding, line 149: This isn't incorrect, but I don't think we will benefit from changing this.

module/User/view/user/api-admin/index.phtml
```diff
@@ -9,16 +9,16 @@
...
             <td><?= $token->getName() ?></td>
             <td><?= $token->getToken() ?></td>
             <td><a class="btn btn-xs btn-danger" href="<?= $this->url('user_admin/api/remove', [
-    'id' => $token->getId()
-]) ?>"><span class="fas fa-times"></span> <?= $this->translate('Remove') ?></a></td>
+                    'id' => $token->getId()
+                ]) ?>"><span class="fas fa-times"></span> <?= $this->translate('Remove') ?></a></td>
         </tr>
-        <?php endforeach ?>
+    <?php endforeach ?>
```
tomudding, line 21: Perhaps this whole block should be indented.

tomudding (commented):
I have reviewed most if not all files. It is looking very promising, however, I think there are a few things that should be done differently. Please let me know what you think!

module/Company/src/Company/Model/CompanyBannerPackage.php
```diff
@@ -46,13 +42,4 @@ public function getType()
...
     }
-
-    /**
-     * Constructor.
-     */
-    public function __construct()
-    {
-        parent::__construct();
-        // todo
-    }
```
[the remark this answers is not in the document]
Koen1999: It only inherits the functionality of the base class. So removing this does not change the functionality. It just gives less code. and less code is better code :)

module/User/src/User/Model/User.php
Koen1999: names in this method shadowed a variable in the class it is contained in

module/Company/src/Company/Model/CompanyBannerPackage.php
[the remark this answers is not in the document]
tomudding: > less code is better code

May I suggest we start following the [No Code](https://github.com/kelseyhightower/nocode) principle? /s

module/User/Module.php
Koen1999: Good suggestion

tomudding (approved):

PR #1065, opened 2021-02-25
Remove Athenaeum promotional material

tomudding:
Removes all Athenaeum promotional material as requested by Wouter.


Koen1999 (approved):
We should check with the CBC that gewis.nl/books and gewis.nl/boeken are also adjusted.

PR #1066, opened 2021-03-11
Added Extensions To Composer.json

Powers-RC:
First attempt at any open source contributions so feed back is appreciated. I believe this is what was requested in the issue. I did notice that the pdo_pgsql extension was not available in the fpm environment and it was configured in the php.ini. Not sure if is suppose to be, and didn't want to go outside the request. Willing to help more.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,8 @@
     "require": {
         "php": ">=5.4.0",
         "ext-gd": "^0.0.0",
+        "ext-pgsql": "*",
+        "ext-pdo": "*",
         "ext-exif": "^0.0.0",
         "ext-imagick": "^3.1",
         "zendframework/zendframework": "2.4.*",


kokx:
Looks like spam.

Powers-RC:
Ouch not spam. Guess I didn't understand what you wanted. I usually configure these things in the php.ini not in composer.json. Tried to add it via composer but those packages aren't available.

Powers-RC:
I think I see what is being requested now, the composer.json threw me off. I will try and reproduce the error, update the Dockerfile with the extensions, and test the docker deployment tonight.

kokx:
@Powers-RC I think you might be at the wrong place. We welcome contributions from the right parties, but looking at your github profile you don't seem like you would have any serious interest in this project.

It might be that I'm wrong about this, and in that case, definitely hop in our email (you should know where to find it in that case) and introduce yourself. But we definitely are not the kind of project accepting random fixes from people we do not know.

Powers-RC:
@kokx  thanks for the response. Didn't mean to intrude, just trying to spend a little more of my free time in the OSS space. I will make sure to reach out to the lead contributors prior to making any PR's in the future.  Thx.

PR #1067, opened 2021-03-21
Improved code climate of the activity-overhaul branch

Koen1999:


jszanto (approved):
Nice, improvements to the coding style are always good!

tomudding (approved):
As previously mentioned in https://github.com/GEWIS/gewisweb/pull/1063#discussion_r582760199 I think that keeping the spaces between `(int)` and what comes after it is a must-have for readability. I'll merge this into the activity-overhaul branch but with that reverted.

PR #1079, opened 2021-04-04
Modified the main nav login to refer to the login page

Koen1999:
Since #1053 , the login form in the main nav is non-functional and refers you to the login page.
This fixes it by removing the form.

afbeelding

### DIFF
--- a/module/Application/view/partial/main-nav.phtml
+++ b/module/Application/view/partial/main-nav.phtml
@@ -35,47 +35,12 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
                     </ul>
                 </li>
                 <?php if (null === $this->identity()): ?>
-                    <li class="visible-xs pull-left">
+                    <li class="pull-right">
                         <a href="<?= $this->url('user') ?>">
                             <span class="fas fa-user"></span>
                             <span class="sr-only"><?= $this->translate('Login') ?></span>
+                            <?= $this->translate('Login') ?>
                         </a>
-                    </li>
-                    <li class="dropdown pull-left hidden-xs">
-                        <a href="<?= $this->url('user') ?>" class="dropdown-toggle" data-toggle="dropdown" role="button"
-                           aria-haspopup="true" aria-expanded="false">
-                            <span class="fas fa-user"></span>
-                            <span class="sr-only"><?= $this->translate('Login') ?></span>
-                            <span class="caret"></span>
-                        </a>
-                        <div class="dropdown-menu dropdown-login">
-                            <form action="<?= $this->url('user') ?>" method="post">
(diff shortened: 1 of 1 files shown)


module/Application/view/partial/main-nav.phtml
```diff
@@ -35,48 +35,13 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
                     </ul>
                 </li>
                 <?php if (null === $this->identity()): ?>
-                    <li class="visible-xs pull-left">
+                    <li class="pull-right">
                         <a href="<?= $this->url('user') ?>">
                             <span class="fas fa-user"></span>
                             <span class="sr-only"><?= $this->translate('Login') ?></span>
```
tomudding, line 41: This is for screen readers, however, as the dropdown has been removed it is no longer needed, as it will read `Login` twice. You might want to consider removing it.

module/Application/view/partial/main-nav.phtml
```diff
@@ -35,48 +35,13 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
...
+                            <?= $this->translate('Login') ?>
                         </a>
                     </li>
-                    <li class="dropdown pull-left hidden-xs">
```
tomudding, line 44: You'd have to check this on a mobile device, but there exists a possibility that there is not enough room to display both the icon and the text (because there is no distinction for different devices).

tomudding (commented):
Looks good, however, there are some minor things that may need to be changed / tested more.

module/Application/view/partial/main-nav.phtml
tomudding: I was referring to line 41, but apparently the comment changed itself.

module/Application/view/partial/main-nav.phtml
Koen1999: I checked and it fits on any normal sized phone

module/Application/view/partial/main-nav.phtml
Koen1999: yess, true!

tomudding (approved):

jszanto:
Why would we completely remove the (in my opinion very functional) dropdown login form instead of fixing the actual issue with it?

Koen1999:
> 
> 
> Why would we completely remove the (in my opinion very functional) dropdown login form instead of fixing the actual issue with it?

I agree that the dropdown login form was somewhat useful. But the reality was that this also caused code duplication. Having two seperate login forms is confusing and unnessesary. Plus, I think that sending a form to every page also introduces some overhead that code make the website slower.
Either way, having this broken form is not desired.

PR #1080, opened 2021-04-06
Created a working Docker image for production and development

Koen1999:
The development version is the production version + a mysql and phpmyadmin container.
To run the development version use: `docker-compose up` or `make rundev` if you want to build using local changes
docker-entrypoint.override.sh and docker-compose.override.yaml are only used by the development edition. The same goes for the php.ini
To deploy the production version, only use docker-compose.yaml. (environment variables should not be in a seperate file for portainer)
Copy .env.dist to .env and make changes where appropriate


Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+        libmagickwand-dev \
+        libmemcached-dev \
+        libpng-dev \
+        libpq-dev \
+        libxml2-dev \
+        unzip \
+        zlib1g-dev
+
+RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ \
+    && docker-php-ext-install \
```
jszanto, line 31: You should add something like opcache, will make things a lot faster. Also since the code doesn't change you can disable the opcache checks for changed files I think.

config/autoload/doctrine.local.development.php.dist
```diff
@@ -40,16 +40,16 @@ return [
             'orm_default' => [
                 // metadata cache instance to use. The retrieved service name will
                 // be `doctrine.cache.$thisSetting`
-                'metadata_cache'    => 'array',
+                'metadata_cache'    => 'memcached',
```
jszanto, line 43: Correct me if I'm mistaken, but this setup doesn't seem to include a memcached server; so how does this exactly work?

jszanto: Also, why would you want to cache these things in development?

docker/nginx/mime.types
```diff
@@ -0,0 +1,50 @@
+types {
```
jszanto, line 1: This seems like mostly the default config? In that case you don't need to specify it.

docker/nginx/fastcgi.conf
```diff
@@ -0,0 +1,21 @@
+fastcgi_param  SCRIPT_FILENAME    $document_root$fastcgi_script_name;
```
jszanto, line 1: This seems like mostly the default config? In that case you don't need to specify it.

docker/nginx/.htpasswd
```diff
@@ -0,0 +1 @@
+web:$apr1$mK6z5n.S$UWux7wb4NpHWT6YDnqrSx0
```
jszanto, line 1: Putting credentials in a public git repo does not seem like a great idea.

public/index.php
```diff
@@ -7,6 +7,11 @@
  */
 chdir(dirname(__DIR__));
 
+// For Docker:
+if (file_exists('/code')) {
```
jszanto, line 11: Why are these lines of code needed? I assume public/index.php is located in /code/public/index.php? So then it should work without this.

jszanto (changes requested):
Besides the comments on the code another note;
If you're using php-fpm you should override the default php fpm configuration and provide something more suitable for production. Otherwise you're going to have some performance issues in production most likely.

Dockerfile
Koen1999: I included memcached

config/autoload/doctrine.local.development.php.dist
Koen1999: Okay, for development it is not really relevant. I did install a memcached server in the docker container (see Dockerfile)

docker/nginx/fastcgi.conf
Koen1999: It indeed is default. Not sure if I had to specify it then, but we can leave it in I guess

docker/nginx/.htpasswd
Koen1999: I agree. Which is why it is not our password!

public/index.php
Koen1999: Well....to get the file sharing with the other container working properly, I created three volumes: /code, /public and /data.
/code contains symlinks to the other volumes. So index.php is located in /public/index.php.
It's not the most neat solution, but it works :)

jszanto: I think you removed the /code volume since it was not needed?
The public volume you can mount to /code/public and the data volume you can mount to /code/data

Dockerfile
jszanto: memcached and opcache are two completely different things..
memcached caches data, opcache caches php opcode so php doesn't have to parse your php files on every request

public/index.php
Koen1999: I indeed removed /code as a volume that was shared. Mmmmhm. Perhaps I could glue it back together

Koen1999: I fixed it!

config/autoload/doctrine.local.development.php.dist
Koen1999: I did however forgot to start the memcached server :p, but that's fixed now too.

Dockerfile
Koen1999: I did some googling, and it turns out they indeed are not the same! Also, thanks for the recommendations @rinkp . I would however suggest to increase `opcache.revalidate_freq` to for example `60` seconds. Zero implies that cache is always revalidated and this is probably not what you want in production. Next commit will include opcache.

rinkp: In production, we have `opcache.validate_timestamps = 0` and then `opcache.revalidate_freq` is ignored. In development we don't really care about performance, so there each request is fine.

Koen1999:
@jszanto I added php-fpm.conf as you suggested. Yet, I am not sure what kind of settings are useful besides the ones I added. Feel free to modify it.

I also created a Wiki page for how to deal with the Docker deployment already: https://github.com/GEWIS/gewisweb/wiki/Docker-Deployment

Dockerfile
Koen1999: I verified that the server response when requesting the frontpage is now just as fast as for the current production website. It used to be three times as high XD

jszanto: > In production, we have `opcache.validate_timestamps = 0` and then `opcache.revalidate_freq` is ignored. 

Indeed, in production your code should never change, so there’s no reason to every invalidate the cache.

docker/nginx/mime.types
jszanto: I believe the only types we have are images and pdf? Both which are included in the default config already.
Also zf should take care of setting the mime type for everything served through there.

jszanto: It is generally better to only include settings in Nginx config which you are overriding. Then there is no confusion about what’s default and what not. This also makes upgrades a lot easier.

.env.dist
```diff
@@ -0,0 +1,43 @@
+# Zend settings
+# Use production or development
+APP_ENV=production
+
+# Nginx settings
+# Use off or "\"Restricted\"" (or another escaped string); also required an entry in docker/nginx/.htpasswd
+NGINX_REQUIRE_AUTH=off
+
+# Mail settings for production
+DOCKER_MAIL_HOST=smtp.tue.nl
```
rinkp, line 10: image

Please note that `smtp.tue.nl` no longer is available without authentication from non-whitelisted hosts such as `gewis.win.tue.nl`, `gewisweb.win.tue.nl`  and `gewiscontainer.win.tue.nl`. The CBC created a workaround for this on `gewis.win.tue.nl` by routing it through the whitelisted host `gewismail.win.tue.nl` but for security reasons they do not allow unauthenticated mail delivery to non-GEWIS domains from hosts that aren't deprecated. 

For applications that send large volumes of mail, the CBC suggests to include a private Postfix TLS relay (such as [juanluisbaptiste/postfix](https://hub.docker.com/r/juanluisbaptiste/postfix)). An example config can be found at the bottom of this comment. This relay can then forward the mail to the authenticated mail server (optionally also updating To/From headers)

If you do not send large volumes of mail, or use an internal queue (you can reuse the connection for efficiency), you can also implement direct connections to `mail.gewis.nl:587` in your application to keep the setup simple. This also allows you to get feedback about the mail delivery in your application which you otherwise would not have. 

```yaml
postfix:
  hostname: web-postfix
  image: juanluisbaptiste/postfix
  environment:
    - SMTP_SERVER=smtp.gewis.nl
    - SMTP_USERNAME=service-web@gewis.nl
    - SMTP_PASSWORD=TODOTODO
    - SMTP_PORT=587
    - SERVER_HOSTNAME=web-postfix
  restart: always
  stop_grace_period: 90s            # We want to make sure all emails are sent before the data is potentially destroyed
  networks:
    web-network:
      aliases:
        - web-postfix
  expose:
    - "25"
```

Koen1999: Zend could not handle the TLS connection up to my expectations, so I went with the postfix setup after all. Thanks for the advice. I now tested it, and it works.

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,78 @@
...
+    gzip_min_length 1000;
+    gzip_buffers 4 8k;
+    gzip_types text/plain application/x-javascript text/css image/png image/jpeg image/gif image/x-icon image/svg+xml;
+    gzip_vary on;
+
+    add_header X-XSS-Protection         "1; mode=block";
+    add_header X-Content-Type-Options   nosniff;
+    add_header Content-Security-Policy  "default-src 'self' https://www.googleapis.com; script-src 'self' 'unsafe-inline'; img-src 'self' data:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.gstatic.com/; frame-src 'self'; object-src 'none'";
+
+    proxy_cache_path        /etc/nginx/cache levels=1:2 keys_zone=web_cache:1m max_size=256m inactive=60m use_temp_path=off;
```
jszanto, line 36: What exactly is being cached here? As far as I can tell you don't have any `proxy_pass` directives anywhere, so this won't do anything.
Also I don't think there is anything you want to cache because nginx is only serving static files and responses from php-fpm (which you definitely don't want to cache in this case).

docker/web/development/Dockerfile
```diff
@@ -0,0 +1,78 @@
+FROM php:5.6-fpm as gewisweb_web_development
```
jszanto, line 1: We could just make this an additional stage in the production docker file, saves a lot of code duplicaiton.

docker/web/production/crontab
```diff
@@ -0,0 +1,12 @@
...
+# │ │ ┌───────────── day of month (1 - 31)
+# │ │ │ ┌───────────── month (1 - 12)
+# │ │ │ │ ┌───────────── day of week (0 - 6) (Sunday to Saturday;
+# │ │ │ │ │                                       7 is also Sunday on some systems)
+# │ │ │ │ │
+# │ │ │ │ │
+# * * * * *  command to execute
+# Don't remove the empty line at the end of this file. It is required to run the cron job
+0 3 * * * ./web photo weeklyphoto >/dev/null 2>&1
+0 2 * * * apt-get update && apt-get upgrade -y
```
jszanto, line 12: Are you sure you want to do this? Seems like a good way to break your container at 2 am.

docker/web/development/Dockerfile
Koen1999: The development version has all tools like npm and composer installed. It makes little sense to install those in production, then copy some stuff over, and then install them again.

docker/web/production/crontab
Koen1999: Well, if you don't want to build images all the time to keep the container up to date, it seems like a good idea. I'm not sure if this is the best solution though.

docker/nginx/nginx.conf
Koen1999: Ah yes... I should've used `fastcgi_cache`. I can imagine a scenario in which (in the future) some responses may be cached. Not sure if a future version of Zend supports that though.

jszanto: If someone is logged in to the website you don't want the response to get cached, because then you might end up returning that to another user.

docker/web/production/crontab
jszanto: I think the code of the website gets updated frequently enough to cover all essential updates. Everytime you rebuild the container you also get all the updates.

Koen1999: That's true. But there are times we do not merge PRs into master for months.

jszanto: Which is probably fine since there usually aren't any important updates. And if there is a very critical (security) update you can rebuild it manually.

docker/nginx/nginx.conf
Koen1999: That's true. nginx determines what to cache and what not to cache based on the cache-related headers according to the documentation. Since all requests from the website currently return with something like this, we shouldn't worry about personal details getting cached.
afbeelding

docker/nginx/mime.types
Koen1999: SVG was not included to give one example

docker/web/production/docker-entrypoint.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+service cron start
+./web orm:generate-proxies
+memcached -m 256 -p 11211 -U 11211 -u www-data -d
```
jszanto, line 4: What happens if memcached crashes? The container won't restart because memcached is not running in the foreground.

I think we should either do it the docker way and run everything in separate containers OR just go ahead and run everything in a single container and run something like supervisor in it as well to make sure everything keeps running.

Koen1999: I agree. I will move memcached to a seperate container

docker/web/production/crontab
Koen1999: Alright, I will remove it

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,94 @@
...
+            fastcgi_hide_header         Pragma;
+            fastcgi_hide_header         Cache-Control;
+            add_header                  Cache-Control                       "private no-cache";
+            fastcgi_cache               web_cache;
+            fastcgi_cache_use_stale     error timeout http_500 http_503;
+            add_header                  X-Cache                             $upstream_cache_status;
+            add_header                  X-XSS-Protection                    $x_css_protection;
+            add_header                  X-Content-Type-Options              $x_content_type_options;
+            add_header                  Content-Security-Policy             $content_security_policy;
+        }
```
tomudding, line 92: This is related to the HTTP security headers. A good start in comparison to what we have now. We will have to manually make the filters stricter over time. There are two things I want to point out:

- `X-XSS-Protection` is functionally dead ([most major browsers have removed their XSS filters](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection)) and should be removed.
- `Referrer-Policy` is missing, although we have to make sure this does not interfere with any of things that use our referer.

I am not saying we should change this now, but it is something we should keep in mind.

There is also the new `Permissions-Policy`, however, it is not widely supported and should therefore not be used. This also applies to `CORP`, `COOP`, and `COEP`.

.dockerignore
```diff
@@ -0,0 +1,51 @@
...
+.env.dist
+.git/
+.gitattributes
+.idea/
+.mysql/
+.project
+.sass-cache/
+.settings
+Vagrantfile
+.vagrant/
```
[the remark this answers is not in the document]
Koen1999: This does prevent them from being committed again though :p

docker/nginx/nginx.conf
Koen1999: I agree we can indeed remove X-XSS protection if it is no longer supported.
What would you propose to do with the other policies? Feel free to make changes accordingly.

tomudding: > What would you propose to do with the other policies? Feel free to make changes accordingly.

The latter four we should not yet implement as they have limited support. `Referrer-Policy` is something we should look into at a later point (when we aren't making any major changes). Regarding the Content Security Policy, I would like to have it also report errors, but this requires additional infrastructure if we do it ourselves or an external company (e.g., Report-Uri).

So, apart from removing the `X-XSS-Protection` header I will not change anything else.

Koen1999: I agree that reporting could be useful indeed. But because of the required effort, I didn't do it for now.

Koen1999:
I just realised this PR needs something else and I am putting it here as a note so we don't forget.
I'm not sure how to do this myself and I think that @jszanto may be able to help with this specific thing.
We need to support the new photo viewer by integrating the image server into the docker-compose file and adjusting the config accordingly.

Koen1999:
> 
> 
> I just realised this PR needs something else and I am putting it here as a note so we don't forget.
> I'm not sure how to do this myself and I think that @jszanto may be able to help with this specific thing.
> We need to support the new photo viewer by integrating the image server into the docker-compose file and adjusting the config accordingly.

Okay, this has been resolved. I added a seperate image for the Glide image server. It also works when you deploy with `make rundev`.

Koen1999:
No more TODOs left :)

docker/nginx/nginx.conf
```diff
@@ -0,0 +1,125 @@
...
+
+    map $host $x_css_protection {
+        default "1; mode=block";
+    }
+    map $host $x_content_type_options {
+        default "nosniff";
+    }
+    map $host $content_security_policy {
+        default "default-src 'self' https://www.googleapis.com; script-src 'self' 'unsafe-inline'; img-src 'self' images.gewis.nl data:; style-src 'self' 'unsafe-inline' https://fonts.googleapis.com/; font-src 'self' https://fonts.gstatic.com/; frame-src 'self'; object-src 'none'; frame-ancestors 'self'; form-action 'self'";
+    }
```
tomudding, line 40: `https://www.googleapis.com` should not be in the `default-src` directive. It defeats the whole purpose of having a CSP in the first place if you add a website you don't manage to all non-existing directives.

To make it work, it should go under `connect-src`. However, you also need `gewis.nl` to load the templates (making it `conntect-src 'self' https://www.googleapis.com`).

**An even bigger problem is that it still won't work, because it requires `'unsafe-eval'` in the `script-src` directive.**

tomudding (changes requested):

docker/nginx/nginx.conf
Koen1999: I fixed the first.

Regarding the latter: we would need to change some things on the website to make it work then. Since we do not have a CSP at the moment, it's still an improvement.

tomudding: What happens without `unsafe-eval` in `script-src`:
image

Koen1999: Yeah, I noticed. It works now!

module/Photo/src/Photo/Service/Admin.php
```diff
@@ -2,7 +2,9 @@
 
 namespace Photo\Service;
 
+use _HumbugBoxa991b62ce91e\Nette\FileNotFoundException;
```
tomudding, line 5: I think this file is internal to PHPStan (or whatever program you used for the static analysis) and it should not be in the code. I tested whether or not it would cause any errors and from what I can tell it runs fine. However, `FileNotFoundException` is not used anywhere in the code and it should therefore be removed (on top of it being _weird_).

See comment from @kokx on [the closed PR here](https://github.com/GEWIS/gewisweb/pull/1095#discussion_r646050472) too.

Koen1999: This is indeed obviously a mistake. I'll remove it.

tomudding (approved):
Ready for deployment on Friday. Merging today to allow Koen to build the images in advance.

PR #1082, opened 2021-04-11
Revert "Removed some code artefact from a board gift"

kokx:
This reverts commit 3211a3fa3d55a563bae6cf882a60a4f74e18c2b4.

Super important fix. Pls merge ASAP.

### DIFF
--- a/module/Application/view/partial/main-nav.phtml
+++ b/module/Application/view/partial/main-nav.phtml
@@ -1,7 +1,57 @@
 <?php
 $lang = $this->plugin('translate')->getTranslator()->getLocale();
 ?>
-<nav class="navbar navbar-gewis navbar-static-top" role="navigation">
+<?php if ($this->identity()): ?>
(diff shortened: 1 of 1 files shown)


Kevinjil (approved):

tomudding (approved):

Koen1999 (changes requested):
I seriously doubt whether we want to keep this in our codebase. They have enjoyed it for long enough.

LeonVreling:
I think it would be nice to keep, but that might just be a biased opinion

kokx:
Might be nice for the current board as well.

Koen1999:
If we want to keep this, sure. But then we should refactor the code a bit. The way I introduced it in the past was not meant to be maintained.

tomudding:
I am closing this pending another PR which will implement this in a better way (using the new `isBoardMember` functions).

Issue #1083, opened 2021-04-23
Fix GetJobs hotfix/speed


Koen1999:
Please see: https://github.com/GEWIS/gewisweb/blob/2232e55ae7ea3fc2b35b84833c58d631115ccfbd/module/Company/src/Company/Service/Company.php#L1014-L1045

Probably adjusting the mapper would be the way to go.

PR #1085, opened 2021-04-23
Fix company overview loading times by removing hotfixes

tomudding:
This is just here to document the changes I made, if we ever need them in the future.


tomudding:
#1083 for reference

PR #1089, opened 2021-04-29
Prevent proxied traffic from getting incorrect user roles

tomudding:
A possibility existed where proxied traffic was incorrectly assigned
the `tueguest` user role, while the original traffic did not originate
from within the specified range.

### DIFF
--- a/config/autoload/local.php.dist
+++ b/config/autoload/local.php.dist
@@ -13,6 +13,15 @@
 
 return [
     /**
+     * Proxy configuration, used when determinining user roles.
+     */
+    'proxy' => [
+        'enabled' => false,
+        'ip_addresses' => [],
+        'header' => 'X-Real-IP',
+    ],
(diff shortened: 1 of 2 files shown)


config/autoload/local.php.dist
```diff
@@ -12,6 +12,15 @@
  */
 
 return [
+    /**
+     * Proxy configuration, used when determinining user roles.
+     */
+    'proxy' => [
+        'enabled' => false,
+        'ip_addresses' => [],
+        'header' => 'X-Forwarded-For',
```
kokx, line 21: Looks good, as long as you change this to `X-Real-Ip`

kokx (approved):

config/autoload/local.php.dist
tomudding: Changed it!

PR #1095, opened 2021-06-04
Fixed some small bugs

Koen1999:
Running code analysis tools is fun. We should do it more often.

### DIFF
--- a/module/Photo/src/Photo/Service/Admin.php
+++ b/module/Photo/src/Photo/Service/Admin.php
@@ -2,7 +2,9 @@
 
 namespace Photo\Service;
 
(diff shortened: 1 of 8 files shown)


Koen1999:
Some where language level issues, so I pushed them to the Docker PR in which the language level was upgraded

module/Photo/src/Photo/Service/Admin.php
```diff
@@ -2,7 +2,9 @@
 
 namespace Photo\Service;
 
+use _HumbugBoxa991b62ce91e\Nette\FileNotFoundException;
 use Application\Service\AbstractAclService;
+use http\Exception\InvalidArgumentException;
```
kokx, line 7: These look really sketchy

PR #1103, opened 2021-06-28
Removed unused modules from composer

Koen1999:
As far as I can tell these were once used for a pdf export functionality for activity data.
Especially the DOMPDF module is no longer actively maintained and blocking upgrade to ZF3.


tomudding (changes requested):
https://github.com/GEWIS/gewisweb/blob/b38227fee6bb8c92e7f0b8c0f01a6fd87b834b26/module/Activity/src/Activity/Controller/AdminController.php#L8

It is actively used in the Activity module to export lists of participants.

tomudding (approved):

PR #1110, opened 2021-07-04
Log cronjob output for diagnostic purposes

Koen1999:


tomudding (approved):
Not a fix in my opinion, but a way to get better diagnostic information.

Regarding diagnostics, it is laughable that php-fpm does not have (r)syslog for proper logging.

PR #1112, opened 2021-07-05
Fixed several bugs

Koen1999:


Koen1999:
Cron can now find the php executable and the max http body size has been increased to allow for uploads of photos

tomudding (approved):

PR #1117, opened 2021-07-09
Made use of Infima API endpoint

Koen1999:
The new implementation uses AJAX to retrieve an infimum for each client. We currently still only do this for authenthicated users.

There is one remaining issue which needs to be fixed by either the CBC or the Supremum committee.
We cannot use the endpoint like this because the CORS header is missing from the response.
afbeelding
https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin


Koen1999:
I tested it with the changes the CBC made to the header and it works now!

tomudding (approved):

Erikct98:
_On behalf of Supremum_

Earlier this week I implemented the Infimum API endpoint on the supremum.gewis.nl website. I'd like to raise four points of concern regarding this:
1. We propose to change the url to `supremum.gewis.nl/api/random_infimum` with the intent to making the name more self-documenting and easier to understand for future maintainers.
2. To prevent non-members from accessing this endpoint, we currently require a valid API key to be sent along with the request as a parameter (i.e. `supremum.gewis.nl/api/random_infimum?key=<insert key here>`). Once we provide you this key, would you be able to work with this construction, would you rather use headers, or perhaps even something else? We are open for input.
3. We cache the random infimum for 5 minutes, the idea being that we don't unnecessarily burden our server (as this endpoint is probably the most heavily used part of our website) and to keep infima new for a longer time. Do you foresee any difficulties with this caching from your side?

Another remark:
1. We provide a 503 along with explanatory message packaged in JSON if for whatever reason we are unable to retrieve an infimum.

Looking forward to your response.

PR #1119, opened 2021-07-12
Upgraded the project to Laminas Framework

Koen1999:
The Frontpage is loading already but we should verify a couple of things and preferably refactor a bit to make the project a bit more maintainable in the future.
While refactoring to upgrade to ZF3 I encountered quite some architectural issues.

With this update we get: 
- A more up-to-date version of our framework. We skip ZF3 and go directly from ZF2 to Laminas Framework.
- A more up-t-date version of PHP. We now have PHP version 7.4 with high chances of being able to upgrade to PHP 8.0 in the future
- Less outdated dependency warnings
- Less abandoned packages warnings
- PSR-4 file structure in modules
- Composer autoloading instead of zend-loader or laminas-loader
- A renewed ACL interface with seperate services for each module. Less code duplication and better seperation of services.
- A more secure website probably
- A faster website?
- Did I say less warnings?

TODO before merging:
- Re-enable the ApiUser role in the `user-role` service after resolving circular dependencies
- See if admin role permissions work correctly
- Review TODO marks I left within the code and resolve where neccessary (note that some relate to pre-existing problems)
- Review effects on ORM patches
- Do manual testing to see if everything works
- Resolve merge conflicts

TODO after merging (will probably create seperate issues for these later):
- Create dedicated services within each module that expose relevant functionality to other modules (facade design pattern). This helps prevent circular dependencies between modules
- God damn, improve that code quality
- Move ACL checks from services to the controller?
- More refactoring?
- Apply updates to individual components (check `php composer.phar outdated`)
- Currently glide-zend still has a dependency on zend-diactoros. This is not blocking functionality but it is blocking upgrades in the future. We should probably drop this dependency at some point. (https://github.com/thephpleague/glide-zend/issues/2)
- Add return types to methods and parameter types as well. These are type-checked contrary to the PhpDoc comments and will thus improve the reliability of our code and debugging experience.


Koen1999:
> 
> 
> It is purely my own preference, but I wouldn't it be better if the array values (keys) get an additional indent?
> 
> Instead of
> 
> ```
> [
> 'key' => 'value',
> ]
> ```
> 
> You get (similar to what we had):
> 
> ```
> [
>     'key' => 'value',
> ]
> ```

I completely agree. I think this is already fixed now. The Laminas migration tool messed some things up, but functionally it worked very well.

module/Activity/src/Activity/Service/ActivityQuery.php
```diff
@@ -121,30 +121,31 @@ public function getAllProposals()
...
      *
      * @param ActivityModel $activity
+     *
      * @return array
      */
     public function getAvailableLanguages($activity)
     {
         return ['nl' => !is_null($activity->getName()->getValueNL()),
-            'en' => !is_null($activity->getName()->getValueEN())];
+            'en' => !is_null($activity->getName()->getValueEN()), ];
```
tomudding: There are now weird cases where commas are added while it does not fit the actual structure of the array. This should be fixed.

module/Company/src/Company/Form/CollectionBaseFieldsetAwareForm.php
```diff
@@ -14,19 +14,11 @@ class CollectionBaseFieldsetAwareForm extends Form
...
-                    '%s expects the $flags argument to be one of "%s" or "%s"; received "%s"',
-                    __METHOD__,
-                    'Laminas\Form\FormInterface::VALUES_NORMALIZED',
-                    'Laminas\Form\FormInterface::VALUES_RAW',
-                    $flags
-                )
-            );
+        if (!in_array($flags, [FormInterface::VALUES_NORMALIZED, FormInterface::VALUES_RAW])) {
+            throw new InvalidArgumentException(sprintf('%s expects the $flags argument to be one of "%s" or "%s"; received "%s"', __METHOD__, 'Laminas\Form\FormInterface::VALUES_NORMALIZED', 'Laminas\Form\FormInterface::VALUES_RAW', $flags));
         }
```
tomudding, line 19: In certain cases these longer methods (most often exceptions) are compressed into a single line. This is something we should try to prevent.

tomudding (commented):
I feel like d20ccf30642cfba873b50fc2a1059b246a9356a0 (specifically the Symfony Coding Standard) actually decreases the readability of the code we have.

This specifically relates to the concatenation operator (`.`) and the removal of the spaces around it. PSR-12 ([6.2 Binary Operations](https://www.php-fig.org/psr/psr-12/#62-binary-operators)) specifically states there _must_ be a space. Symfony deviates from this and to prevent constantly changing things we should adhere to one standard (perhaps we should vote on this during the next meeting).

An additional note, but this is purely my own opinion. I do not like the Yoda style comparisons Symfony dictates to be used (however, this is something I can probably get used to).

module/Activity/src/Activity/Service/ActivityQuery.php
Koen1999: For multiline arrays I think it's better to have them.

module/Company/src/Company/Form/CollectionBaseFieldsetAwareForm.php
Koen1999: I agree. I'll see what I can do.

module/Activity/src/Activity/Service/ActivityQuery.php
tomudding: Yeah, but then we should change the layout of the array (in this specific case that would mean putting the `];` on a new line)

tomudding:
> This specifically relates to the concatenation operator (`.`) and the removal of the spaces around it. PSR-12 ([6.2 Binary Operations](https://www.php-fig.org/psr/psr-12/#62-binary-operators)) specifically states there _must_ be a space. Symfony deviates from this

It appears this was specifically force pushed away at some point, so this is no longer an issue.

Koen1999:
I looked at the PHP-CS-Fixer in the Makefile again and see that I configured it poorly. So Symfony wasn't even applied as a coding style. I removed it and added DoctrineAnnotations Coding style

Koen1999:
Also, it's a shame that codeclimate gave up on this big PR. I thought I had to approve to trigger the analysis, but that was a mistake XD

Koen1999:
I think most errors have been removed from the PR now :D

module/Activity/src/Model/Activity.php
```diff
@@ -469,7 +470,7 @@ public function setRequireGEFLITST($requireGEFLITST)
     }
 
     /**
-     * @return ArrayCollection
```
tomudding, line 475: These are `ArrayCollection`s and `Collection` is an interface (which is implemented by `ArrayCollection`), I don't think we should change them (applies to most, if not all, changes in this commit).

Koen1999: If we don't use any other functionality that is not present in ArrayCollection we should use Collection. Interfaces change less often and it thus makes our code more maintainable :)

tomudding (approved):
Approved to start resolving issues with the upgrade to Laminas Framework.

Issue #1122, opened 2021-07-16
Laminas: Re-enable the ApiUser AuthorizationService


Koen1999:
There used to be a circular dependency. This needs to be resolved (if still present)

PR #1129, opened 2021-07-16
Re-enabled the ApiAuth service in a new form

Koen1999:
Also swapped get_class comparisons for instanceof comparisons

There's no circular dependency now :)


Koen1999:
This fixes #1122

tomudding (approved):

PR #1130, opened 2021-07-16
Fix issue where privileged `UserRole`s were not correctly applied

### LINKED ISSUE #1123: Laminas: Verify that admin role permissions work

tomudding:
Laminas processes inherited roles LIFO (Last-In-First-Out), this was not the case for Zend. Hence people with custom `UserRole`s were incorrectly stripped from their privileges (because people with custom `UserRole`s are almost always a member of an organ).

Furthermore, board members no longer had all privileges, with similar reasoning as above. As the checks stopped before reaching the `IsBoardMember` assertion as part of the `user` role.

To accomodate for this, a new `board` role has been added which is applied to `User`s who are currently a board member.

This fixes #1123.


Koen1999 (changes requested):
Please see #1131 as a proposed alternative solution.

tomudding:
Superseded by #1131.

PR #1131, opened 2021-07-17
Fixed issues with the ACL where the wrong user role was used.

### LINKED ISSUE #1123: Laminas: Verify that admin role permissions work

Koen1999:
Supersedes #1130 and fixes #1123.

I propose this solution as a more elegant solution instead of #1130 .

Essentially, we were already abusing the ACL by adding a role for every user.
Now we do it in a way that was intended by the authors of Laminas.


module/User/src/Model/User.php
```diff
@@ -152,13 +152,29 @@ public function getRoleNames()
...
+        // TODO: We could create an assertion for this. (member check for C4)
+        if (in_array('company_admin', $this->getRoleNames())) {
+            return 'company_admin';
+        }
+
+        if (in_array('photo_guest', $this->getRoleNames())) {
+            return 'photo_guest';
+        }
+
+        // TODO: We could create an assertion for this.
```
tomudding, line 172: Same as the first comment, I think that making this an assertion creates unnecessary overhead.

module/User/src/Model/User.php
```diff
@@ -152,13 +152,29 @@ public function getRoleNames()
...
      */
-    public function getRoleId()
+    public function getRoleId(): string
     {
-        return 'user_' . $this->getLidnr();
+        if (in_array('admin', $this->getRoleNames())) {
+            return 'admin';
+        }
+
+        // TODO: We could create an assertion for this. (member check for C4)
```
tomudding, line 163: I do not think we should make this an assertion. The assertion is an additional check for every `isAllowed()` that is performed in an execution, while the `Member` object is only created once during an execution. Hence making it an assertion would create overhead.

module/User/src/Model/User.php
```diff
@@ -152,13 +152,29 @@ public function getRoleNames()
...
+        if (in_array('company_admin', $this->getRoleNames())) {
+            return 'company_admin';
+        }
+
+        if (in_array('photo_guest', $this->getRoleNames())) {
+            return 'photo_guest';
+        }
```
tomudding, line 170: Perhaps this should actually be the first return, to ensure that `photo_guest` can only have those permissions.

module/User/src/Model/User.php
```diff
@@ -152,13 +152,29 @@ public function getRoleNames()
...
     /**
-     * Get the user's role ID.
-     *
      * @return string
      */
-    public function getRoleId()
+    public function getRoleId(): string
     {
-        return 'user_' . $this->getLidnr();
+        if (in_array('admin', $this->getRoleNames())) {
```
tomudding, line 159: Perhaps it is worth it to not do `$this->getRoleNames()` for the role check (i.e. make it a variable and pass the variable to `in_array()`).

module/User/src/Authorization/GenericAclService.php
```diff
@@ -33,17 +34,27 @@ public function __construct(
...
+     *
+     * The role that should take precedence should be returned first.
+     * This is because of the behaviour of {@link \Laminas\Permissions\Acl\Role\Registry}.
      */
     protected function getRole()
     {
         if ($this->authService->hasIdentity()) {
-            return $this->authService->getIdentity();
+            $user = $this->authService->getIdentity();
+            if ($user instanceof User) {
```
tomudding, line 45: I understand the additional check (`instanceof`), however, if the `authService` has an identity it is guaranteed to be a `User`.

Also, I think it is a bit nicer to have an empty line between the two lines.

tomudding (commented):
While this is a more elegant solution (and probably a better fit for the ACL), I really liked the idea of assigning multiple roles to users, as you could easily see which roles applied to the user. This is of course not needed as we have proper inheritance of roles, so the changes proposed here seem appropriate.

> Essentially, we were already abusing the ACL by adding a role for every user.

I would not classify it as abusing the ACL, it was merely creating some extra overhead with roles that were never used.

tomudding (changes requested):
I did expect this to be an issue, as I encountered the same problem with my own PR. Board members do not have full permissions as they are (almost) always limited to their `active_member` role.

I propose we remove the assertion and add a new `board` role, which is applied in the same way as your proposed changes (after `admin` and before `active_member`).

module/User/src/Model/User.php
Koen1999: I disagree. Highest priority roles should be returned first.

Koen1999:
The new changes are also deployed on test.gewis.nl for testing.

module/User/src/Service/AclService.php
```diff
@@ -52,34 +52,16 @@ protected function createAcl()
...
-
-            $this->acl->addRole($user, $roles);
-        }
-
         // admins are allowed to do everything
         $this->acl->allow('admin');
 
         // board members also are admins
-        $this->acl->allow('user', null, null, new IsBoardMember());
+        $this->acl->allow('active_member', null, null, new IsBoardMember());
```
tomudding, line 64: This does not work with my comment regarding the `company_admin` check.

We already discussed the part below, but I like to mention it here for documentation purposes.

Furthermore, I disagree that we can make this assumption. Yes, in practice this will always be the case, but we cannot guarantee it. I would rather see a separate `board` role which is applied using the same logic but as part of the `User->getRoleId()`.

That way it is clearly that a board member has a clear and predefined hierarchy & role (that of a board member) and not an active member with extended capabilities.

module/User/src/Model/User.php
```diff
@@ -152,13 +152,34 @@ public function getRoleNames()
...
+            return 'admin';
+        }
+
+        if (in_array('company_admin', $roleNames)) {
+            return 'company_admin';
+        }
+
+        if (empty($roleNames)) {
+            return 'user';
+        }
```
tomudding, line 170: A user who is an active member (and does not have an `admin` or `company_admin` role) does not get the `active_member` role, as this will evaluate to true and return. I suggest moving it below the active member check.

module/User/src/Model/User.php
```diff
@@ -152,13 +152,34 @@ public function getRoleNames()
...
     {
-        return 'user_' . $this->getLidnr();
+        $roleNames = $this->getRoleNames();
+        if (in_array('admin', $roleNames)) {
+            return 'admin';
+        }
+
+        if (in_array('company_admin', $roleNames)) {
+            return 'company_admin';
+        }
```
tomudding, line 166: If a board member has the `company_admin` role (i.e. they are part of C4) they do not get the correct board privileges.

tomudding (changes requested):

module/User/src/Model/User.php
Koen1999: Not sure why it's in this spot indeed. I'll move it.

module/User/src/Service/AclService.php
Koen1999: I'll come up with a way that applies this assertion to all roles instead.

module/User/src/Model/User.php
Koen1999: They simply shouldn't be I'll come up with a way that applies this assertion to all roles instead.

module/User/src/Service/AclService.php
```diff
@@ -52,34 +52,19 @@ protected function createAcl()
...
         // admins are allowed to do everything
         $this->acl->allow('admin');
 
-        // board members also are admins
-        $this->acl->allow('user', null, null, new IsBoardMember());
+        foreach ($this->acl->getRoles() as $role) {
+            // board members also are admins
+            // since they can have any other role, we need to apply this assertion to all roles
+            $this->acl->allow($role, null, null, new IsBoardMember());
+        }
```
tomudding, line 67: This actually breaks the `admin` role (nothing is allowed). Furthermore, it still does not give board members the appropriate privileges.

tomudding (changes requested):
Changes requested because privileges are still not correctly applied.

---
General comment:

I don't agree, as I believe an assertion should be used for rules that are not absolute (such as ownership, like our other assertions) and not to specify a complete separate role (which is pretty absolute: grant this user all privileges).

I'd recommend one of our senior members takes a look at this and decides what we should do.

---

I would also like to suggest a new PR when we have a final solution to get rid of these unnecessary commits.

module/Decision/src/Model/Member.php
```diff
@@ -730,4 +730,33 @@ public function clearLists()
...
+     * Check if this is a current board member.
+     *
+     * @return bool
+     */
+    protected function isCurrentBoard(BoardMember $boardMember): bool
+    {
+        $now = new DateTime();
+
+        return $boardMember->getInstallDate() <= $now
+            && (null === $boardMember->getDischargeDate() || $boardMember->getDischargeDate() >= $now);
```
jszanto, line 760: P.S. this is an invalid assumption. The board is often discharged 3-6 months after they stop being board. It is probably desirable to remove website admin rights earlier.

jszanto:
Instead of abusing the ACL to build RBAC you might want to look into using the RBAC package of Laminas.

tomudding:
> 
> 
> Instead of abusing the ACL to build RBAC you might want to look into using the RBAC package of Laminas.

I suppose we could do this, however, the one things that makes ACL better in my opinion is how resources are handled. With the ACL we can define resources more easily and than you can just pass the whole resource object (in this example an activity) to handle checking if someone is allowed to do `edit` (permission).

```php
// $user->getRoleId() is active_member
$acl = new Acl();
$acl->addRole(new Role('active_member')); // imagine some role inheritance

// (...)

$acl->allow('active_member', 'activity', 'edit', new IsCreatorOrOganMember()); //*

// (...)

if ($acl->isAllowed($user, $activity, 'edit')) {
    // $user can edit $activity
}
```

Whereas with RBAC you will have to pass the assertion every time you are checking if the `activity.edit` (permission) is granted:

```php
// $user->getRole() is active_member
$rbac = new Rbac();
$rbac->addRole(new Role('active_member');  // imagine some role inheritance

(...)

$rbac->getRole('active_member')->addPermission('activity.edit'); // *

// (...)

$assertion = new IsCreatorOrOganMember($user, $activity);
if ($rbac->isGranted($user->getRole(), 'activity.edit', $assertion)) {
    // $user can edit $activity
}
```

`*`: With ACL you can define many permissions for a certain resource and role (`'activity', ['create', 'edit', '...']`). With RBAC you will have to define everything verbose (e.g., `activity.create`, `activity.edit`, `activity....`).

module/Decision/src/Model/Member.php
tomudding: This is the initial definition of the `IsBoardMember` assertion ([is still the same since its creation in 2015](https://github.com/kokx/gewisweb/commit/4d688bb96e04bf9e1d4f7f3c30cf3fa56ebc5d05#diff-084dcf54fe33d2065303f60e660218e039d61b3bef794b1a3413b91290356d18)).

I think we have had this discussion in the past and came to the conclusion that it is useful for the old board to still have the admin rights, but maybe I am not remembering this correctly.

module/User/src/Model/User.php
```diff
@@ -157,7 +158,7 @@ public function getRoleNames()
     public function getRoleId(): string
     {
         $roleNames = $this->getRoleNames();
-        if (in_array('admin', $roleNames)) {
+        if (in_array('admin', $roleNames) || $this->getMember()->isBoardMember()) {
             return 'admin';
         }
```
tomudding: This is essentially what I suggested in #1147, just without the new `board` role (which I personally really like) and the check in `Member` instead of `User`.

I suggest we await some more feedback, but otherwise this seems like a good solution.

module/User/src/Permissions/Assertion/IsBoardMember.php
```diff
@@ -36,25 +34,6 @@ public function assert(Acl $acl, RoleInterface $role = null, ResourceInterface $
...
-     */
-    protected function isCurrentBoard(BoardMember $boardMember)
-    {
-        $now = new DateTime();
-
-        return $boardMember->getInstallDate() <= $now &&
-            (null === $boardMember->getDischargeDate() || $boardMember->getDischargeDate() >= $now);
+        return $member->isBoardMember();
     }
 }
```
tomudding, line 39: Something I forgot to mention, this assertion is not used anywhere else in the website and can finally be removed.

Koen1999: I removed the assertion

module/User/src/Model/User.php
Koen1999: Since board and admin permissions are not different, I propose not having seperate roles.
The installations required for checking wheteher someone is a board member are all located in the decision module. So, the check should be there too.

Koen1999:
> 
> 
> Instead of abusing the ACL to build RBAC you might want to look into using the RBAC package of Laminas.

This is an interesting suggestion, but I also agree with Tom. Moreover, the effort to change this would be higher than we like and wouldn't make the overall application a lot better.

module/Decision/src/Model/Member.php
Koen1999: I think we discussed this within the committee at some point. We didn't change it following that. So I don't see why we would change it now.

tomudding (approved):
It works.

I propose more rigorous testing of implemented changes in the future, to ensure we do not have to keep debugging PRs. A lot of time was wasted on trying to get the `IsBoardMember` assertion to work, whereas #1130 already stated that the assertion did no longer work.

#1147 is a squashed version of this PR, but closed in favour of this PR (which has more comments as documentation). I still recommend a squash merge on this PR to keep the history somewhat clean.

PR #1142, opened 2021-07-21
Fixes #1137: Organ membership where user has no function results in `Undefined index`

### LINKED ISSUE #1137: Organ membership where user has no function results in `Undefined index`
https://github.com/GEWIS/gewisweb/blob/26825507320b2ec64687abb9068516db3e76faa7/module/Decision/view/decision/member/view.phtml#L62

Tries to display a function of an organ membership for a specific user, however, this key is not always defined. As a result, when a member is just a `Lid` of an organ this line will result in an `Undefined index` notice.

tomudding:
Fixes #1137.


Koen1999 (approved):
Noticed aren't big deals, but I agree it's always nice to perform such checks.

PR #1144, opened 2021-07-22
Fix being unable to add new courses

### LINKED ISSUE #1136: Unable to add new courses
Zend's `string_to_upper` does not exist in Laminas:

https://github.com/GEWIS/gewisweb/blob/26825507320b2ec64687abb9068516db3e76faa7/module/Education/src/Form/AddCourse.php#L107

And should be replaced with `StringToUpper`.

tomudding:
Fixes #1136.

### DIFF
--- a/module/Education/src/Form/AddCourse.php
+++ b/module/Education/src/Form/AddCourse.php
@@ -2,6 +2,7 @@
 
 namespace Education\Form;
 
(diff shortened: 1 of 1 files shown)


Koen1999 (approved):
Appears I forgot some things, but it's kind of impossible to spot everything in one go XD

PR #1146, opened 2021-07-22
Added a PhpStan shortcut to evaluate PR changes & moved PhpStan files

Koen1999:
This makes it easier to review PRs and the main directory is a bit cleaner


tomudding:
Quick question, [phpstan/phpstan-baseline.neon](https://github.com/GEWIS/gewisweb/blob/cdce9ac026af4c644584131fc2341f3b2341bc9d/phpstan/phpstan-baseline.neon) has ignored errors. Is this intentional? Because I would think we should fix those "errors".

Koen1999:
> 
> 
> Quick question, [phpstan/phpstan-baseline.neon](https://github.com/GEWIS/gewisweb/blob/cdce9ac026af4c644584131fc2341f3b2341bc9d/phpstan/phpstan-baseline.neon) has ignored errors. Is this intentional? Because I would think we should fix those "errors".

Yep, this is intentional. Laminas returns an interface that does not very accurately describe the object being returned. I actually created an issue for this hoping that it will be improved in a later version. Perhaps with a new interface that has more of the methods that we use. (https://github.com/laminas/laminas-mvc/issues/77)
The errors from PhpStan which we ignore aren't actually errors. It could be nice to add checks, but this would be quite some work that doesn't add functionality.

Koen1999:
@tomudding can we merge this so we can start using it? :)

tomudding (approved):
Looks to be working. But we may have to be careful with the `git checkout --detach master`, as not every branch is guaranteed to be based on the `master` branch.

PR #1147, opened 2021-07-23
Fix role order to ensure correct privileges for each user

### LINKED ISSUE #1123: Laminas: Verify that admin role permissions work

tomudding:
I suggest this as a clean working version of #1131 (which in turn superseded #1130). It can be tested via the test website, where a debug statement has been added to show which role has been applied.

---

The Laminas ACL changed the order in which inherited roles are applied. Previously, all roles would be checked, however, this is no longer the case. Now roles are checked LIFO (last-in-first-out), which means that the most privileged role should be applied last.

This was not the case, which meant that users who should get most privileges (`company_admin`s, `admin`s, and board members) did not get the correct privileges.

To prevent this from happening, we devised a cleaner solution which directly applies the role of a User in the correct order. The order of most privileged to least priviliged is defined as follows:
- `admin` / `board`
- `company_admin`
- `active_member`
- `user`
- `photo_guest`

This means we have now also introduced a new role for board members (`board`), this is because the previously existing assertion does not correctly function when only applied to the `user` role (as board members are almost always `active_member` or `company_admin`). Applying the assertion to all roles caused other issues with the all privileged `admin` role. Since the role of a board member is absolute, an assertion (which should be used to determine rules that are not absolute) is ill-suited for determining the privileges for a board member.

This fixes #1123.

Co-Authored-By: @Koen1999


Koen1999:
@tomudding 
I would prefer my solution given that it makes better use of how the ACL is supposed to work. If you insist on removing the assertion check in the ACL, I propose changing the User model from my PR to return 'admin' as role whenever the assertion passes. What do you think of this approach?

Koen1999:
Please have a look at what I did in my last commit on #1131

tomudding:
> 
> 
> @tomudding
> I would prefer my solution given that it makes better use of how the ACL is supposed to work. If you insist on removing the assertion check in the ACL, I propose changing the User model from my PR to return 'admin' as role whenever the assertion passes. What do you think of this approach?

I already commented on your PR, but that is essentially what I did in this PR (with the addition of the `board` role).

Not that it really matters, but your PR blurs the history with -what were untested- commits, whereas this PR changed everything in one go with **proper documentation** (the most important part imho) on why we chose to do it this way.

tomudding:
Closed in favour of #1131.

PR #1148, opened 2021-07-23
Native attributes for model annotation

### LINKED ISSUE #1149: Remove Carbon as dependecy
Carbon is severely outdated and used in only one place -the `Job` model in the `Company` module. A `DateTime` representation suffices in that instance and therefore we can remove `nesbot/carbon` (`Carbon/Carbon`) as a dependency.

### LINKED ISSUE #1186: Subdecision ordering not working due to not renamed function
eb4de6dc176fb439598745bce7496e081d29f302 renamed `getSubDecisionsUploadOrdered` to `getOrderedSubdecisions`, however, this was not reflected in the view it is used in:

https://github.com/GEWIS/gewisweb/blob/4c626dacd25116e536cb89a9924f575658d2b019/module/Decision/view/decision/organ/show.phtml#L33

### LINKED ISSUE #1188: tooltip.js missing for activity creation/updating form

### LINKED ISSUE #1200: Activity (dis)approver not being saved
This is fixed in #1148.

### LINKED ISSUE #1227: FileUrl ViewHelper breaks PHPStan analysis
The `FileUrl` ViewHelper utilises `ServiceLocatorInterface`, however, all other ViewHelpers which require access to the `ServiceLocator` use `Interop\Container\ContainerInterface`.

When running `phpstan analyse`, it hangs on this file. Changing the container interface does not change the behaviour, but allows PHPStan to correctly analyse the file.

tomudding:
This changes the Doctrine ORM annotations on our entities. Instead of the DocBlock annotations I propose we use the native attributes introduced in PHP 8.0. This means we no longer depend on something which is not an official standard (the PSR for PHPDoc is still `proposed`) and we have better documentation for our code (because attributes will show up in your IDE).

Unfortunately, a small bug exists in `doctrine/doctrine-module` (see https://github.com/doctrine/DoctrineModule/issues/728), hence we use a wrapper for the ORM `AttributeDriver`.

Apart from the annotation change, this PR also includes proper typing of all entity attributes and functions. Theoretically, with the proper typing of the attributes the `type` parameter can be omitted on the `#[Column]`s, `#[OneToOne]`s, and `#[ManyToOne]`s, however, as we also use non-standard types I have refrained from doing this.

Progress (module):
- [X] Activity
- [x] Company
- [x] Decision
- [x] Education
- [x] Frontpage
- [x] Photo
- [x] User

This resolves #1149, fixes #1186, fixes #1188, fixes #1200, and fixes #1227.


tomudding:
As mentioned in 1da3e18 this uses a temporary wrapper for the ORM `AttributeDriver` till an issue in DoctrineModule is resolved.

tomudding:
There a few more unintentional side-effects which need fixing before this pull request can be reviewed.

tomudding:
There are only two things left to do; fix creation and updating of everything in the `Company` module (i.e. the `bind`ing needs to go), and run PHPStan. The latter one will take some time as I have changed a lot of files.

Koen1999:
I adjusted the PhpStan setup to provide more useful comments on the project given the changes in your PR. Here is an excerpt:
```

 ------ -------------------------------------------------------------------------------
  Line   module/Activity/src/Service/Activity.php
 ------ -------------------------------------------------------------------------------
  454    Call to an undefined method Doctrine\Common\Collections\Collection::unwrap().
 ------ -------------------------------------------------------------------------------

 ------ ------------------------------------------------------------------------------------
  Line   module/Application/src/Extensions/Doctrine/AttributeDriver.php
 ------ ------------------------------------------------------------------------------------
  14     Constructor of class Application\Extensions\Doctrine\AttributeDriver has an unused
         parameter $reader.
 ------ ------------------------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------
  Line   module/Company/src/Mapper/Category.php
 ------ ---------------------------------------------------------------------------------------------
  40     Parameter #1 $languageNeutralId of method Company\Model\JobCategory::setLanguageNeutralId()
         expects int, null given.
 ------ ---------------------------------------------------------------------------------------------

 ------ -----------------------------------------------------------------------------------------
  Line   module/Company/src/Model/Company.php
 ------ -----------------------------------------------------------------------------------------
  355    PHPDoc tag @param has invalid value ($category): Unexpected token "$category", expected
         type at offset 178
 ------ -----------------------------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------
  Line   module/Company/src/Model/CompanyI18n.php
 ------ ---------------------------------------------------------------------------------------------
  87     PHPDoc tag @param has invalid value ($company): Unexpected token "$company", expected type
         at offset 67
  87     PHPDoc tag @param has invalid value ($locale): Unexpected token "$locale", expected type at
         offset 45
 ------ ---------------------------------------------------------------------------------------------

 ------ -----------------------------------------------------------------------------------------
  Line   module/Company/src/Model/CompanyJobPackage.php
 ------ -----------------------------------------------------------------------------------------
  56     PHPDoc tag @param has invalid value ($category): Unexpected token "$category", expected
         type at offset 71
  68     PHPDoc tag @param has invalid value ($category): Unexpected token "$category", expected
         type at offset 82
 ------ -----------------------------------------------------------------------------------------

 ------ ---------------------------------------------------------------------------------------------
  Line   module/Company/src/Model/CompanyPackage.php
 ------ ---------------------------------------------------------------------------------------------
  146    PHPDoc tag @param has invalid value ($category): Unexpected token "$category", expected
         type at offset 123
  188    Method Company\Model\CompanyPackage::getType() should return string but return statement is
         missing.
 ------ ---------------------------------------------------------------------------------------------

 ------ ------------------------------------------------------------------------------------------
  Line   module/Company/src/Model/Job.php
 ------ ------------------------------------------------------------------------------------------
  482    Strict comparison using === between null and Doctrine\Common\Collections\Collection will
         always evaluate to false.
 ------ ------------------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------------------------
  Line   module/Decision/src/Model/AssociationYear.php
 ------ ----------------------------------------------------------------------------------
  40     Unsafe usage of new static().
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static
  55     Unsafe usage of new static().
         💡 See: https://phpstan.org/blog/solving-phpstan-error-unsafe-usage-of-new-static
 ------ ----------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------------
  Line   module/Decision/src/Model/Member.php
 ------ ----------------------------------------------------------------------
  332    Property Decision\Model\Member::$lidnr (int) does not accept string.
 ------ ----------------------------------------------------------------------

 ------ --------------------------------------------------------------------------------------------
  Line   module/Decision/src/Service/Decision.php
 ------ --------------------------------------------------------------------------------------------
  258    Call to function is_null() with Decision\Model\MeetingNotes will always evaluate to false.
  294    Call to function is_null() with Decision\Model\MeetingNotes will always evaluate to false.
 ------ --------------------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------
  Line   module/Education/src/Service/Exam.php
 ------ ----------------------------------------------------------------
  230    Call to an undefined method Education\Model\Exam::setAuthor().
 ------ ----------------------------------------------------------------

 ------ -----------------------------------------------------------
  Line   module/Photo/src/Model/Album.php
 ------ -----------------------------------------------------------
  123    Method Photo\Model\Album::getChildren() should return
         Doctrine\Common\Collections\Collection but returns array.
 ------ -----------------------------------------------------------

 ------ --------------------------------------------------------------------------------------
  Line   module/Photo/src/Model/Photo.php
 ------ --------------------------------------------------------------------------------------
  472    Property Photo\Model\Photo::$focalLength (float|null) does not accept string|null.
  482    Property Photo\Model\Photo::$exposureTime (float|null) does not accept string|null.
  601    PHPDoc tag @param has invalid value (Vote @vote): Unexpected token "@vote", expected
         variable at offset 64
 ------ --------------------------------------------------------------------------------------

 ------ -----------------------------------------------------------------------------------
  Line   module/Photo/src/Service/Metadata.php
 ------ -----------------------------------------------------------------------------------
  30     Parameter #1 $focalLength of method Photo\Model\Photo::setFocalLength() expects
         string|null, float given.
  31     Parameter #1 $exposureTime of method Photo\Model\Photo::setExposureTime() expects
         string|null, float given.
 ------ -----------------------------------------------------------------------------------

 ------ ------------------------------------------------------------------------------------
  Line   module/Photo/src/Service/Photo.php
 ------ ------------------------------------------------------------------------------------
  652    Parameter #2 $voter of class Photo\Model\Vote constructor expects User\Model\User,
         Decision\Model\Member given.
 ------ ------------------------------------------------------------------------------------


 [ERROR] Found 24 errors

```

Koen1999:
I have one question for when we need to deploy this. Are database changes required or is it (almost) only changes on the side of our application?

tomudding:
> 
> 
> I have one question for when we need to deploy this. Are database changes required or is it (almost) only changes on the side of our application?

**Significant** changes to the database (as this includes the changes to the Company module from [gewisweb/feature/company-module-localisedtext](https://github.com/GEWIS/gewisweb/tree/feature/company-module-localisedtext)). It will be very similar to the activity overhaul update, so we have enough experience, but we should have a 2 hour window during an evening/weekend to carefully deploy these changes.

---

I have rebased this branch onto master, so it is easier to check the changes. I am currently running PHPStan to fix some of the remaining issues. Unfortunately, it appears to be stuck somewhere, so I do not know when it will finish.

Koen1999:
> 
> 
> Did you encounter any memory issues when running PHPStan? Whenever I run it (both the normal and the pr version) it never uses more than 100MB.

Yeah, I've had this multiple times in the past as well. Sometimes it gives an out of memory error, but it's not very predictable. The speed of PhpStan increased greatly after changing the memory limit anyway.

Koen1999:
> 
> 
> This is a non-issue, which shows that even PHPStan makes mistakes. In this case we can keep it this way, as we otherwise keep changing the code, but see below for why this was not really necessary (performance-wise there is less than a sub-microsecond difference between the two).
> 
> PHPStan reports:
> 
> > ------ ----------------------------------------------------------------
> > Line     module/Education/src/Service/Exam.php
> > ------ ----------------------------------------------------------------
> > 241      Call to an undefined method Education\Model\Exam::setAuthor().
> > ------ ----------------------------------------------------------------
> 
> Which is not possible. `get_class()` strictly returns just the class of the object, not the class of the parent(s). When within the if-statement, the object (`$exam`) is therefore guaranteed to be `SummaryModel` (which is of course `Education\Model\Summary`). At that point, the method `setAuthor()` is defined at all times.

I agree it is not really an issue, but it supresses the warning raised by PhpStan. Moreover, any class extending SummaryModel will also implement getAuthor. So the new way of doing things is more general.

Koen1999 (approved):
I see no blocking issues that suggest we should not merge.

PR #1159, opened 2021-07-30
Fixed several PhpStan issues

Koen1999:
Mostly just type hints in PhpDoc.

### DIFF
--- a/module/Application/src/Module.php
+++ b/module/Application/src/Module.php
@@ -174,12 +174,6 @@ class Module
 
                     return $helper;
                 },
-                'infima' => function (ContainerInterface $container) {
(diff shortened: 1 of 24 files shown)


tomudding:
I'll review this when I have more time. I also want to reiterate that most, if not all, issues related to the files in the `*/Model/` directories are already fixed in #1148 (which will only require the removal of the `bind` statements to become functional), you don't have to change it but it would be very useful if we do not have a lot of merge conflicts for either of the PRs.

Koen1999:
In case of conflicts it would then probably suffice to just choose the changes of the other PR.

tomudding (approved):

PR #1167, opened 2021-08-04
Fixed access to getLocale on Translator 

Koen1999:


tomudding (approved):
It still looks weird doing it like this, but now we use the same `Translator` everywhere, which is great.

PR #1168, opened 2021-08-04
Make Google Calendar events one day shorter in option calendar

### LINKED ISSUE #1116: Duration day-long activities option calendar
### Current situation
At the moment, the option calender supports day-long activities, coming in from the Google Calendar. This is a nice option in my opinion, so I would like to keep this

### Problems
Currently, these activities are displayed for two days. I think this is the case since the website calendar thinks that the activity ends at 00:00, which is technically the next day.

### Proposed solution
When parsing the data change the times of the events when the endtime is not set to 23:59. I think this is done somewhere [here](https://github.com/GEWIS/gewisweb/blob/master/module/Activity/view/activity/activity-calendar/index.phtml#L202) after some looking through the code

Hope you can do something with it !

LeonVreling:
This fixes #1116.

### DIFF
--- a/module/Activity/view/activity/activity-calendar/index.phtml
+++ b/module/Activity/view/activity/activity-calendar/index.phtml
@@ -209,11 +209,11 @@ $this->headLink()
     });
 
 
-    function getEndDate(end) {
-        if (end.dateTime !== undefined) {
-            return new Date(end.dateTime).getTime();
+    function getEndDate(item) {
+        if (item.dateTime !== undefined) {
+            return new Date(item.dateTime).getTime();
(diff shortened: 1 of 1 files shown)


LeonVreling:
Fixes [issue 1116](https://github.com/GEWIS/gewisweb/issues/1116)

tomudding:
This is not really a requirement, however, I think screenshots showing a before and after would be useful for these kind of issues.

LeonVreling:
Before: 
image

After: 
image

Koen1999 (approved):

Issue #1157, opened 2021-07-28
Unable to upload summary or exam


tomudding:
Same issue as with #1136: `string_to_upper` does no longer exist. Additionally the call to

https://github.com/GEWIS/gewisweb/blob/5338b66a4ca3b78e75831d54bfe2094dbb8df52a/module/Education/src/Service/Exam.php#L267

does not work as `$post` is a Laminas `Parameter` object and not an array.

PR #1171, opened 2021-08-08
Fix company language profiles not updating correctly

### LINKED ISSUE #958: Removing a language from a company does not remove language from database
When a company with a certain language has that language removed, the language remains in the database without any associated properties to a company. This causes "artefacts" that will remain in the database as it is no longer associated with a company upon deleting the company.

### LINKED ISSUE #1118: Potential issue when removing language from company profile
According to Anne's e-mail, it is possible to (accidentally) remove all languages from a company's profile. This results in an exception, ask a company needs to have at least one language.

This would mean that #958 is no longer valid, but we have to investigate this.

### LINKED ISSUE #1170: Unauthenticated user can add companies

tomudding:
This PR addresses multiple issues. First and foremost, a security vulnerability existed in the company module allowing unauthenticated users to add new companies.

Secondly, users were not able to add new companies and/or jobs due to incorrect typing (see #1157 for more information).

Lastly, this fixes the issue where when editing a company profile, the language profiles for that company would be destroyed. Doctrine cannot perform merge operations on its own, the last operation performed is what a UnitOfWork will execute. In this case it meant that saving a company's information would not only not save any new information regarding its I18n language profiles, the language profiles would be proactively deleted.

To fix this, edits to language profiles are now first persisted and after that has completed, any languages which are to be removed are removed.

This fixes #958, fixes #1118, and fixes #1170.

### DIFF
--- a/module/Company/src/Model/Company.php
+++ b/module/Company/src/Model/Company.php
@@ -88,8 +88,6 @@ class Company // implements ArrayHydrator (for zend2 form)
      */
     protected $packages;
 
-    private $languageNeutralId;
-
     /**
      * Constructor.
      */
@@ -112,15 +110,11 @@ class Company // implements ArrayHydrator (for zend2 form)
     /**
      * Get the company's translations.
      *
-     * @return Collection|array
+     * @return Collection
      */
     public function getTranslations()
     {
-        if (!is_null($this->translations)) {
-            return $this->translations;
-        }
-
-        return [];
+        return $this->translations;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


module/Company/src/Model/Company.php
```diff
@@ -399,47 +393,35 @@ function ($package) {
...
+            function ($value) {
+                return $value->getLanguage();
+            }
+        );
     }
 
     /**
      * If this object contains an translation for a given locale, it is returned, otherwise null is returned.
+     *
+     * @param $locale
```
Koen1999, line 412: You need to fix the @param tag by adding a type in front of the variable name

module/Company/src/Service/Company.php
```diff
@@ -712,21 +756,23 @@ public function createJob($packageId, $data, $files)
...
      * @param array $jobs The Job to save
-     * @param array $data The (new) data to save
-     * @param array $files The (new) files to save
+     * @param Parameters $data The (new) data to save
+     * @param Parameters $files The (new) files to save
      *
      * @return bool
      */
-    public function saveJobData($languageNeutralId, $jobs, $data, $files)
+    public function saveJobData($languageNeutralId, $jobs, Parameters $data, Parameters $files)
```
Koen1999, line 764: If you want to change these parameter types from array to Parameters, you should also change the type where the function is called. PhpStan reports such errors. With the actual type outside the PhpDoc, this would probably even throw errors.

module/Company/src/Service/Company.php
```diff
@@ -3,26 +3,34 @@
...
-use Company\Mapper\FeaturedPackage;
-use Company\Mapper\Label;
-use Company\Mapper\LabelAssignment;
-use Company\Mapper\Package;
-use Company\Model\Job;
-use Company\Model\Job as JobModel;
-use Company\Model\JobCategory as CategoryModel;
-use Company\Model\JobLabel as LabelModel;
-use Company\Model\JobLabelAssignment;
+use Company\Form\{
```
Koen1999, line 6: By changing these imports, most of the getXxxForm methods now have incorrect PhpDoc comments. Please fix these. They are also reported by PhpStan

module/Company/src/Service/Company.php
```diff
@@ -3,26 +3,34 @@
...
+use Company\Mapper\{
+    BannerPackage as BannerPackageMapper,
+    Category as CategoryMapper,
+    FeaturedPackage as FeaturedPackageMapper,
+    Job as JobMapper,
+    Label as LabelMapper,
+    LabelAssignment as LabelAssignmentMapper,
+    Package as PackageMapper,
+};
+use Company\Model\{
```
Koen1999, line 22: This also caused some invalid type hints in the PhpDoc. They are also reported by PhpStan

Koen1999 (changes requested):
Please run `make phpstan` and fix some of the type issues and PhpDoc issues reported in the files you changed.

module/Company/src/Service/Company.php
tomudding: Stupid PhpDocs :/ All these functions always used `Parameters`, but all DocBlocks mentioned `array`.

Koen1999 (approved):
Looks good now!

PR #1175, opened 2021-08-16
Updated dependencies

Koen1999:


tomudding (approved):
Looks good! I do not think this will break in the foreseeable future, but we may need to develop a strategy regarding updates to ORM and our patch.

Also, no updates to the NodeJS packages?

PR #1176, opened 2021-08-16
Created a BaseMapper to reduce code duplication and create a common interface

Koen1999:
Currently only the User module was changed because I would like some feedback before I do the rest.
Is this a good idea what I am doing and what could be improved?


module/Application/src/Mapper/BaseMapper.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @param object $entity
+     * @throws ORMException
+     */
+    public function persist(object $entity) {
+        $this->em->persist($entity);
+        $this->em->flush();
+    }
```
tomudding, line 34: I would add the return type `void` here, like:

```php
public function persist(object $entity): void {
    $this->em->persist($entity);
    $this->em->flush();
}
```

This also applies to `remove()`, `removeById()`, `flush()`, and `detach()`.

module/Application/src/Mapper/BaseMapper.php
```diff
@@ -0,0 +1,134 @@
...
+    public function count(mixed $criteria): int
+    {
+        return $this->getRepository()->count($criteria);
+    }
+
+    /**
+     * @return string the name of the entity repository
+     * e.g. "User/Model/User"
+     */
+    protected abstract function getRepositoryName() : string;
```
tomudding, line 133: There is a bonus ` ` (space) here.

module/User/src/Mapper/ApiApp.php
```diff
@@ -33,10 +22,10 @@ public function findByAppId($appId)
...
     /**
-     * @return EntityRepository
+     * @inheritDoc
      */
-    public function getRepository()
+    protected function getRepositoryName(): string
     {
-        return $this->em->getRepository(ApiAppModel::class);
+        return 'User\Model\ApiUser';
     }
```
tomudding, line 30: Implementation specific; I recommend using the `::class` constant to make obtaining the FQCN easier to understand for the maintainers, as a proper IDE will require a `use` for the entity and link it.

**You are now returning `ApiUser`, but this is the `ApiApp` mapper.**

```php
<?php

namespace User\Mapper;

// (...)
use User\Model\ApiApp as ApiAppModel;

class ApiApp extends BaseMapper
{
    // (...)

    /**
     * @inheritDoc
     */
    protected function getRepositoryName(): string
    {
        return ApiAppModel::class;
    }
}
```

I also recommend we always rename (`as`) the `use` to make it clear which class you are actually using (as we have mappers, models, and services with the same name).

module/Application/src/Mapper/BaseMapper.php
```diff
@@ -0,0 +1,134 @@
...
+        $this->em->flush();
+    }
+
+    /**
+     * Remove an entity by its ID using find
+     *
+     * @param mixed $id
+     * @throws ORMException
+     */
+    public function removeById($id)
```
tomudding, line 51: You can also use the `mixed` type for the parameter in PHP 8.

tomudding (commented):
This looks very interesting and will keep our codebase maintainable. I have some general remarks outlined in the comments below.

PR #1177, opened 2021-08-17
Fix issue with being unable to logout with persisting sessions

### LINKED ISSUE #1173: Logout not working
As we discovered during our working session (and now also confirmed by other GEWIS members), the logout functionality is not working.

The cookie is correctly expired (`now() - 1 year`), however, after redirecting the server still sends the old authentication cookie. This likely means the Session storage is not correctly cleared when logging out.

tomudding:
An issue existed where sessions based on our JWT cookie could not be invalidated. This meant that users could not logout without clearing cookies.

This also properly validates the `exp` in the JWT cookie to ensure that no one can use JWT cookies from the past. Additionally, we would occasionally put the whole `User`/`Member` model into the JWT cookie before overwriting it with the `lidnr`. To ensure we never accidentally expose this information I have removed this "functionality" (`lidnr` is used at all times).

Fixes #1173.

### DIFF
--- a/module/User/src/Authentication/AuthenticationService.php
+++ b/module/User/src/Authentication/AuthenticationService.php
@@ -19,22 +19,21 @@ class AuthenticationService implements AuthenticationServiceInterface
      *
      * @var Session
      */
-    protected $storage;
+    protected Session $storage;
 
     /**
      * Authentication adapter.
      *
      * @var Mapper|PinMapper
      */
-    protected $adapter;
+    protected Mapper|PinMapper $adapter;
 
     /**
      * Constructor.
      *
-     * @param Session $storage
-     * @param Mapper|PinMapper $adapter
+     * @param StorageInterface $storage
+     * @param AdapterInterface $adapter
      *
-     * @throws RuntimeException
      */
     public function __construct(StorageInterface $storage, AdapterInterface $adapter)
     {
@@ -47,7 +46,7 @@ class AuthenticationService implements AuthenticationServiceInterface
      *
      * @return Mapper|PinMapper
      */
-    public function getAdapter()
+    public function getAdapter(): PinMapper|Mapper
     {
         return $this->adapter;
     }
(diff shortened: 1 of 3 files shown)


module/User/src/Authentication/AuthenticationService.php
```diff
@@ -59,13 +58,14 @@ public function getAdapter()
      *
      * @return self Provides a fluent interface
      */
-    public function setAdapter(AdapterInterface $adapter)
+    public function setAdapter(AdapterInterface $adapter): static
```
Koen1999, line 61: Is it really static what is returned? It seems more like an instance.

module/User/src/Authentication/AuthenticationService.php
```diff
@@ -91,10 +91,11 @@ public function getStorage()
      *
      * @return self Provides a fluent interface
      */
-    public function setStorage(StorageInterface $storage)
+    public function setStorage(StorageInterface $storage): static
```
Koen1999, line 94: Same remark as above here.

Koen1999 (commented):
I also observe that you change the `array` type to the `Parameters` type at several locations. Is there a reason for this? It doesn't seem incorrect, but I am wondering why.

module/User/src/Authentication/AuthenticationService.php
tomudding: In some locations we require `array`s, however, Laminas always returns `Parameters` (array-like). In https://github.com/GEWIS/gewisweb/pull/1158/commits/57c6a945f17c78f102a68b31d9a3c3b8b2bf8ffe I first made this change and while a bit vague we discussed it would be better to keep it `Parameters` in the controllers and handle any conversions in the services.

If needed I can always change all occurrences of this.

Koen1999: Alright, noted.

Koen1999: But this does not relate to the static/instance comment.

tomudding: @Koen1999 I cannot read, but `static` is a special class name that refers to the class a method was called on. In this specific instance (and the one below) we do `return $this;` which makes it a "fluent" method. See [PHP RFC: Static return type](https://wiki.php.net/rfc/static_return_type) for more information (and why `return $this;` is of return type `static`).

Koen1999 (approved):

PR #1181, opened 2021-08-18
Added more details to the 403 production error page

Koen1999:


tomudding (approved):
Looks good and should also fix some of the issues with the referer when a login fails.

As I mentioned in private to Koen, ideally we implement a special mode which allows us to look at the website as if we are someone else.

PR #1183, opened 2021-08-18
Updated the glide image to PHP 8.0

Koen1999:
and updated other dependencies

Having to build images for multiple PHP versions is bad for my cache

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4907,16 +4907,16 @@
         },
         {
             "name": "league/flysystem",
-            "version": "2.2.0",
+            "version": "2.2.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/thephpleague/flysystem.git",
-                "reference": "29a3ba148287db3142412ef34f6fdcbc22c957cd"
+                "reference": "629cd411b1eb56808ce85b1dd6f9b5bf422932ad"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/thephpleague/flysystem/zipball/29a3ba148287db3142412ef34f6fdcbc22c957cd",
-                "reference": "29a3ba148287db3142412ef34f6fdcbc22c957cd",
+                "url": "https://api.github.com/repos/thephpleague/flysystem/zipball/629cd411b1eb56808ce85b1dd6f9b5bf422932ad",
+                "reference": "629cd411b1eb56808ce85b1dd6f9b5bf422932ad",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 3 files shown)


tomudding (changes requested):
Can you also update the PHP version in glide's composer.json?

As long as  don't share the same machine when building the images (we really need a pipeline) we will have bad caching of the layers.

Koen1999:
> 
> 
> Can you also update the PHP version in glide's composer.json?
> 
> Side note:
> As long as don't share the same machine when building the images (we really need a pipeline) we will have bad caching of the layers.

Well, the image also works with PHP 5.6. So we don't need to bump the requirement in my opinion.

tomudding:
> > Can you also update the PHP version in glide's composer.json?
> > Side note:
> > As long as don't share the same machine when building the images (we really need a pipeline) we will have bad caching of the layers.
> 
> Well, the image also works with PHP 5.6. So we don't need to bump the requirement in my opinion.

`league/flysystem` (and some other packages) requires at least PHP 7.2(.5), so we should bump it to that in my opinion.

Koen1999:
> 
> 
> > > Can you also update the PHP version in glide's composer.json?
> > > Side note:
> > > As long as don't share the same machine when building the images (we really need a pipeline) we will have bad caching of the layers.
> > 
> > 
> > Well, the image also works with PHP 5.6. So we don't need to bump the requirement in my opinion.
> 
> `league/flysystem` (and some other packages) requires at least PHP 7.2(.5), so we should bump it to that in my opinion.

But since this dependency is in composer.json, we will require it transitively. And if dependencies are installed using the lock file, this will be required for sure.

tomudding:
> 
> 
> > > > Can you also update the PHP version in glide's composer.json?
> > > > Side note:
> > > > As long as don't share the same machine when building the images (we really need a pipeline) we will have bad caching of the layers.
> > > 
> > > 
> > > Well, the image also works with PHP 5.6. So we don't need to bump the requirement in my opinion.
> > 
> > 
> > `league/flysystem` (and some other packages) requires at least PHP 7.2(.5), so we should bump it to that in my opinion.
> 
> But since this dependency is in composer.json, we will require it transitively. And if dependencies are installed using the lock file, this will be required for sure.

I am only asking because you asked the same for the main composer.json when I updated the main containers to PHP 8.0 (we now have PHP 8.0 specific functions, but back then we still supported 7.4). Some consistency is needed in my opinion.

I will test this thoroughly tomorrow before approving.

tomudding (approved):

PR #1185, opened 2021-08-20
Fix activities not being visible to guests

### LINKED ISSUE #1180: Guests with proper priviliges not allowed to see certain pages
[`getIdentityOrThrowException()`](https://github.com/GEWIS/gewisweb/blob/d51d83be5703b18324c324c51d2633317a389105/module/User/src/Authorization/GenericAclService.php#L68) is used in multiple places, however, in most occurrences guests do have proper privileges to do something.

An example is viewing activities, guests are allowed to do this, however, it is not possible because an exception is thrown.

tomudding:
This fixes #1180.


Koen1999 (approved):
Looks good, no comments

PR #1190, opened 2021-08-24
Fix `LocalFileReader` not returning `Stream` for downloadable files

### LINKED ISSUE #1189: Files in the Public Archive are not downloadable
_Initially reported by Rink._

An bug was introduced into the [`LocalFileReader`](https://github.com/GEWIS/gewisweb/blob/master/module/Decision/src/Controller/FileBrowser/LocalFileReader.php) via 349c94cb72cb837e63a5e0192bf85cd74502ce69:

https://github.com/GEWIS/gewisweb/blob/4c626dacd25116e536cb89a9924f575658d2b019/module/Decision/src/Controller/FileBrowser/LocalFileReader.php#L61

Previously, the `downloadFile()` method would return the `Stream` object containing the file. This was changed to return `true` to denote the creation of the `Stream` was successful. However, the `Stream` object is never used in the actual request. Furthermore, if it were to work, returning `true` would always results in a `404 Not Found` error due to the (now) flawed `?:` operator.

https://github.com/GEWIS/gewisweb/blob/4c626dacd25116e536cb89a9924f575658d2b019/module/Decision/src/Controller/DecisionController.php#L194-L197

tomudding:
This fixes #1189.

### DIFF
--- a/module/Decision/src/Controller/DecisionController.php
+++ b/module/Decision/src/Controller/DecisionController.php
@@ -191,9 +191,10 @@ class DecisionController extends AbstractActionController
                 ]
             );
         }
+
         //download the file
         $result = $fileReader->downloadFile($path);
 
-        return $result ? $this->notFoundAction() : $result;
+        return (false === $result) ? $this->notFoundAction() : $result;
     }
 }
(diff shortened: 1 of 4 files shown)


Koen1999 (approved):
Looks good to me!

Issue #888, opened 2019-05-10
Introduce safe deletion of members


Koen1999:
We must ensure that tags, activity signups and any other data attached to the user are also deleted accordingly

tomudding:
Small note with the introduction of #1419, when removing a vote from a user, their vote should be added to the anonymous count. Rink also mentioned that we can do this before a member leaves the association (like every month), that way it is completely automated and because we do not actually need the information.

tomudding:
Worth connecting this to GEWIS/gewisweb#2168 and GEWIS/gewisdb#577.

The reason this is hard is structural rather than incidental: the member rows this data hangs off arrive by a one-way, column-by-column copy from GEWISDB's ReportDB on a ~30-minute cron. So the tags, signups, votes and applications reference a row the database has no foreign key for, the importer has nowhere to put a cascade because it copies tables rather than running domain logic, and "everything attached to this member is gone" is not a statement that can be checked with a query.

Those issues discuss merging the two applications, which would turn this into ordinary referential integrity — real foreign keys, one declared cascade, one policy for what deletion means. If that direction is taken, this issue is largely solved by it; if it is not, this needs a deletion path that is explicitly designed to cross the sync, including what happens to a member removed between two runs.

PR #1198, opened 2021-08-27
Fixed UserSignup Model

Koen1999:
Sometimes a user does not have a corresponding Member because database constraints are not always met in reality


tomudding:
As discussed during our meeting, I'll close this as we favour a proper solution (for all issues that fall under #888).

Issue #1124, opened 2021-07-16
Move AclService->isAllowed checks to controllers


Koen1999:
I think this is more neat since permissions more closely relate to actions in the controller than to functions in the services.

tomudding:
Will be handled automatically in GEWISWEB-ng.

PR #1205, opened 2021-08-30
Fix `./web` not working

### LINKED ISSUE #1204: `./web` is broken, which causes the cronjobs to fail
`./web` fails because the script is unaware of any environment variables (primarily `APP_ENV`).

tomudding:
Fixes #1204.

### DIFF
--- a/web
+++ b/web
@@ -1,1 +1,15 @@
-./vendor/bin/laminas
+#!/usr/bin/env php
+<?php
+// Redefine the `APP_ENV`.
+define('APP_ENV', getenv('APP_ENV') ?: 'production');
+
+// This script should only be executed via cli.
+if (php_sapi_name() != 'cli') {
+    throw new RuntimeException("This script can only be called from the command line.");
+}
+
+// Make sure we are in the correct directory.
+chdir(dirname(__DIR__));
+
+require 'vendor/bin/laminas';
+


tomudding:
4aefcfc ensures `Laminas\Cli` can 'load' the application, however, the ACL breaks this when trying to create `activity_form_calendar_proposal` because this requires [`getEditableOrgans()`](https://github.com/GEWIS/gewisweb/blob/2899880725ac7180ae05ace3147fd40bebd2c856/module/Decision/src/Service/Organ.php#L118).

Koen1999:
> 
> 
> [4aefcfc](https://github.com/GEWIS/gewisweb/commit/4aefcfc4cdac639730a05258b60f2179d68519a6) ensures `Laminas\Cli` can 'load' the application, however, the ACL breaks this when trying to create `activity_form_calendar_proposal` because this requires [`getEditableOrgans()`](https://github.com/GEWIS/gewisweb/blob/2899880725ac7180ae05ace3147fd40bebd2c856/module/Decision/src/Service/Organ.php#L118).

This relates to #1124 and is a good example of why we should implement these changes. I propose doing so in a seperate PR. This PR already solves some issues if I am not mistaken.

Koen1999 (approved):

Issue #1209, opened 2021-09-02
Become member button


Gijsdeman:
Recently, we have received a lot of questions about how you can become a member. Of course, [join.gewis.nl](https://join.gewis.nl/) exists, but there is no way to know about this website without knowing the link. This way, a lot of new students end up on the [registration page](https://gewis.nl/user/register), which just leaves them confused.

I (read, the 40th board), think it would be a great addition to the website of in the navigation bar, a "Become member" button is added that redirects to the right page. This way, it should be clear on the homepage where you should go if you want to become a member of GEWIS.

The only issues this would give the board is to check whether new registrations are now part of the M&CS department, but this is something we can check quite easily and we already do on a daily basis.

Something like this would be fine:
![image](https://user-images.githubusercontent.com/43375368/131816615-4eef3aaf-6950-40c3-813d-4500ab22eee0.png)

Koen1999:
This is a good idea and I think a valuable addition for the beginning of the academic year. I would opt for removing this button after the 1st of December or so. But that's something for later consideration. I'll see to it that the button is implemented.

Koen1999:
@Gijsdeman I have one additional request though. join.gewis.nl does not highlight any of the benefits of becoming a GEWIS member. I propose you create a page on the website that describes these. This page can then link to join.gewis.nl and we can have the button in the navbar link to the new page (e.g. gewis.nl/become-member). Please let me know what you think.

Gijsdeman:
@Koen1999 sounds like a great idea, I will see to that such a page is created. It will most likely be a page that will just fall under /association. Thanks for the great suggestion!

tomudding:
I am closing this issue, as the changes have been merged and deployed some time ago.

PR #1212, opened 2021-09-02
Highlighted how to become a member

Koen1999:
This solves #1209


Koen1999:
Please note that I opted for a shorter text as described in the issue because it would disturb the navbar too much on mobile devices.
afbeelding
afbeelding

Koen1999:
I am also adjusting the registration page to look like this
afbeelding

rinkp:
Perhaps you can call the current registration the "activation" of one's account? That makes it distinct from registering for GEWIS

Koen1999:
> 
> 
> Perhaps you can call the current registration the "activation" of one's account? That makes it distinct from registering for GEWIS

I think we currently have the idea of registering an account and subscribing to become a member. So they are distinct, but it's a bit vague, I agree.

Koen1999:
Also, the first commit in this PR is to make pages like https://gewis.nl/vereniging/exceptionalmembers work with content from subdomains. Not really related, but I guess we need to process this change some time.

LeonVreling (approved):
Looks good

LeonVreling:
Will the board need to make the page with benefits of becoming a member, the one referenced at gewis.nl/vereniging/lidworden (or what the English URL turned out to be)

Koen1999:
> 
> 
> I think that the board still needs to make the page with the benefits of becoming a member to make it fully functional, the one referenced at [gewis.nl/vereniging/lidworden](https://gewis.nl/vereniging/lidworden) (or what the English URL turned out to be)

Yep, that's what I am currently waiting for. I will merge and deploy once that page is created in case I need to adjust the URL in the code.

tomudding (commented):
> 
> 
> afbeelding

I do not agree with this change, the navbar now takes up 1/3rd of your screen. On every single page. This affects many modern phones, as the viewport is commonly 360 pixels (it does not fit) to 412 pixels (it barely fits), see [Common Viewport Sizes](https://yesviz.com/viewport/).

Furthermore, if we have to keep adding and removing this link each year, just make it hide itself based on some date range (this can be done later).

tomudding:
I recommend we more or less go back to what we had in the past. A submenu, however, with just two links. One to the login page, and one to the new `lidworden` page. _Sidenote_, why isn't there a `-` between `lid` and `worden`? It makes the URL look a lot better.

This is what it looks like on a phone:
GEWIS-WEB #1212 Improved Menubar Becoming A Member Collapsed - cropped
GEWIS-WEB #1212 Improved Menubar Becoming A Member Extended - Cropped

And for desktop users:
image

Koen1999:
@tomudding The changes you made look good!

PR #1214, opened 2021-09-02
Added a basic watermarking setup

### LINKED ISSUE #1049: Feature request: watermark uploaded PDF's for education archive
There has been an incident in which someone copied many documents from our education archive and published them for their own gain on websites like Studeersnel. To prevent this and to get leverage on taking these down, we need to "protect" our documents. The best and probably easiest way to do this, is to watermark every document in our archive with the GEWIS logo. At a glance anyone can see these documents are "owned" by us.

With this issue, I would like to request that all PDF's that are uploaded to the education archive get watermarked automatically. This seems to be possible within PHP with a library called FPDF. However, I am not knowledgable with PHP nor the GEWISweb repository. Existing documents can be watermark all at once manually.

Koen1999:
This fixes #1049

### DIFF
--- a/module/Application/src/Service/WatermarkService.php
+++ b/module/Application/src/Service/WatermarkService.php
@@ -67,6 +67,8 @@ class WatermarkService
             $pdfPage->annotateImage($drawSettings, $sizeX / 2, $sizeY / 2, 60, $watermarkText);
             $pdfPage->mergeImageLayers(Imagick::LAYERMETHOD_FLATTEN);
 
+            $pdfPage->profileImage('icc', null);
+
             $pdf->addImage($pdfPage);
         }
 
--- a/module/Application/src/Service/WatermarkService.php
+++ b/module/Application/src/Service/WatermarkService.php
@@ -3,7 +3,6 @@
 namespace Application\Service;
 
 use DateTime;
-use Howtomakeaturn\PDFInfo\PDFInfo;
 use Imagick;
 use ImagickDraw;
 use ImagickDrawException;
@@ -14,7 +13,7 @@ use User\Authentication\AuthenticationService;
 class WatermarkService
 {
     // The font size of the watermark
-    private const FONT_SIZE = 48;
+    private const FONT_SIZE = 32;
 
     // The quality of the produced PDFs
     private const DPI = 150;
@@ -65,7 +64,7 @@ class WatermarkService
             $sizeX = $sizes['width'];
             $sizeY = $sizes['height'];
 
-            $pdfPage->annotateImage($drawSettings, $sizeX / 2, $sizeY / 2, 55, $watermarkText);
+            $pdfPage->annotateImage($drawSettings, $sizeX / 2, $sizeY / 2, 60, $watermarkText);
             $pdfPage->mergeImageLayers(Imagick::LAYERMETHOD_FLATTEN);
 
             $pdf->addImage($pdfPage);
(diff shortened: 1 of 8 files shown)


Koen1999:
Watermarks look as follows:
afbeelding

Koen1999:
Currently, there is only one issue remaining.
Adobe Acrobat cannot open the PDF and says "Insufficient data for an image". All other PDF viewers I have tried worked so far.
I'm not sure if we can resolve this issue and how.

Koen1999:
This probably relates to https://github.com/ImageMagick/ImageMagick/issues/2070 .
Apparently our php-fpm image uses a somewhat outdated Imagick version.
By adding `$pdfPage->profileImage('icc', null);` we can solve the issue for ourselves (removed ICC profile which is malformed I believe)

tomudding:
> 
> 
> Apparently our php-fpm image uses a somewhat outdated Imagick version.

We use a recent version of Imagick (as we need PHP 8 support), however, ImageMagick is old, as the debian image php-fpm is based on uses 6.9(.10). I do not recommend manually updating ImageMagick.

Can add a comment (and possibly update your commit message) to better reflect that this is not needed if we were able to upgrade to ImageMagick >7.0.10-19?

Koen1999:
> > 
> > 
> > Apparently our php-fpm image uses a somewhat outdated Imagick version.
> 
> We use a recent version of Imagick (as we need PHP 8 support), however, ImageMagick is old, as the debian image php-fpm is based on uses 6.9(.10). I do not recommend manually updating ImageMagick.
> 
> Can you add a comment (and possibly update your commit message) to better reflect that this is not needed if we were able to upgrade to ImageMagick >7.0.10-19?
> 
> 

I think the documentation in the PR is sufficient. Also can't update it on my phone currently :p

module/Application/src/Service/WatermarkService.php
```diff
@@ -51,7 +51,7 @@ public function watermarkPdf(string $path): string
         $drawSettings->setFont(self::FONT);
         $drawSettings->setTextAlignment(Imagick::ALIGN_CENTER);
         $drawSettings->setFillColor($fillPixelLight);
-        $drawSettings->setFillOpacity(0.20);
+        $drawSettings->setFillOpacity(0.40);
```
Koen1999, line 54: Doesn't this affect the readability too much?

tomudding (changes requested):
I have added some fixes which popped up when testing this with quite an assortment of PDFs. First and foremost, `Ghostscript` is necessary to process most PDF files. Sometimes, Imagick tries to load an unknown font (`Helvetica`), this is fixed by using the same font we use for the Activity CAPTCHAs. Additionally, the image profile `icc` does not always exist and 'removing' it may result in an error (fixed by adding a check).

The biggest issue preventing us from merging this is the abysmal performance for large PDF files. The error is "cache resources exhausted" (i.e., a memory limit), this occurs at `annotateImage` and `mergeImageLayers` for PDFs with many pages.

module/Application/src/Service/WatermarkService.php
tomudding: I personally do not have issues with it being `0.40`, but we should probably go to `0.30` to ensure it stays readable for everyone (`0.20` makes it unintelligible when the background is not white).

Koen1999:
Then we should probably just increase the memory limit in the php.ini :/

tomudding:
> 
> 
> Then we should probably just increase the memory limit in the php.ini :/

We cannot haphazardly increase PHP's memory limit, as we only have so much memory available on the server. It is probably something we need to discuss with the CBC. I do want to find another solution, but removing `mergeImageLayers` does not have a significant impact on the RAM usage.

tomudding (approved):
I can no longer reproduce the memory issues I had. It still takes a bit of time to generate the PDF, but I think that is acceptable.

PR #1215, opened 2021-09-02
Enable editing of activities via admin panel

LeonVreling:


tomudding:
Closing because stale, I'll work on an actual solution in the future.

RubenLWF:
LGTM! :rocket:

PR #1216, opened 2021-09-03
Reduced image sizes

Koen1999:
A tip taken from the Docker best practices


tomudding:
Perhaps we can also add `--no-install-recommends` to the `apt-get install` to install less packages. This needs to be tested though, as sometimes a "recommended" package is actually required (which is frustrating).

tomudding (approved):

PR #1217, opened 2021-09-03
Made User emails include both Dutch and English

### LINKED ISSUE #727: Always have 2 languages in emails
Currently, the language of the website determines the language of the registration email sent. This might cause some unneeded confusion, so it is probably better to send the same message in both languages, using a preamble like 'English below'

We could still use the current language to determine the order of the languages, e.g. place Dutch first if the current language is Dutch.

Koen1999:
This fixes #727

### DIFF
--- a/module/User/view/user/email/register.phtml
+++ b/module/User/view/user/email/register.phtml
@@ -1,16 +1,33 @@
-<?= sprintf($this->translate('Dear %s,'), $member->getFullName()) . "\n" ?>
+<?php $lang = 'en'; ?>
+For English, please see below.<br/><br/>
 
-<?= $this->translate('You have been registered on the website of Study association GEWIS, with the following information:') . "\n" ?>
 
-- <?= $this->translate('Membership number:') ?> <?= $user->getLidnr() . "\n" ?>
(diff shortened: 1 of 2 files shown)


module/User/view/user/email/register.phtml
```diff
@@ -1,16 +1,33 @@
-<?= sprintf($this->translate('Dear %s,'), $member->getFullName()) . "\n" ?>
+<?php $lang = 'en'; ?>
+For English, please see below.<br/><br/>
 
-<?= $this->translate('You have been registered on the website of Study association GEWIS, with the following information:') . "\n" ?>
 
-- <?= $this->translate('Membership number:') ?> <?= $user->getLidnr() . "\n" ?>
-- <?= $this->translate('Email address:') ?> <?= $user->getEmail() . "\n" ?>
+<?= sprintf($this->translate('Beste %s,'), $member->getFullName()) ?><br/><br/>
```
tomudding, line 5: Is it not easier to remove this `translate()` as well (and the seemingly unused `$lang`)?

tomudding (approved):

PR #1220, opened 2021-09-07
Update Supremum Infima API

### LINKED ISSUE #1202: Update Supremum Infima API Endpoint
Supremum will be changing the API endpoint for obtaining a random infimum. We need to change to `https://supremum.gewis.nl/api/random_infimum`. As we have discussed internally, we can no longer use the direct client approach we have currently be using as we will require an API key to access the new endpoint.

We will have the clients make a request to the website itself (e.g., `/api/infimum`), which will authenticate the user. Then we make an internal request to Supremum's API leveraging one of Laminas' `HTTP\Client\Adapters` (preferably [`cURL`](https://docs.laminas.dev/laminas-http/client/adapters/#the-curl-adapter)).

The API key is to be provided using the `X-Api-Key` header (and should be added to the Docker environment).

**Edit:** We still have to say we are `gewis.nl` to ensure we have the proper access control validation.

tomudding:
Closes #1202.

To retrieve an infimum from Supremum's API, we now have to use an API key. As discussed internally, we do not want to expose the API key to the client, however, we want to keep client-side requests to retrieve an infimum. This prevents us from having a blocking request in our responses.

The changes in the PR introduce a new `Infimum` service, which utilises Laminas' cURL client adapter to fetch an infimum. To improve response times and reduce the number of resources used to perform the requests to the Supremum API we cache the Infimum for 5 minutes in `memcached`.

Remarks: the `application_cache_infimum` can be inlined in `application_service_infimum`, however, I think that the service configuration is easier to understand if they are separate.

---

I will update the translations after merging this, to prevent this PR from being cluttered with language changes.


Koen1999 (approved):
I think it could have been done without the controller by just using the infama helper. But it looks good otherwise and would work.

tomudding:
> 
> 
> I think it could have been done without the controller by just using the infama helper. But it looks good otherwise and would work.

I initially implemented it as a `ViewHelper`, however, this would slow down the initial response. I personally think it is preferable to have the page load first and then any additional content (which includes infima).

PR #1232, opened 2021-09-27
Bring back the rainbow navigation for board members

tomudding:
As the cobo for the board is this week, I thought it would be nice to give them a little present. This brings back the beloved rainbow navigation for board members. I have improved the rainbow a bit to make sure they love it even more.

**After:**
output


Koen1999 (approved):
It looks more neat than it used to.

PR #1240, opened 2021-10-12
Improve cookie and Matomo tracking notice (and switcher)

### LINKED ISSUE #1108: Change Matomo opt-out behaviour from iframe to AJAX requests
I would like to change the Matomo behaviour for opting out of tracking. Currently we use an iframe which is blocked by all major ad blocking browser extensions. This leaves a lot of space at the bottom of the page which is not used.

I suggest installing [Ajax Opt Out](https://plugins.matomo.org/AjaxOptOut) for Matomo. This plugin provides an API endpoint to handle opting in/out of tracking. Then we can remove the iframe and add a smaller button to the bottom left of the page to handle opting in/out.

The idea is that the button will stay expanded until the user has dismissed it (e.g., clicks on "Continue") or has opted out. The button will stay at all times.

To give an impression of the button (not expanded):
Screenshot 2021-06-29 at 19-47-15 GEWIS Website

When expanded, the button will display the opt-out text and button with a link to the privacy policy.

tomudding:
This currently depends on https://github.com/matomo-org/matomo/issues/18133 being fixed and can therefore not be merged.

To ensure compliance with GDPR (and relevant laws) I have updated (read: replaced) the tracking notice produced by Matomo. In conjunction with the new privacy policy (updated by the board) this improve the experience of our users.

To do this, I have authored/forked a Matomo plugin: [TrackingOptOut](https://github.com/GEWIS/gewisweb-analytics-opt-out). This plugin provides an API-endpoint to replace the iframe functionality. It allows the client to determine and update their current tracking status.

**Old notice (no DNT):**
image

**Old notice (with DNT):**
image

New notice (no DNT):
image

New notice (with DNT):
image

When dismissed, the notice will become just the icon with reduced transparency. This is to make it non-intrusive, but also allows users to directly access their settings. The notice will stay dismissed for 1 year, each dismissal updates the expiry time.
image

Closes #1108.

### DIFF
--- a/docker/matomo/Dockerfile
+++ b/docker/matomo/Dockerfile
@@ -18,6 +18,12 @@ RUN curl -o SecurityInfo.zip \
       && rm SecurityInfo.zip \
       && mv SecurityInfo /usr/src/matomo/plugins
 
+RUN curl -o TrackingOptOut.zip \
+      https://github.com/GEWIS/gewisweb-analytics-opt-out/archive/refs/tags/v1.0.0-gewisweb.zip \
+      && unzip -j TrackingOptOut.zip -d "TrackingOptOut" \
+      && rm TrackingOptOut.zip \
+      && mv TrackingOptOut /usr/src/matomo/plugins
+
 COPY --chown=www-data:www-data config.ini.php /var/www/html/config/config.ini.php.template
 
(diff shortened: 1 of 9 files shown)


Koen1999 (approved):

tomudding:
Matomo 4.6 (and 4.6.1) has been released, with the fix for a bug which prevented us from deploying this. Undrafting this, after the Matomo images are available we should test it again to make sure everything still works as intended.

Koen1999:
I rebased on master

Koen1999:
The docker image has been updated

PR #1246, opened 2021-10-16
Add phpDoc docs

tomudding:
Adds the documentation from the PHP files as generated by `phpDocumentor` (hereafter `phpdoc`). Can be accessed via https://gewis.github.io/gewisweb (when merged).

There is currently no way to automate generation of the docs, as `phpdoc` cannot be easily installed via Composer.

Draft because I don't know if this is something I actually want to do.


Koen1999:
You can probably automate this using github actions though!

PR #1027, opened 2020-08-21
Make the new photo viewer the default photo viewer

tomudding:
This changes the routing such that the new photo viewer is the default photo viewer for normal albums and member albums.

Users can switch to the old photo viewer if they encounter problems.

This fixes #1025 and fixes #461.

PR #1255, opened 2021-10-21
Switch photo viewer

### LINKED ISSUE #461: Design: make photo module design fancy
It would be nice to have a fancier design for the photo module. Possibly including dynamic loading of images.

### LINKED ISSUE #1025: Make new photo viewer the default photo viewer
After #1017 is merged, as discussed during one of our meetings, the new photo viewer should be made the default photo viewer. To prevent problems the old photo viewer will be available via a message similar to the new photo viewer currently is.

This also concludes #461.

### LINKED ISSUE #1238: Reduce beta album viewer processing time by lazy loading image thumbnails
Currently, when viewing an album in the beta photo viewer, the browser tries to load all photo thumbnails 'at once'. This significantly increases loading time for large albums. To mitigate this we can apply lazy loading to the images, this should prevent images outside of the viewport from loading.

This feature is not supported in all browsers (not in Safari by default), but I do think it is still something want to use.

Can be enabled by adding the attributes `loading="lazy"` to our `img` tags. We can always force the first `n` images to load directly with `loading="eager"`.

### LINKED ISSUE #1247: Clicking arrows for navigation in new photo viewer skips every other photo
Using the arrows to navigate through photos using the new photo viewer skips every other photo. If you start with the first photo you will only see photos 1, 3, 5, ..., and starting with the second photo you will only see photos 2, 4, 6, .... Using the arrow keys on a keyboard does not have the same issue.

tomudding:
This supersedes #1027. This closes #1025 and closes #461. Replaces the old photo viewer with the new photo viewer.

Refer to #1025 for a list of things that need to be done (I'll add them to this PR eventually).

Fixes #1238 and fixes #1247.


tomudding:
I have updated PhotoSwipe to v5. The only thing which is currently missing is the metadata/information popup.

tomudding:
With regard to how we let users know they are supposed to vote on photos for them to become photo of the week:

1) Multiple announcements via the newsletter. Assuming we release towards the end of December, we can get at least a message in 3 newsletters.
2) If a user has not voted in the last 2-3 months, show a pulsating dot next to the vote button. When the user clicks/hovers, show a message that photos of the week are based on votes and ask if they want to vote for this photo. **(not yet implemented)**

Koen1999:
Related to this, it may be a good idea to let the images on the frontpage such as photo of the week and the birthdays picture be served from images.gewis.nl. I think this will improve our page loading time since this is currently the bottleneck.

tomudding:
> Related to this, it may be a good idea to let the images on the frontpage such as photo of the week and the birthdays picture be served from images.gewis.nl. I think this will improve our page loading time since this is currently the bottleneck.

Also add profile photos to this list.

tomudding:
Should be ready for general use. If a user has not voted the past month, a pulsing dot will be shown on the vote button. I would have loved to also add a modal with more information (when you open the photo viewer for the first time), but that is currently too complicated to implement.

image

Glide is now also used for PoTW, birthday photos, and user profile photos. Theoretically, we can now also remove thumbnails, but it is probably better to do that somewhere in the future.

Koen1999 (approved):
This PR looks good to me.

But since you intend to deploy it together with #1287 , please test it again and make sure the composer.lock file is a version that incorporates changes from both PRs.

tomudding:
Rebased and updated dependencies.

PR #1261, opened 2021-11-05
Fix exception when viewing non-existent meetings

### LINKED ISSUE #1258: Viewing non-existent meeting results in exception
Currently, visiting a non-existent meeting (e.g., BV 4242) results in an exception:

> Error: Call to a member function getType() on null in /code/module/Decision/view/decision/decision/view.phtml:2

This can be improved by showing a 404 (`return $this->notFoundAction()`) if the result to the query is `null`:

https://github.com/GEWIS/gewisweb/blob/373adcbd6de9c1585c8fdcc8be301185f9392c77/module/Decision/src/Controller/DecisionController.php#L87

tomudding:
Improves checking of return types to prevent everything from breaking down.

Fixes #1258.


Koen1999 (approved):
Looking good!

PR #1264, opened 2021-11-06
Upgrade to `laminas/laminas-form` v3 and remove datetimepicker

### LINKED ISSUE #1150: Laminas Form v3 Upgrade
There is a new update available for this package that will require us to make some changes. Most notable, we need to remove the DateTime fields and replace them by something else.

https://docs.laminas.dev/laminas-form/v3/migration/v2-to-v3/

### LINKED ISSUE #1228: Introduce `LocalisableForm` to reduce duplication in forms
After merging #1148 I recommend we implement a `LocalisableForm`, which removes some of the code duplication currently found in forms of which the model utilises `LocalisedText` (either `ActivityLocalisedText` or `CompanyLocalisedText`).

The form should take care of some of the basic functionalities (mostly related to the language checkboxes).

### LINKED ISSUE #1235: Move away from jQuery DateTimePicker and use native input interfaces
With the release of Firefox 93, `datetime-local` now also has a UI in Firefox (already present in other major browsers). This means we can remove our dependency on jQuery's DateTimePicker and use the native implementations of browsers. `Laminas\Form` already has support for this specific type.

I do recommend we do wait some time before actually doing this, as FF93 still needs to be adopted.

tomudding:
With the release of `doctrine/doctrine-module` v4.2.0 a bit more than a week ago we can finally upgrade `laminas/laminas-form` to v3.0.0.

This deprecates the `datetime` form element and view helper. To fix this, I have changed everything to `datetime-local`, which is supported with native UI by every browser.

This means that the jQuery datetimepicker is obsolete.

Closes #1150, closes #1228, and closes #1235.


Koen1999 (approved):
I did not have the time to really review this in depth, but I see no big issues so far.

tomudding:
Most of the changes are just changes to the way we "use" parts of the `laminas-form`, so no functional changes. With regards to the new behaviour of dates in forms, I have tested all those forms and did not find any issues. Additionally, the new `LocalisableForm` does also not change the behaviour of the affected forms.

PR #1278, opened 2021-11-22
Replaced i tags by em tags and b tags by strong tags

Koen1999:


tomudding (changes requested):
I have two comments:

1) You are also changing a few translations, make sure they are properly translated (by updating the Dutch versions).

2) I do not recommend replacing `i` for all FontAwesome things with `em`. We have already abused the former, and using the latter does not make it any better. Instead we should use the `span` element.

tomudding:
And we should add this to the [style guidelines](https://github.com/GEWIS/gewisweb/wiki/Style-Guidelines) (which are not yet complete), as a lot of people incorrectly use `i` / `em` and `b` / `strong`.

Koen1999:
> I have two comments:
> 
>     1. You are also changing a few translations, make sure they are properly translated (by updating the Dutch versions).
> 
>     2. I do not recommend replacing `i` for all FontAwesome things with `em`. We have already abused the former, and using the latter does not make it any better. Instead we should use the `span` element.

I have added two commits to address these issues.

tomudding (approved):

PR #1283, opened 2021-11-27
Enforce nonces in CSP for scripts and styles

### LINKED ISSUE #454: Stored XSS on membership info page

### LINKED ISSUE #1282: Enforce nonce for all inline styles and scripts
To ensure that it is impossible to exploit some of the functionalities of the website to insert custom JavaScript (and CSS) I propose we start enforcing the CSP a bit stricter with the usage of nonces.

The idea is that in the environment there is a _secret_ replacement string `NONCE_REPLACEMENT_STRING=...` (regenerated every so often). This string is present in all templates using `getenv('NONCE_REPLACEMENT_STRING')`.

Then NGINX replaces this static content with a dynamic, secure, and random identifier (`ssl_session_id`). This does not require significant changes to the NGINX Docker container/image, as the required module `ngx_http_sub_module` is already included. We only need to inject the environment variable with `envsubst` and then we can just do:

```
add_header Content-Security-Policy "default-src 'none'; script-src (...) 'nonce-$ssl_session_id'; (...)";
```

and

```
sub_filter_once off;
sub_filter_last_modified on; # <-- Preserve Last-Modified (disabled by default)
sub_filter {{ NONCE_REPLACEMENT_STRING }} $ssl_session_id;
```

### LINKED ISSUE #1335: Upgrade to CKEditor v5
This is part of the Company module overhaul and should fix the security issues associated with our outdated version of CKEditor.

tomudding:
Enforce nonces in CSP to mitigate any potential XSS attacks. Utilises NGINX's capabilities to filter a specific placeholder and replace it with a cryptographically secure random string (the nonce).

While I mentioned the use of `$ssl_session_id` in #1282, this is not possible, as NGINX acts as a reverse proxy and hence no SSL/TLS connections are created. `$request_id` is a secure alternative, as long as OpenSSL is compiled with NGINX (which is the case).

~~Currently, `laminas/laminas-developer-tools` toolbar does not work, as there is no way to inject the appropriate `nonce`-attributes into the toolbar.~~

There are two separate CSP configurations, one for `production` and one for `development`. Both differ a tiny bit in terms of what they accept. To make sure that both the `nginx` and `web` containers share the same `APP_ENV` I have created a shared set of environment variables (see `docker-compose.yml`).

Progress for each module:
- [X] Activity
- [X] Application
- [X] Company
- [X] Decision
- [X] Education
- [X] Frontpage
- [X] Photo
- [X] User

I have changed my mind about also limiting `style-src`, see https://scotthelme.co.uk/can-you-get-pwned-with-css/ for more information.

Closes #1335, closes #1282, and closes #454.


tomudding:
While I have tested most things we might want to use a service like Report URI for a month to see if everything is as we expected it to be.

tomudding:
`NONCE_REPLACEMENT_STRING` is populated manually, this should always be done using

```sh
openssl rand -base64 64
```

tomudding:
Important note about CKEditor 5, nav pills are not supported in their current form (`ul` and `li`). Functionality that may previously have been allowed is now probably stripped from the editor.

PR #1284, opened 2021-11-28
Did some code cleanup

Koen1999:


module/Activity/src/Controller/AdminCategoryController.php
```diff
@@ -29,8 +29,9 @@ class AdminCategoryController extends AbstractActionController
      */
     public function __construct(
         ActivityCategoryService $categoryService,
-        Translator $translator
-    ) {
+        Translator              $translator
+    )
+    {
```
tomudding, line 34: This is not PSR-12 compliant, the `)` and `{` should be on the same line and I would add a `,` after `$translator` (and personally I wouldn't indent `$translator` to the same level as `$categoryService`).

module/Activity/src/Controller/AdminCategoryController.php
```diff
@@ -92,7 +92,7 @@ public function deleteAction()
         $request = $this->getRequest();
 
         if ($request->isPost()) {
-            $categoryId = (int) $this->params('id');
+            $categoryId = (int)$this->params('id');
```
tomudding, line 95: I think we have been changing these type casts many times. We should really decide what to do; space, no space, or no type cast at all (if the route is correctly specified).

module/Activity/src/Controller/AdminCategoryController.php
```diff
@@ -138,7 +137,12 @@ public function editAction()
         $categoryData['language_english'] = null !== $categoryData['nameEn'];
         $form->setData($categoryData);
 
-        $viewModel = new ViewModel(['form' => $form, 'action' => $translator->translate('Update Activity Category')]);
+        $viewModel = new ViewModel(
+            [
+                'form' => $form,
+                'action' => $this->translator->translate('Update Activity Category')
```
tomudding, line 143: We should add a `,` here.

module/Activity/src/Controller/AdminController.php
```diff
@@ -371,9 +371,8 @@ protected function redirectActivityAdminRequest(
     public function externalSignoffAction()
     {
         $signupId = (int)$this->params('id');
-        $signupMapper = $this->signupMapper;
```
tomudding, line 374: This empty line can potentially be removed.

module/Company/src/Form/JobLabel.php
```diff
@@ -72,16 +72,6 @@ public function __construct(Translator $translator)
...
-    /**
-     * @inheritDoc
-     *
-     * @return array
-     */
-    public function getInputFilterSpecification(): array
-    {
-        return parent::getInputFilterSpecification();
-    }
-
```
tomudding, line 84: While this can be removed, I think it makes it clearer to the developer that the `InputFilter` of this form will contain the necessary elements from the `Localisable` form (the checkboxes).

module/Decision/src/Controller/DecisionController.php
```diff
@@ -179,15 +179,14 @@ public function filesAction()
             $path = '';
         }
 
-        $fileReader = $this->fileReader;
-        if (!$fileReader->isAllowed($path) || 1 === preg_match('(\/\.\.\/|\/\.\.$)', $path)) {
+        if (!$this->fileReader->isAllowed($path) || 1 === preg_match('(\/\.\.\/|\/\.\.$)', $path)) {
```
tomudding, line 182:
```php
if (
    !$this->fileReader->isAllowed($path)
    || 1 === preg_match('(\/\.\.\/|\/\.\.$)', $path)
) {
```

module/Education/src/Service/Exam.php
```diff
@@ -225,8 +225,8 @@ protected function bulkEdit(array $data, string $type): bool
          * to process the upload. This does allow us to get the ID of the
          * exam, which we need in the upload process.
          */
-        $storageService = $this->storageService;
```
tomudding, line 228: 🤔 (good change)

tomudding (changes requested):
Looks good overall, some minor things I believe should be changed.

module/Company/src/Form/JobLabel.php
Koen1999: The functionality is exactly the same. I found this very confusing mostly. There's no need to add this. Less code is better code.

module/Activity/src/Controller/AdminCategoryController.php
Koen1999: Fixed this and added it to Phpstorm code style

module/Activity/src/Controller/AdminCategoryController.php
Koen1999: Not sure how this happened....

Koen1999:
Sadly, I cannot get PhpStorm to automatically group or even keep grouped use statements.

tomudding (approved):

PR #1295, opened 2021-12-04
Added a basic testing framework using laminas-test

Koen1999:
Tests can be run using `make runtest`.
This PR is more about adding the framework than anything else. Therefore, there are only 5 basic tests testing index pages of some modules.

Currently, it is using the 4.x branch of laminas-test which is still in development. This is because the most recently released version is incompatible with our project. PHP 8.1 support also still has to be added for laminas-test, but I am confident that it will work out.

I found the first bug with the testing framework already :p

```
PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

....F                                                               5 / 5 (100%)

Time: 00:00.434, Memory: 30.00 MB

There was 1 failure:

1) PhotoTest\ControllerTest::testIndexActionCanNotBeAccessed
Failed asserting response code "403", actual status code is "500"

Exceptions raised:
Exception 'TypeError' with message 'Photo\Service\Album::getAlbumsByYear(): Argument #1 ($year) must be of type int, null given, called in /code/module/Photo/src/Controller/PhotoController.php on line 54' in /code/module/Photo/src/Service/Album.php:145

/code/vendor/laminas/laminas-test/src/PHPUnit/Controller/AbstractControllerTestCase.php:436
/code/module/Photo/test/ControllerTest.php:12

FAILURES!
Tests: 5, Assertions: 4, Failures: 1.
```


Makefile
```diff
@@ -44,6 +44,15 @@ rundev: builddev
 stop:
 		@docker-compose down
 
+runtest: rundev
+		@sleep 1
```
tomudding, line 48: Is this `sleep` is really necessary to make this work?

Makefile
```diff
@@ -44,6 +44,15 @@ rundev: builddev
...
+		@docker-compose exec -T web ./vendor/phpunit/phpunit/phpunit --bootstrap ./bootstrap.php --configuration ./phpunit.xml
+
+runcoverage: rundev
```
tomudding, line 51: Because it is not specified here.

tomudding (commented):
General comment, otherwise it looks good.

Makefile
Koen1999: I found that somehow the database container was not fully ready before the next command was executed resulting in errors. This prevented that.

Makefile
Koen1999: This isn't run in GitHub actions. It behaved slightly different even though it should not

Koen1999: (referring to #1297 )

tomudding (approved):

PR #1297, opened 2021-12-04
Added Github actions

Koen1999:
Please see this PR for an example of what the checks look like: https://github.com/Koen1999/gewisweb/pull/5

The PhpUnit workflow will only work after #1295 

Preferably merge #1295 and #1298  first.

There will be some merge conflicts in the lockfile. Just run `php composer.phar update -W` to resolve them.

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -3,8 +3,6 @@ on: [push]
 jobs:
     PHPCS:
         runs-on: ubuntu-latest
-        env:
-            DOCKER_BUILDKIT: '1'
         steps:
             - name: Check out repository code
               uses: actions/checkout@v2
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -9,4 +9,4 @@ jobs:
             - run: cd ${{ github.workspace }}
             - run: php composer.phar install
             - name: Running PHPCS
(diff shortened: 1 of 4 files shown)


tomudding:
I do not agree with running everything inside Docker, especially all the static analysis (PHP_CS, PHPStan, Psalm, etc.). It is a lot easier to just use the infrastructure provided by GH Actions (and user actions). When I have time later today I will provide a working example to show you what I mean (please don't do it yourself).

Koen1999:
> I do not agree with running everything inside Docker, especially all the static analysis (PHP_CS, PHPStan, Psalm, etc.). It is a lot easier to just use the infrastructure provided by GH Actions (and user actions). When I have time later today I will provide a working example to show you what I mean (please don't do it yourself).

A problem with this is that PHPStan, for example, does not work outside the Docker environment because environment variables are not set. And the Doctrine extension for PHPStan requires a connection with the database.

tomudding (approved):

tomudding:
I opened #1308 to show you how I think this should be done. It has one caveat, it does not support the PHPStan baseline generation on whatever branch the PR is based on (can be easily inferred from the `pull_request` object). Because of this the job fails until we have actually fixed the errors on master (which I am currently working on).

Koen1999:
Obsoleted by #1314

PR #1306, opened 2021-12-20
Exchange rsync for sFTP to fix public archive sync

### LINKED ISSUE #1303: Update public archive sync

tomudding:
The old file server will stop working not too far in the future and the public archive is already present on the new server. Hence, we can already switch to the new server.

Unfortunately, this server does not have rsync capabilities, nor does it allow SCP. Hence, we need to switch to the only supported protocol, sFTP.

Previously, rsync would keep track of files to be removed. Implementing this with sFTP is not feasible, as such it is just easier to remove all files before downloading them. If something fails during the download, the files will be missing, but I think this is acceptable, as we can easily manually sync the files (unless they are no longer present on the remote).

Additionally, I have put the remote address in an environment variable (`SSH_REMOTE`) to make it easier to configure it if necessary.

Code has been tested on the test website (you can check it out).

Fixes #1303.

### DIFF
--- a/.env.dist
+++ b/.env.dist
@@ -43,6 +43,7 @@ DOCKER_DB2_DATABASE=database
 # These settings are for accessing the server containing the public archive for synchronization (production only)
 SSH_USERNAME=gewis
 SSH_PASSWORD=gewis
+SSH_REMOTE=gewis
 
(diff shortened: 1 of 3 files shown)


Koen1999 (approved):
Looks good!

PR #1308, opened 2021-12-22
Add GitHub Actions for analysis of pull requests

tomudding:
This is an alternative for #1297. It provides faster completion times for jobs, caching, and proper annotations on files (with `cs2pr`).

However, it currently has one big downside. There is no easy way to implement the PHPStan baseline functionality, as such the job will fail because master currently has errors. Most of these errors are easily fixed and that should happen before we would merge this PR.

You can see an example here or at https://github.com/tomudding/gewisweb/pull/1 (the annotations do not appear, but that is because the file modified is not in the list).

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -1,0 +1,68 @@
+name: PHP Static Analysis
+
+# Only run this action on pull requests (creation, synchronisation, and reopening).
+on: [pull_request]
+
+jobs:
+    phpstan:
+        name: PHPStan
+        runs-on: ubuntu-latest
+
+        # Create a MySQL service.
+        services:
+            mysql:
+                image: mariadb
+                env:
+                    MYSQL_ROOT_PASSWORD: gewis
+                    MYSQL_DATABASE: gewis
+                    MYSQL_USER: gewis
+                    MYSQL_PASSWORD: gewis
+                ports:
+                    - 3306:3306
+                options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+        steps:
+            -   name: Checkout head branch
+                uses: actions/checkout@v2
+
+            -   name: Install PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: '8.1'
+                    extensions: calendar, curl, exif, gd, intl, opcache, pgsql, pdo_mysql, pdo_pgsql, soap, zip, imagick, memcached, xdebug
+                    tools: cs2pr
+
+            -   name: Get Composer cache directory
+                id: composer-cache-head
+                run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            -   name: Cache dependencies
+                uses: actions/cache@v2
+                with:
+                    path: ${{ steps.composer-cache-head.outputs.dir }}
+                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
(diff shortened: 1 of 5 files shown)


phpstan.neon
```diff
@@ -1,6 +1,5 @@
 includes:
     - phpstan/phpstan-baseline.neon
-    - phpstan/phpstan-baseline-pr.neon
```
Koen1999, line 3: Why remove this?

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,68 @@
+name: PHP Static Analysis
```
Koen1999, line 1: We should still be able to do this by generating a baseline on master branch automatically. I actually managed to get that part working in my PR.

.github/workflows/static-analysis.yml
```diff
@@ -0,0 +1,68 @@
...
+
+            -   name: Cache dependencies
+                uses: actions/cache@v2
+                with:
+                    path: ${{ steps.composer-cache-head.outputs.dir }}
+                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+                    restore-keys: ${{ runner.os }}-composer-
+
+            -   name: Install dependencies
+                run: composer install --prefer-dist --no-interaction
```
Koen1999, line 47: I think you can leaveout `--prefer-dist`

Koen1999 (commented):
I still need to actually run the actions as well. But I think the biggest issue is that PhpStan does not work by generating a baseline from the master branch.

Koen1999:
Another note after having looked at the PR on your repository is that I probably prefer the output without cs2pr because it hides a lot of useful information like the file containing the error and the line number.

tomudding:
> Another note after having looked at the PR on your repository is that I probably prefer the output without cs2pr because it hides a lot of useful information like the file containing the error and the line number.

The idea of cs2pr is that it automatically annotates the problems in the files you changed. See https://github.com/staabm/annotate-pull-request-from-checkstyle for an example.

Makefile
```diff
@@ -76,21 +76,6 @@ copyconf:
 		cp config/autoload/laminas-developer-tools.local.php.dist config/autoload/laminas-developer-tools.local.php
 
 phpstan:
-		@docker-compose exec web echo "" > phpstan/phpstan-baseline-pr.neon
-		@docker-compose exec web vendor/bin/phpstan analyse -c phpstan.neon --memory-limit 1G
-
-phpstanpr:
```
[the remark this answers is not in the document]
tomudding: The idea behind this was that if master is cleaned up you do no longer need this, as it will be only comparing new changes.

phpstan.neon
tomudding: For the same reason as above.

Koen1999:
> > Another note after having looked at the PR on your repository is that I probably prefer the output without cs2pr because it hides a lot of useful information like the file containing the error and the line number.
> 
> The idea of cs2pr is that it automatically annotates the problems in the files you changed. See https://github.com/staabm/annotate-pull-request-from-checkstyle for an example.

Ooooh, like that.
Then ignore my last comment.
Overall, it's nice to see the running time improved by quite a bit.

Makefile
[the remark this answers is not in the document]
tomudding: But it does require a clean master (which I am currently working on).

tomudding:
Closed in favour of #1314

PR #1309, opened 2021-12-22
Fix PHPStan issues

tomudding:
To enable #1308 (if chosen) to work or as a general improvement, this fixes most of the current PHPStan issues present on master.

There are some remaining issues, of which most should be ignored in the baseline:

1) There is currently no functionality to better type these values in `laminas/form`.
```php
 ------ ------------------------------------------------------------------------------
  Line   module/Activity/src/Controller/AdminOptionController.php
 ------ ------------------------------------------------------------------------------
  92     Call to an undefined method Laminas\Form\ElementInterface::setCount().
  102    Call to an undefined method Laminas\Form\ElementInterface::count().
  103    Call to an undefined method Laminas\Form\ElementInterface::setCount().
  114    Call to an undefined method Laminas\Form\ElementInterface::populateValues().
  165    Call to an undefined method Laminas\Form\ElementInterface::setCount().
  185    Call to an undefined method Laminas\Form\ElementInterface::setCount().
 ------ ------------------------------------------------------------------------------
```

2) There is currently no functionality to better type these values in `laminas/form`.
```php
 ------ -------------------------------------------------------------------------------
  Line   module/Activity/src/Form/Activity.php
 ------ -------------------------------------------------------------------------------
  201    Call to an undefined method Laminas\Form\ElementInterface::getValueOptions().
  206    Call to an undefined method Laminas\Form\ElementInterface::setValueOptions().
  208    Call to an undefined method Laminas\Form\ElementInterface::getValueOptions().
  213    Call to an undefined method Laminas\Form\ElementInterface::setValueOptions().
  220    Call to an undefined method Laminas\Form\ElementInterface::setValueOptions().
  282    Call to an undefined method Laminas\Form\ElementInterface::getFieldSets().
 ------ -------------------------------------------------------------------------------
```

3) `unwrap()` is part of `PersistentCollection`. We could change the type (for all `(Array)Collections`) to `Collection|PersistentCollection`, just `PersistentCollection` does not work as we have to initialise the attributes when creating the entities (before persisting them).
```php
 ------ -------------------------------------------------------------------------------
  Line   module/Activity/src/Service/Activity.php
 ------ -------------------------------------------------------------------------------
  461    Call to an undefined method Doctrine\Common\Collections\Collection::unwrap().
 ------ -------------------------------------------------------------------------------
```

4) There is currently no functionality to better type these values in `laminas/form`.
```php
 ------ -------------------------------------------------------------------------------
  Line   module/Decision/src/Form/Notes.php
[...]

PR #1311, opened 2021-12-22
Fix typing for all modules

### LINKED ISSUE #1126: PHP 7.4: Add type hints to functions and properties
This is a newly supported feature but also affects the behaviour of the runtime execution.

I propose first improving the PhpDoc for all functions and reviewing the stability with PhpStan.
After that we can automatically generate the type hints with the PHP Code Beautifier.

Would require serious testing.

tomudding:
This PR is based on #1309 and should be rebased once that is merged. Fixes (almost) all typing issues which are currently present in the code.

To accomplish this, I have locally changed my PHPStan configuration to be like this:

```yml
level: 6
treatPhpDocTypesAsCertain: true
checkMissingIterableValueType: false
checkGenericClassInNonGenericObjectType: false
```

The last two are necessary at the moment, because I do not have the time and motivation to determine all correct types for `arrays`, `Collection`s, etc.

**Modules:**
- [X] Activity
- [x] Application
- [x] Company
- [x] Decision
- [x] Education
- [x] Frontpage
- [x] Photo
- [x] User

This closes #1126.

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -16,5 +16,8 @@
       <option name="level" value="5" />
       <option name="config" value="$PROJECT_DIR$/phpstan.neon" />
     </inspection_tool>
+    <inspection_tool class="PhpUnused" enabled="true" level="WEAK WARNING" enabled_by_default="true">
+      <option name="SUPPRESS_GETTERS_AND_SETTERS" value="true" />
+    </inspection_tool>
   </profile>
 </component>
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -8,7 +8,10 @@
       <option name="SHOW_SNIFF_NAMES" value="true" />
       <option name="EXTENSIONS" value="php" />
     </inspection_tool>
-    <inspection_tool class="PhpStanGlobal" enabled="true" level="WEAK WARNING" enabled_by_default="true">
+    <inspection_tool class="PhpMissingFieldTypeInspection" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="PhpMissingParamTypeInspection" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="PhpMissingReturnTypeInspection" enabled="true" level="WARNING" enabled_by_default="true" />
+    <inspection_tool class="PhpStanGlobal" enabled="false" level="WEAK WARNING" enabled_by_default="false">
       <option name="FULL_PROJECT" value="true" />
       <option name="level" value="5" />
       <option name="config" value="$PROJECT_DIR$/phpstan.neon" />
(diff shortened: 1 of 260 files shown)


Koen1999:
I think the order could better be not alphabetic due to inter-module dependencies. Try doing the User and Decision module next and the Frontpage last.

`Fatal error: Declaration of User\Authorization\GenericAclService::getRole() must be compatible with Application\Service\AbstractAclService::getRole(): Laminas\Permissions\Acl\Role\RoleInterface|string in /code/module/User/src/Authorization/GenericAclService.php on line 42`

tomudding:
> I think the order could better be not alphabetic due to inter-module dependencies. Try doing the User module next.
> 
> `Fatal error: Declaration of User\Authorization\GenericAclService::getRole() must be compatible with Application\Service\AbstractAclService::getRole(): Laminas\Permissions\Acl\Role\RoleInterface|string in /code/module/User/src/Authorization/GenericAclService.php on line 42`

This exposes my development strategy a bit. I can only do this because we (at the moment) cannot use `git bisect`. I push per module, but locally I have the changes from all other parts that are broken.

When I pushed the updates for these three branches everything was already fixed locally but not on the remote. Regarding the order, a bit too late because I already have finished all modules in alphabetical order (I just need to push them).

I do intend to squash the 8 commits unless you prefer I keep it this way.

Koen1999:
> > I think the order could better be not alphabetic due to inter-module dependencies. Try doing the User module next.
> > `Fatal error: Declaration of User\Authorization\GenericAclService::getRole() must be compatible with Application\Service\AbstractAclService::getRole(): Laminas\Permissions\Acl\Role\RoleInterface|string in /code/module/User/src/Authorization/GenericAclService.php on line 42`
> 
> This exposes my development strategy a bit. I can only do this because we (at the moment) cannot use `git bisect`. I push per module, but locally I have the changes from all other parts that are broken.
> 
> When I pushed the updates for these three branches everything was already fixed locally but not on the remote. Regarding the order, a bit too late because I already have finished all modules in alphabetical order (I just need to push them).
> 
> I do intend to squash the 8 commits unless you prefer I keep it this way.

You can squash since the codebase is not very functional inbetween commits.

Koen1999:
I suggest you rebase this on master so we can see the GitHub actions in action

tomudding:
> I suggest you rebase this on master so we can see the GitHub actions in action

Yes, I messed up the squash it is taking a bit longer than expected.

tomudding:
@Koen1999 The results are available, it would have been nice if we actually had unit tests for everything. It does unfortunately appear that the PHPStan action is borked.

Koen1999:
I fixed it on master branch @tomudding

Koen1999 (approved):

tomudding:
Please do not merge just yet. While I am pretty sure there aren't any more bugs then there were previously in master (probably a bit less even) I do need at least one more day to check it.

Koen1999:
Since you also fixed the level 6 issues, I suggest raising the Phpstan level in the config as well.

tomudding:
I'll be merging this in a bit as I do not see any issues related to the typing changes we have implemented.

PR #1313, opened 2021-12-23
Fixes for testing framework

### LINKED ISSUE #1305: ControllerTest in User does not comply with PSR-4 autoloading
Its namespace `UserTest\Authentication`, while it is expected to be `UserTest`.

Koen1999:
This fixes #1305 .
Moreover, it also addresses a deprecation warning thrown by DoctrineORMModule with PHP 8.1.


tomudding (approved):
Yay! They finally released v4.2.0, no more deprecation warnings.

Koen1999:
> Yay! They finally released v4.2.0, no more deprecation warnings.

Shhhh, they didn't release it yet.
But I changed the minimum stability to dev to circumvent that. Nevertheless, I ensured stable releases are preferred. So with the next release/dependency update, we should automatically be on the stable 4.2.0 release.
So, this does not affect other packages as stable releases are always chosen if available.

tomudding:
> > Yay! They finally released v4.2.0, no more deprecation warnings.
> 
> Shhhh, they didn't release it yet. But I changed the minimum stability to dev to circumvent that. Nevertheless, I ensured stable releases are preferred. So with the next release/dependency update, we should automatically be on the stable 4.2.0 release. So, this does not affect other packages as stable releases are always chosen if available.

Changes requested! No, I think we will be fine as long as we do not haphazardly keep updating dependencies whenever there is an update.

PR #1314, opened 2021-12-23
Add GitHub Actions for analysis of pull requests

Koen1999:
This supersedes #1297 and #1308 .
Actions are run as in #1308 , so without too much usage of Docker and thus ensuring fast actions.
Issues raised in #1308 are addressed however. Most notably. A baseline is automatically generated from the master branch and the PR verifies that no new issues are introduced. So, this does not necessarily require resolving all issues of a certain level in case we desire to raise the level at some point.

For an example, refer to https://github.com/Koen1999/gewisweb/pull/7
Currently PHPStan somehow is not able to add all errors to the baseline, thus raising one error. But this should be resolved after #1309 .

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,26 @@
+name: PHP Coding Standards
+
+# Only run this action when pushing to master or on pull requests (creation, synchronisation, and reopening).
+on:
+    push:
+        branches:
+            - master
+    pull_request:
+
+jobs:
+    php-codesniffer:
+        name: PHP_CodeSniffer
+        runs-on: ubuntu-latest
+
+        steps:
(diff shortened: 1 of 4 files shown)


tomudding:
You could have committed the changes directly to #1308 but this looks good too.

For PHP_CodeSniffer we might actually want to disable warnings with `-n`/`--warning-severity=0` to prevent the creation of annotations for lines which exceed 120 characters. While it will also suppress other warnings, I think that at least for GitHub Actions this is a good compromise.

Koen1999:
I think we can still include the warnings. But we shouldn't see them as a blocking issue.

tomudding (approved):

tomudding:
I'll merge this even though PHPStan is failing :P

PR #1321, opened 2021-12-24
Added automatic tests asserting non-500 status code

Koen1999:
Some tests clearly are better than no tests. But we should still work on writing some better tests.

The main thing with testing I am currently struggling with is that many pages require authentication which is difficult to test.

### DIFF
--- a/module/Application/test/AutomaticControllerTest.php
+++ b/module/Application/test/AutomaticControllerTest.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace ApplicationTest;
+
+use Laminas\Router\Exception\InvalidArgumentException;
+use Laminas\Router\Http\Literal;
+use Laminas\Router\Http\Part;
+use Laminas\Router\Http\Segment;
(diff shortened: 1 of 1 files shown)


tomudding:
> The main thing with testing I am currently struggling with is that many pages require authentication which is difficult to test.

I haven't check this PR yet, but I want to comment on this. I have had the idea for some time to replace the exceptions we are currently throwing with a `notAllowedAction($some parameter)` (just like `notFoundAction`). That way we do not have to deal with all the problems of breaking the pages (which I still believe is the source of the many memcache/session errors we get). We would need an extension of the `AbstractActionController` but I think it is reasonable. And it ties in with #1124.

I'll add an issue for this specific idea, to allow for a discussion.

tomudding (commented):
All the parse functions can be typed with `void` too, right? And I would add the trailing comma after the `get_class()` calls. Otherwise it looks good.

tomudding (approved):

PR #1327, opened 2021-12-24
Added more tests

Koen1999:
This should be merged after #1309 

Still WIP, a lot more mappers todo....


Koen1999:
@tomudding , if you have time, could you help me out fix the bug causing the newly added tests in the Photo module to fail?
It appears as if the Album Service does not use the mock for the Acl service.

tomudding:
> @tomudding , if you have time, could you help me out fix the bug causing the newly added tests in the Photo module to fail? It appears as if the Album Service does not use the mock for the Acl service.

Very peculiar, as far as I can tell it should work. The structure is the same as controllers and services in some of the other modules. I'll take a closer look tomorrow as I do not really have the time to do so right now.

Koen1999:
@tomudding , I think I found the cause!

Koen1999:
Your help is still welcome in figuring out the new bugs for the Activity module though :p

I think you are way more familiar with that module.

My guess would be that the activity module may invoke `isAllowed` checks on ACL services of other modules which are not mocked. If that's the case, I suggest refactoring the code to remove these isAllowed checks. These inter-module dependencies are very annoying.

Koen1999:
I think this is done for now.

Future ideas are: 
- Extending the BaseControllerTest in each module and mocking some mappers in these extension classes to provide the application with data to process to increase branch/line coverage.
- Adding more tests to check that access is denied for some routes
- Really adding lots of tests for mappers
- Really adding lots of tests for models
- Really adding lots of tests for forms
- Really adding lots of tests for services

Coverage looks like this right now. Still not amazing, but definitely better than no coverage.
Screenshot_20211226_124634

tomudding:
From time to time I am getting the following failure:

> 1) EducationTest\ControllerTest::testEducationCourseDownloadActionCanBeAccessedAsUser
> Failed asserting response code was NOT "500"
> 
> Exceptions raised:
> Exception 'PHPUnit\Framework\Error\Notice' with message 'tempnam(): file created in the system's temporary directory' in /code/module/Application/src/Service/WatermarkService.php:64
> 
> /code/vendor/laminas/laminas-test/src/PHPUnit/Controller/AbstractControllerTestCase.php:451
> /code/module/Education/test/ControllerTest.php:44

Clearly it is not a failure of our code, but it is frustrating because this test (like the automatic controller test) takes a long time on my system (which is pretty good).

tomudding (changes requested):
For some reason I cannot push to this branch: so here is the git diff of what I believe should be changed:

```diff
diff --git a/module/Application/test/BaseControllerTest.php b/module/Application/test/BaseControllerTest.php
index 18502878..6612e2f7 100644
--- a/module/Application/test/BaseControllerTest.php
+++ b/module/Application/test/BaseControllerTest.php
@@ -11,8 +11,10 @@ use Laminas\ServiceManager\ServiceManager;
 use Laminas\Test\PHPUnit\Controller\AbstractHttpControllerTestCase;
 use PHPUnit\Framework\MockObject\MockObject;
 use User\Authentication\AuthenticationService;
-use User\Model\User;
-use User\Model\UserRole;
+use User\Model\{
+    User,
+    UserRole,
+};

 abstract class BaseControllerTest extends AbstractHttpControllerTestCase
 {
@@ -37,7 +39,7 @@ abstract class BaseControllerTest extends AbstractHttpControllerTestCase
         $this->getApplication();
     }

-    protected function setUpMockedServices()
+    protected function setUpMockedServices(): void
     {
         $this->setUpMockAuthService();
         $this->setUpMockUserMapper();
@@ -69,7 +71,7 @@ abstract class BaseControllerTest extends AbstractHttpControllerTestCase
     /**
      * Variation of {@link Application::init} but without initial bootstrapping.
      */
-    private static function initServiceManager($configuration = []): ServiceManager
+    private static function initServiceManager(array $configuration = []): ServiceManager
     {
         // Prepare the service manager
         $smConfig = $configuration['service_manager'] ?? [];
@@ -85,8 +87,10 @@ abstract class BaseControllerTest extends AbstractHttpControllerTestCase
         return $serviceManager;
     }

-    private function bootstrapApplication($serviceManager, $configuration = []): Application
-    {
+    private function bootstrapApplication(
+        ServiceManager $serviceManager,
+        array $configuration = [],
+    ): Application {
         // Prepare list of listeners to bootstrap
         $listenersFromAppConfig = $configuration['listeners'] ?? [];
         $config = $serviceManager->get('config');
```

Koen1999:
I should note that I had difficulties creating mocks for the BaseMapperTest. I wanted to mock the EntityManager to some degree to be able to test whether our query builders worked without actually executing the queries. But I was unable to do this. Therefore, no tests were written that could potentially alter the database state.

tomudding:
> I should note that I had difficulties creating mocks for the BaseMapperTest. I wanted to mock the EntityManager to some degree to be able to test whether our query builders worked without actually executing the queries. But I was unable to do this. Therefore, no tests were written that could potentially alter the database state.

I recommend we provide an alternative configuration for the tests to make sure that Doctrine uses SQLite (`PDO_SQLITE`) instead of the actual database. Should be as simple as calling `EntityManager::create()` to create a custom `EntityManager`.

Additionally, SQLite is pretty light weight so no big impact while testing using an in-memory database.

Koen1999:
> > I should note that I had difficulties creating mocks for the BaseMapperTest. I wanted to mock the EntityManager to some degree to be able to test whether our query builders worked without actually executing the queries. But I was unable to do this. Therefore, no tests were written that could potentially alter the database state.
> 
> I recommend we provide an alternative configuration for the tests to make sure that Doctrine uses SQLite (`PDO_SQLITE`) instead of the actual database. Should be as simple as calling `EntityManager::create()` to create a custom `EntityManager`.
> 
> Additionally, SQLite is pretty light weight so no big impact while testing using an in-memory database.
> 
> **Edit:** See this [SO answer](https://stackoverflow.com/a/19214981/5914775) for more information.

I tried something similar already and I just gave up on this for now.

tomudding:
@Koen1999 I have seen the error about the usage of `CONCAT_WS`, which is not supported in SQLite. For what it is worth, I'd recommend adding the function to SQLite via DBAL (if you haven't looked into it already).

If you do not want to write the code yourself, Drupal has [this](https://git.drupalcode.org/project/drupal/-/blob/10.0.x/core/lib/Drupal/Core/Database/Driver/sqlite/Connection.php#L278-290) licensed under GPLv2, which is compatible with our GPLv3 license (requires proper attribution etc.).

Then it should be as simple as calling:

```php
$db = $entityManager->getConnection()->getWrappedConnection();
$db->sqliteCreateFunction('CONCAT_WS', [$this, 'sqlFunctionConcatWs']));
```

The `$this` can be replaced with the FQCN of the class the function is in (e.g., `ApplicationTest\TestConfigProvider::class`).

Koen1999:
@tomudding  I have simply marked that test as skipped for now.

Koen1999:
Alright. So, for the time being I am done with testing. So you can review and merge this PR again if you have time.

.github/workflows/unit-tests.yml
```diff
@@ -63,13 +50,6 @@ jobs:
...
-            -   name: Create database
-                # GitHub Actions do not support custom hostnames like in Docker, hence we need to overwrite this value.
-                # Furthermore, we cannot declare it above, as it will be overwritten by the previous step.
-                env:
-                    DOCKER_DB_HOST: 127.0.0.1
-                run: ./orm orm:schema-tool:update --force
-
             -   name: Run PHPUnit
                 env:
                     DOCKER_DB_HOST: 127.0.0.1
```
tomudding, line 55: I guess that this line can also be removed?

module/Activity/src/Controller/AdminController.php
```diff
@@ -22,8 +22,7 @@
     AbstractContainer,
     Container as SessionContainer,
 };
-use Laminas\Stdlib\Parameters;
-use Laminas\Stdlib\ResponseInterface;
+use Laminas\Stdlib\{Parameters, ResponseInterface};
```
tomudding, line 25: Would you mind changing this to:

```php
use Laminas\Stdlib\{
    Parameters,
    ResponseInterface,
};
```

module/Application/src/Mapper/BaseMapper.php
```diff
@@ -5,10 +5,10 @@
 use Doctrine\DBAL\Connection;
 use Doctrine\ORM\{
     EntityManager,
+    EntityNotFoundException,
     EntityRepository,
     OptimisticLockException,
-    Exception\ORMException,
-};
+    Exception\ORMException};
```
tomudding, line 11:
```php
    Exception\ORMException,
};
```

module/Application/test/BaseControllerTest.php
```diff
@@ -2,15 +2,198 @@
...
+use DateTime;
+use Decision\Model\Member;
+use Doctrine\Common\Collections\ArrayCollection;
+use Laminas\Mvc\Application;
+use Laminas\Mvc\Service\ServiceManagerConfig;
+use Laminas\ServiceManager\ServiceManager;
 use Laminas\Test\PHPUnit\Controller\AbstractHttpControllerTestCase;
+use PHPUnit\Framework\MockObject\MockObject;
+use User\Authentication\AuthenticationService;
+use User\Model\{User, UserRole};
```
tomudding, line 14: Also here:

```php
use User\Model\{
    User,
    UserRole,
};
```

module/Application/test/BaseControllerTest.php
```diff
@@ -2,15 +2,198 @@
...
+
+        // Load modules
+        $serviceManager->get('ModuleManager')->loadModules();
+
+        return $serviceManager;
+    }
+
+    private function bootstrapApplication(
+        ServiceManager $serviceManager,
+        array $configuration = []
```
tomudding, line 87:
```php
        array $configuration = [],
```

(trailing comma)

module/Application/test/Mapper/BaseMapperTest.php
```diff
@@ -0,0 +1,188 @@
...
+        $events = $this->application->getEventManager();
+        $this->application->getServiceManager()->get('SendResponseListener')->detach($events);
+
+        return $this->application;
+    }
+
+    /**
+     * Variation of {@link Application::init} but without initial bootstrapping.
+     */
+    private static function initServiceManager($configuration = []): ServiceManager
```
tomudding, line 62: `array` typing like with the `BaseControllerTest`.

module/Application/test/Mapper/BaseMapperTest.php
```diff
@@ -0,0 +1,188 @@
...
+
+        // Load modules
+        $serviceManager->get('ModuleManager')->loadModules();
+
+        return $serviceManager;
+    }
+
+    protected function setUpMockedServices()
+    {
+    }
```
tomudding, line 80: If this is still required it can be typed `: void`.

module/Application/test/Mapper/BaseMapperTest.php
```diff
@@ -0,0 +1,188 @@
...
+
+    private function bootstrapApplication($serviceManager, $configuration = []): Application
```
tomudding, line 82:
```php
    private function bootstrapApplication(
        ServiceManager $serviceManager,
        array $configuration = [],
    ): Application {
```

tomudding (commented):
A few tiny things that should be changed, after that it can be merged.

module/Frontpage/test/ControllerTest.php
```diff
@@ -8,6 +8,7 @@ class ControllerTest extends BaseControllerTest
 {
     public function testIndexActionCanBeAccessed(): void
     {
+//        $this->markTestSkipped('Decision/Mapper/Member::findBirthdayMembers is not supported by SQLite');
```
tomudding, line 11: Also, since you changed the mapper this can also be removed I suppose.

tomudding (approved):
Awesome work!

PR #1328, opened 2021-12-24
Fixed automatic controller tests

Koen1999:
Now it performs 355 assertions corresponding to approximately 150+ routes.

parsePart did not perform assertions as expected.

Still not all routes are being tested, however.

### DIFF
--- a/docker/web/development/php.ini
+++ b/docker/web/development/php.ini
@@ -1,6 +1,6 @@
 date.timezone = Europe/Amsterdam
 
-memory_limit = 256M
+memory_limit = -1
 max_input_time = 600
 max_execution_time = 300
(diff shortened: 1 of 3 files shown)


docker/web/development/php.ini
```diff
@@ -1,6 +1,6 @@
 date.timezone = Europe/Amsterdam
 
-memory_limit = 256M
+memory_limit = -1
```
tomudding, line 3: I know this is for the development image, but do we really need this to be "unlimited"?

Koen1999: These settings are taken from the PhpUnit documentation. These are the advised settings.

PR #1329, opened 2021-12-28
Company module overhaul

tomudding:
This PR contains everything related to the overhaul of the company module based on the work done by the SEP group last academic year.

All important files can be found in shared committee folder, otherwise you can always ask me for specific details.

**Work (to be) done:**
- [X] Add details for a company representative to the `Company` model
- [X] Introduce a `CompanyUser` which can be used by said representative to log in
- [X] `NewCompanyUser` is automatically created when creating a company
- [X] Rewrite the `AuthenticationService` to allow logging in with a company account
- [X] Rewrite the ACL and identity checks to support company accounts
- [X] Allow for optional contract number to be added to packages
- [X] Implement foundation for proposing updates for companies and jobs
- [X] Implement foundation for approving or rejecting update proposals
- [X] Allow `CompanyUser`s to propose (read: create) jobs
- [X] Allow `CompanyUser`s to propose updates for jobs
- [X] Allow `CompanyUser`s to delete jobs
- [ ] Allow `CompanyUser`s to propose updates for their company
- [x] Implement proposal approval queue for `company_admin`s
- [ ] When a company profile proposal has been approved, the `CompanyUser`s reference to the company should be correctly updated
- [ ] Replace current description editor with CKEditor v5 (#1335)
- [X] Allow jobs to be transferred from an expired job package to a non-expired job package
- [ ] TBA

**What is currently broken:**
- [ ] TBA


tomudding:
PHPCS and PHPStan momentarily fail as they have no knowledge about enums at this point in time (PHPStan has to be upgraded to 1.3.0, for PHPCS there is no known supported version).

Koen1999:
Since you upgraded some doctrine packages, you may also as well upgrade `doctrine/doctrine-laminas-hydrator`. I think it should not require any changes to our codebase.

tomudding:
I know that the mentioned PR in 060ec03 is already merged, however, it targets ORM v2.12 which will not be released for some time.

@Koen1999 Wouldn't it have been better to do the adjustments to the runtime configuration in a separate PR to try to limit this PR? Otherwise it will become a review hell. Or we need to restructure the way we are tackling this issue by making this the main PR and merging new features related to this into this PR (instead of doing everything in this PR).

Koen1999:
I think your last suggestion seems appropriate. I think there are little other issues that need to be addressed besides this, so I don't expect many other PRs. Otherwise this PR would require like x more rebases.

Koen1999:
@tomudding , question regarding this TODO:
`Allow jobs to be transferred from an expired job package to a non-expired job package`
Wouldn't it be a better idea to allow increasing the expiration date of the package to a date in the future?

module/Application/config/module.config.php
```diff
@@ -126,6 +127,11 @@
             'bootstrapElementError' => function () {
                 return new BootstrapElementError();
             },
+            'companyIdentity' => function (ContainerInterface $container) {
```
Koen1999, line 130: I also propose not creating a seperate helper for this. Instead, it may be a better idea to introduce an interface for the Identity which both the UserModel and the CompanyUserModel implement. Then you could just reuse the same view helper.

tomudding:
> @tomudding , question regarding this TODO: `Allow jobs to be transferred from an expired job package to a non-expired job package` Wouldn't it be a better idea to allow increasing the expiration date of the package to a date in the future?

We'd have to discuss that with Eline, because originally the idea was to have the packages be linked to specific contracts and extending an original contract would change the "contract number" hence result in a new package. Which would also give the company a clear history of what they have being doing in terms of contracts.

And to prevent a company from creating the same job a million times for different packages I thought it would be easier to just transfer them out.

tomudding (commented):
While I do think some of the changes here are particularly useful, I personally feel that some parts abstract too much away from what we are trying to achieve.

module/Application/config/module.config.php
tomudding: The problem is that the normal `identity` view helper only accepts a single `AuthenticationService` and our logic very clearly separates the two identities we have because they use different underlying models (`User` vs `CompanyUser`) and there is no way to separate the two without us having to store more information in our sessions.

Reworking the `AuthentiationService` to be one thing for both identities or changing our storage logic (how would you deal with the storage-level separation we currently have?) does not outweigh the benefits of not having a separate `companyIdentity()` view helper.

Additionally, we would have to change all our invocations of `identity()` to make sure we are working with a `User` and not a `CompanyUser`.

module/User/src/Module.php
```diff
@@ -356,6 +357,20 @@ public function getServiceConfig(): array
                     return $remote->getIpAddress();
                 },
                 'user_service_acl' => AclServiceFactory::class,
+                'user_service_storage' => function (ContainerInterface $container) {
+                    $translator = $container->get('translator');
+                    $storageConfig = $container->get('config')['storage'];
+                    $watermarkService = $container->get('user_service_watermark');
+
+                    return new FileStorage($translator, $storageConfig, $watermarkService);
+                },
```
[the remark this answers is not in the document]
Koen1999: It makes use of the watermark service, which also makes use of several other parts of the User module. Therefore, the User module is the most logical spot for these components.

module/User/src/Authentication/Service/LoginAttempt.php
```diff
@@ -11,9 +12,9 @@
     User as UserMapper,
 };
 use User\Model\{
+    User as UserModel,
```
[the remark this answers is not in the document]
Koen1999: Automatic refactoring didn't like the way imports are done sadly.

module/User/src/Service/User.php
```diff
@@ -426,7 +430,11 @@ public function companyLogin(array $data): ?CompanyUserModel
             return null;
         }
 
-        return $this->companyUserAuthService->getIdentity();
+        $identity = $this->userAuthService->getIdentity();
+        if ($identity instanceof CompanyUserModel) {
+            return $identity;
+        }
+        return null;
```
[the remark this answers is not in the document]
Koen1999: I will fix this bug, my bad!

module/User/src/Authorization/GenericAclService.php
```diff
@@ -108,37 +121,25 @@ public function getIdentityOrThrowException(): CompanyUserModel|UserModel|null
...
-            || $this->companyUserAuthService->hasIdentity()
-            || $this->apiAuthService->hasIdentity();
+        $identity = $this->getIdentity();
+
+        if ($identity instanceof CompanyUserModel) {
+            return $identity;
+        }
+
+        return null;
     }
```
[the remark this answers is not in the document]
Koen1999: Will do!

module/Application/config/module.config.php
tomudding: We could change the names, make `userIdentity()` an alias of `identity()` and change `companyIdentity()` to `companyUserIdentity()`. Just like the `get...` methods you introduced.

Koen1999: One way to make this easier, would be to have a superclass for User and CompanyUser. Then you could have one mapper for this. The main difficulty with this is that UserModels have a lidnr, and CompanyUserModels do not.

tomudding: Then I'd say that it is out of scope for this PR. There are many things that are more important and should be done instead.

If we make it work in the future that be great, although all the interdependencies may undo what you are trying to accomplish.

Koen1999:
Well, atleast 6cf1a2a613c27eb4027df7facd25293093731f0d made it work locally. Didn't realize the workflow had it's own seperate setup.

module/Application/view/partial/admin.phtml
```diff
@@ -17,7 +17,7 @@ $flash = $this->plugin('FlashMessenger');
                 <ul class="nav navbar-nav">
                     <?php if (null === $this->identity()): ?>
                         <li>
-                            <a href="<?= $this->url('user') ?>">
```
Koen1999, line 20: Another question. Why is this change necessary? Wouldn't it be a lot easier to just have implemented user_type member as a default in the route?

Koen1999:
A note on the conflicts in composer.json and the lock. You can safely override whatever is on master since the same patch is applied on this branch.

module/Application/view/partial/admin.phtml
tomudding: I think I did it like this as the admin page is specifically for members. However, in the very same commit I made `member` the default route of `user_type` (see `User/config/module.config.php`), so I probably forgot to change it back (or I thought being explicit was better).

tomudding:
Due to the sheer size of this PR it has become quite difficult to keep it up-to-date (50+ commits and 200+ files changed). Rebasing onto master is a frustrating process that can take hours to fix everything. Additionally, this means that I cannot really work on other issues.

Therefore, I will merge this PR as is, to make this possible without breaking anything I have disabled the creation of company accounts. This ensures that everything that is not yet finished is not accessible. For the normal interface nothing should have changed in a way that breaks something.

Psalm currently fails because it uses the wrong baseline.

PR #1331, opened 2021-12-28
Added referrer policy

### LINKED ISSUE #1299: Add Referer-Policy header
I propose setting it to either `strict-origin-when-cross-origin`.
This should be adjusted in the nginx.conf

Koen1999:
This fixes #1299


docker/nginx/nginx.conf
```diff
@@ -38,6 +38,12 @@ http {
...
     }
+    map $host $referrer_policy {
+        default "strict-origin-when-cross-origin";
+    }
+    map $host $permissions_policy {
+        default "accelerometer=("https://www.youtube-nocookie.com" "https://player.vimeo.com"), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), cross-origin-isolated=(), display-capture=(), document-domain=(), encrypted-media=("https://www.youtube-nocookie.com" "https://player.vimeo.com"), execution-while-not-rendered=(), execution-while-out-of-viewport=(), fullscreen=(self "https://www.youtube-nocookie.com" "https://player.vimeo.com"), geolocation=(), gyroscope=("https://www.youtube-nocookie.com" "https://player.vimeo.com"), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(), payment=(), picture-in-picture=("https://www.youtube-nocookie.com" "https://player.vimeo.com"), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(), xr-spatial-tracking=(), clipboard-write=(self), interest-cohort=()";
+    }
```
tomudding, line 47: NGINX does not accept this due to quotes in the `Permissions-Policy`. The outer qoutes should be changed to `'`, as `"` is required for the URLs.

tomudding (changes requested):

docker/nginx/nginx.conf
Koen1999: I fixed and tested it now

tomudding (approved):

PR #1333, opened 2022-01-02
Updated dependencies

Koen1999:
Most prominent update is node-sass v7

Also another note: doctrine recently received v5, but not all packages have compatible releases yet. So, it's not included.


tomudding (approved):
Looks good and PHPStan v1.3.0 is much appreciated for the company module PR!

PR #1336, opened 2022-01-17
Add infrastructure for automatic documentation generation

tomudding:
This supersedes #1246 and adds automatic generation (and deployment) of the documentation based on our PHP files.

I have moved the generated items to a separate branch `gh-pages`, this branch is an orphan and has no knowledge about any other commits/files. I have created the initial version of `gh-pages` myself, you can find it live at https://gewis.github.io/gewisweb/.

The GitHub Action borrows heavily from the Jekyll equivalent. It will automatically commit when the documentation has been updated, commit is authored (and committed) by the GitHub Actions bot, original commit author is added as a co-author.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ config/bash.env
 .idea/sonarlint-state.xml
 phpstan/phpstan-baseline-temp.neon
 phpstan/phpstan-baseline-pr.neon
+docs/_site/
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,47 @@
+name: Docs
+
+# Only run this action when pushing to master.
+on:
+    push:
+        branches:
+            - master
+
+jobs:
+    phpunit:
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -30,3 +30,4 @@ config/bash.env
 .idea/sonarlint-state.xml
 phpstan/phpstan-baseline-temp.neon
 phpstan/phpstan-baseline-pr.neon
+docs/_site/
```
Koen1999, line 33: If this is the output directory, and we ignore it. Then what is automatically committed?

tomudding: It is just there to ensure we do not accidentally push anything from the generated documents onto one of our normal branches. The directory itself is a separate git repository in the action itself.

One thing I may have forgotten is a `.gitignore` for the caches on the `gh-pages` branch, but we'll have to check that out after the action has run.

Koen1999 (approved):

PR #1340, opened 2022-01-29
Fix option calendar async ajax

### LINKED ISSUE #1339: Option calendar calendar not working in Chromium-based browsers
As reported by Leon, the calendar for the option calendar does not work in Chromium-based browsers (Chrome/Edge). The following error is reported:

> calendar.min.js:2 Uncaught TypeError: this.options.templates[this.options.view] is not a function
>     at j._render (calendar.min.js:2:7912)
>     at j.view (calendar.min.js:2:13307)
>     at new j (calendar.min.js:2:2932)
>     at n.fn.init.a.fn.calendar (calendar.min.js:2:22353)
>     at initCalendar ((index):4839:35)
>     at Object.success ((index):4822:9)
>     at i (jquery.min.js:2:27449)
>     at Object.fireWith [as resolveWith] (jquery.min.js:2:28213)
>     at y (jquery.min.js:4:22725)
>     at XMLHttpRequest.c (jquery.min.js:4:26929)

tomudding:
Fixes #1339.

Somewhere near the end of 2021, Chromium-based browsers appeared to have changed their behaviour regarding synchronous `XMLHttpRequests` (AJAX). This resulted in a race condition, where the templates would not be loaded before they were (supposed to be) used.

This commit slightly modifies the calendar library (as it has not been updated in a few years, I do not foresee any future problems), to make the AJAX requests asynchronous and adapt the code to the new behaviour.

I have removed the full version of the file, as it is not used. Non-minified version of the changed code can be found here: https://gist.github.com/tomudding/3f05ab7a162631633bb0050512028f34 (contains some comments).


Koen1999 (approved):
I think this is a good solution for now. But in the future, we may want to consider moving to a library that still receives support.

tomudding:
> I think this is a good solution for now. But in the future, we may want to consider moving to a library that still receives support.

I took some time last night to see what is possible, but there isn't really a lot out there. A lot of the libraries are quite complex or do not allow for easy modifications. The best contender is [fullcalendar.io](https://fullcalendar.io), very similar and actively supported (and as far as I can tell no jQuery dependency nor does it need underscorejs for templating), but that will take some time to integrate into the website.

Perhaps we can get Leon to work on that!

PR #1352, opened 2022-02-03
Replace Google Fonts with local versions

### LINKED ISSUE #1346: Move away from Google Fonts due to GDPR
Based on German court rulings, we may have to move away from using Google Fonts. While we can download Raleway and host it ourselves, it might be a good time to switch away from the custom font and just use whatever is available on the user's operating system.

Would also help us prepare for switching to bootstrap 5 later this year (but Tom, I didn't know we would do this? Correct, but I will be doing it during the summer).

tomudding:
Fixes #1346.


public/scss/_fonts.scss
```diff
@@ -0,0 +1,59 @@
+/* lato-italic - latin */
+@font-face {
```
Koen1999, line 2: I actually was going to provide feedback on all these font-face blocks.
I recommend adding the `font-display` property and setting it to `swap`.

https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

Koen1999: Essentially, this allows for rendering the webpage earlier (before fonts are retrieved). So, it makes the website look more responsive.

Koen1999: @tomudding

tomudding: I don't see why this would suddenly be a problem. When we used the Google CDNs this was never a problem and the fonts are still cached similarly (although we might want to change the NGINX config to actually enforce it).

The biggest blocking resource is the `gewis-theme.css`, as long as that is not loaded there won't be any fonts anyway.

tomudding: But please do make a PR and I'll merge it.

tomudding: @Koen1999 Perhaps we should preload the fonts too:

```html
<head>
    <!-- ... -->
    <link rel="preload" href="/raleway/raleway-v22-latin-500.woff2" as="font" type="font/woff2">
    <!-- ... -->
</head>
```

tomudding: To really prioritise loading them as early as possible.

PR #1354, opened 2022-02-03
Refactor `Course` view template

### LINKED ISSUE #1345: Hide `Course url: ` if not present in education exam/summary archive
If no course page is present, some info should be hidden.
For example, see https://gewis.nl/education/course/2IMN15
afbeelding

tomudding:
Fixes #1345.

Course URLs can now be `nullable` and are by default `null`. Hides the non-existent URL when it is non-existent. Restructured the whole template to have less repetition.

**Before:**
image

**After (custom uploads):**
image

### DIFF
--- a/module/Education/view/education/education/course.phtml
+++ b/module/Education/view/education/education/course.phtml
@@ -6,86 +6,98 @@ $this->headTitle($this->translate('Education'));
 ?>
 <section class="section">
     <div class="container">
-        <h1><?= $course->getCode() ?> - <?= $course->getName() ?></h1>
-
-        <ul>
-            <li><strong>Course url:</strong> <a href="<?= $course->getUrl() ?>">
-                    <?= $course->getUrl() ?>
-                </a></li>
-        </ul>
-
-        <?php $children = $course->getChildren() ?>
-        <?php if (!empty($children)): ?>
-            <h2><?= $this->translate('Course parts') ?></h2>
-            <ul>
-                <?php foreach ($course->getChildren() as $child): ?>
-                    <li><strong><?= $child->getCode() ?></strong> <?= $child->getName() ?></li>
-                <?php endforeach; ?>
-            </ul>
-        <?php endif; ?>
-
-        <h2><?= $this->translate('Exams and summaries') ?></h2>
-
-        <ul>
-            <?php foreach ($course->getChildren() as $child): ?>
-                <?php foreach ($child->getExams() as $exam): ?>
-                    <li><a href="<?= $this->url('education/course/download', [
-                            'code' => $exam->getCourse()->getCode(),
-                            'id' => $exam->getId()
-                        ]) ?>">
-                            <strong>(<?= $child->getCode() ?>)</strong>
-                            <?php if ($exam instanceof Summary): ?>
-                                <?= sprintf($this->translate('Summary by %s on %s (%s)'), $exam->getAuthor(), $exam->getDate()->format('Y-m-d'), $exam->getLanguage()) ?>
-                            <?php else: ?>
-                                <?php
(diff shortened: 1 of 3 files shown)


module/Education/view/education/education/course.phtml
```diff
@@ -6,86 +6,96 @@ $this->headTitle($this->translate('Education'));
...
+                    <p><?= $this->translate('This course does not have any subcourses.') ?></p>
+                <?php endif; ?>
+            </div>
+            <div class="col-md-8">
+                <h2><?= $this->translate('Exams and summaries') ?></h2>
+                <ul>
+                    <?php
+                    // Doctrine `Collections` are `Traversable`s, for which we can use `iterator_to_array` instead of
+                    // `array_push` or `foreach`. Appending also faster than unshifting.
+                    $courses = iterator_to_array($course->getChildren());
```
Koen1999, line 47: I propose changing this to the following to keep the original order while still putting the main course up front:
```
                    $courses = iterator_to_array($course->getChildren());
                    $coursesReverse = array_reverse($courses);
                    $coursesReverse[] = $course;
                    $courses = array_reverse($coursesReverse);
```

module/Education/view/education/education/course.phtml
```diff
@@ -6,86 +6,96 @@ $this->headTitle($this->translate('Education'));
...
+                </div>
+            <?php
+            endif;
+            ?>
+        </div>
+        <hr>
+        <div class="row">
+            <div class="col-md-4">
+                <h2><?= $this->translate('Subcourses') ?></h2>
+                <?php if (!$course->getChildren()->isEmpty()): ?>
```
Koen1999, line 27: I would put this if statement before opening `<div class="col-md-4">`. Otherwise the page still looks a bit odd if there are no subcourses. Then you could also completely leave out the else statement.

module/Education/view/education/education/course.phtml
```diff
@@ -6,86 +6,96 @@ $this->headTitle($this->translate('Education'));
 ?>
 <section class="section">
     <div class="container">
-        <h1><?= $course->getCode() ?> - <?= $course->getName() ?></h1>
-
-        <ul>
-            <li><strong>Course url:</strong> <a href="<?= $course->getUrl() ?>">
```
Koen1999, line 12: I would keep this strong. In the new version the URL appears out of nowhere. Context is useful for the user.

Koen1999 (changes requested):
I have some remarks. But mostly, the changes look good.

module/Education/view/education/education/course.phtml
tomudding: Mistake on my part, I had originally put the URL in the `<h1></h1>` but in the end that didn't look right. I'll add it back,.

module/Education/view/education/education/course.phtml
tomudding: Yeah I guess that for the subcourses that is preferable.

module/Education/view/education/education/course.phtml
tomudding: Is there really a need to keep the original ordering? As there isn't a strict order at this point, it is just the order in which the exams/summaries are added to the website. If someone decides to upload old exams, the order on the website as it currently is will also be wrong.

Koen1999: I think that front-end should adhere to the ordering provided by back-end.

tomudding: In that case I'll change it to `array_unshift`, that should be slightly faster than reversing the array twice.

Koen1999 (approved):
Looks good with the changes!

PR #1356, opened 2022-02-03
Fix being unable to set profile photo from photo viewer

### LINKED ISSUE #1338: Cannot set profile photo in the new photo viewer
As reported by Jelmer, there is currently no button to set a photo as your profile photo. We should add a button in the menu bar which can be used to set a photo (in which you have been tagged) as your profile photo.

tomudding:
Fixes #1338.

**After:**
image

### DIFF
--- a/module/Photo/src/Controller/PhotoController.php
+++ b/module/Photo/src/Controller/PhotoController.php
@@ -91,17 +91,16 @@ class PhotoController extends AbstractActionController
     /**
      * For setting a profile picture.
      */
-    public function setProfilePhotoAction(): Response
+    public function setProfilePhotoAction(): JsonModel|ViewModel
     {
-        $photoId = $this->params()->fromRoute('photo_id');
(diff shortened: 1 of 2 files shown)


module/Photo/src/Service/Photo.php
```diff
@@ -575,6 +575,7 @@ private function storeProfilePhoto(
      */
     public function setProfilePhoto(int $photoId): void
     {
+        // TODO: Add null check for `$photo`.
```
Koen1999, line 578: This TODO mark should either be addressed or removed if it is not necessary.

tomudding: It should definitely be done at some point, but I don't want to do it now. I'll make an issue for it and remove it from this PR.

Koen1999: Alright, sounds good too.

Koen1999 (approved):

PR #1358, opened 2022-02-04
Fix incorrect validation of activity and sign-up list date and times

### LINKED ISSUE #1337: Broken validation logic for activity start date and time
As reported by Noa, there appears to be an issue with the validation logic for the start date and time of an activity. It is possible to specify a start after the activity has ended.

tomudding:
Fixes #1337 and implements the previously unused validation for the sign-up list open and close date and time.

### DIFF
--- a/module/Activity/src/Form/Activity.php
+++ b/module/Activity/src/Form/Activity.php
@@ -238,10 +238,9 @@ class Activity extends LocalisableForm implements InputFilterProviderInterface
         array $context = [],
     ): bool {
         try {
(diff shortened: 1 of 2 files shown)


tomudding:
PHPStan issue is a result cache "problem" and can be safely ignored. We do not have to change the PHPStan action, as this is a very good thing to know.

Koen1999:
I suggest you update the baseline to disregard the change in count

Koen1999 (approved):

PR #1360, opened 2022-02-05
Replace old activity option calendar with new upgraded alternative

### LINKED ISSUE #1280: Hide approve button for already approved options in the option calendar slide-out
A button to approve an activity was added in a recent PR within the option calendar. This option should be hidden if the activity is already approved.

tomudding:
This replaces the old bootstrap calendar with FullCalendar, very similar but also not. Comes with a lot of changes:

- No longer dependant on the successful retrieval of the pre-planned activities in the Google Calendar, proposed options will always be shown.
- Better monthly view, it is easier to see the duration of activities which span multiple days.
- Improved day view for approval of proposed options. Furthermore, if an option has been approved, the approval button is no longer shown.
- Less dependencies; jQuery and underscore.js are no longer required.

Unfortunately, also a small caveat, while Bootstrap is supported as a theme, our v3.4 version is not compatible. Hence, internal styles and icons are used. This means a changed CSP (see NGINX config).

Is available on [test.gewis.nl](https://test.gewis.nl/activity/calendar).

This also fixes #1280.

**After (arrows are not shown because CSP was not updated):**
image

image

image

image

And if you aren't allow to approve an option proposal:
image


tomudding:
I am aware there is a slight spacing issue when you have multi day activities, however, I have not been able to find a solution to this.

Koen1999 (approved):
I think the changes look very good!

Just one note, the month/day view is not very intuitive. It's not very clear when you switch view when clicking an option. But I'm not sure how we could improve that.

Don't forget to recompile the stylesheets when merging.

PR #1362, opened 2022-02-05
Improved resource loading on frontpage

Koen1999:
This preloads all resources used on the front page by adding them to the head tag. This should increase page load time, especially on first visits. Rendering time is reduced. So nice enhancement :)

This PR was made to address concerns raised in #1352 .


tomudding (approved):
Looks good, I'll try to get a measurement before and after deploying this. That way we really know the impact of this change (specifically for first page load).

tomudding:
Here is a before:

image

Most of the bytes downloaded are the result of jQuery. Timings are influenced mostly by the images (specifically the photo of the week, which should be scaled down a bit).

tomudding:
And after:

image

Blocking time has indeed been reduced, but TTI is significantly more.

Koen1999:
Perhaps this is a recommendation we can work on in a future PR:
Screenshot_20220206_131137

PR #1363, opened 2022-02-06
Fixed permissions error on the member search page

### LINKED ISSUE #1316: Member search page is exposed
You will not be able to search any members, but the page is still exposed. Would be preferable to also have it behind an ACL check.

Koen1999:
This fixes #1316 .


tomudding (approved):
👍

PR #1364, opened 2022-02-06
Added some sanity checks on setProfilePhoto

### LINKED ISSUE #1357: No `null` check for photos in `setProfilePhoto()`
https://github.com/GEWIS/gewisweb/blob/9e5d2a28fd2c52ed9c3f03a03f4e42702bd72050/module/Photo/src/Service/Photo.php#L576-L589

No null checking on either `$photo` or the identity in the `AclService` (which can be `null`).

Koen1999:
This fixes #1357

### DIFF
--- a/module/Photo/src/Service/Photo.php
+++ b/module/Photo/src/Service/Photo.php
@@ -8,6 +8,7 @@ use DateTime;
 use Decision\Model\Member as MemberModel;
 use Decision\Service\Member as MemberService;
 use Doctrine\ORM\Exception\ORMException;
+use Doctrine\ORM\NoResultException;
 use Exception;
 use Laminas\Http\Response\Stream;
 use Laminas\I18n\Filter\Alnum;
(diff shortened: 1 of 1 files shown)


tomudding (approved):
I don't know if `NoResultException` is preferable, perhaps changing the method to return a boolean is better. Then we can properly show that it failed to the JavaScript (by returning slightly different JSON) and not have unnecessary errors in our logs.

But it works, so that is great.

Koen1999:
> I don't know if `NoResultException` is preferable, perhaps changing the method to return a boolean is better. Then we can properly show that it failed to the JavaScript (by returning slightly different JSON) and not have unnecessary errors in our logs.
> 
> But it works, so that is great.

This exception is automatically translated to a 404 in the controller. I think that status code is preferable.

tomudding:
> This exception is automatically translated to a 404 in the controller. I think that status code is preferable.

Forgot about that, although it is a bit unfortunate, as the JS expects a JSON result and not plain HTML.

PR #1365, opened 2022-02-06
Change default Glide encoding to `webp`

tomudding:
This changes the default encoding for Glide to `webp` which should provide better compression and thus file sizes in comparison to what we currently have.

Also halves the width (and correspondingly the height) of the Photo of the Week on the frontpage, we do not need large images in a rectangle which is less than 400px across.


module/Frontpage/view/frontpage/frontpage/home.phtml
```diff
@@ -206,7 +206,7 @@ use Activity\Model\Activity;
                                     'h' => round($wThumbnailWidth * $wAspectRatio)
                                 ];
 
-                                $wLargeWidth = $photoConfig['large_thumb_size']['width'];
```
Koen1999, line 209: Please do the same for the birthday photo.

Koen1999: And the company banner perhaps?

tomudding: I have changed the birthday photo. Company banners are pretty small, so I don't think it will be necessary for now.

tomudding:
@Koen1999 If you'd like you can commit your changes to the panorama to this branch.

Koen1999:
> @Koen1999 If you'd like you can commit your changes to the panorama to this branch.

Will do

Koen1999 (approved):

Koen1999:
Some quick and premture testing showed that the version of this PR is about 2 times as fast over a normal 3G connection than what we have currently deployed.

Koen1999:
@tomudding You can merge and deploy if you like

tomudding:
After:

image

Only things which really are in the way of fast load times are `gewis-theme.css` and `jQuery`. When we migrate to Bootstrap 5, we must remove all jQuery code and just use "vanilla" JS.

Koen1999:
Dropping jQuery and moving towards bootstrap 5 sounds like a good future plan!

PR #1370, opened 2022-02-12
Improved internationalization of the website

### LINKED ISSUE #1367: Language defined in layout.phtml in insensitive to translator locale
https://github.com/GEWIS/gewisweb/blob/2066bfa74a0188a2060e93114fa247ebca234ac2/module/Application/view/layout/layout.phtml#L3

This tag should vary depending on the locale set in the translator.

Moreover, we should also consider adding http-equiv meta tags for the other languages. But for this, we should implement some sort of referral mechanism for switching languages that does not just redirect to the homepage.

Please see: https://www.seobility.net/en/wiki/Declaring_language_in_HTML

Koen1999:
This fixes #1367 .

We now also automatically detect the preferred language from the Accept-Language header. If not present, we default to English.

### DIFF
--- a/module/Application/config/module.config.php
+++ b/module/Application/config/module.config.php
@@ -25,11 +25,14 @@ return [
             'lang' => [
                 'type' => Segment::class,
                 'options' => [
-                    'route' => '/lang/:lang/',
+                    'route' => '/lang/:lang',
                     'defaults' => [
                         'controller' => IndexController::class,
                         'action' => 'lang',
                         'lang' => 'nl',
+                    ],
+                    'constraints' => [
+                        'lang' => 'nl|en',
                     ],
                 ],
                 'priority' => 100,
--- a/module/Application/config/module.config.php
+++ b/module/Application/config/module.config.php
@@ -29,7 +29,7 @@ return [
                     'defaults' => [
                         'controller' => IndexController::class,
                         'action' => 'lang',
-                        'lang' => 'nl',
+                        'lang' => 'en',
                     ],
                     'constraints' => [
                         'lang' => 'nl|en',
--- a/module/Application/config/module.config.php
+++ b/module/Application/config/module.config.php
@@ -25,11 +25,12 @@ return [
             'lang' => [
                 'type' => Segment::class,
                 'options' => [
-                    'route' => '/lang/:lang',
+                    'route' => '/lang/:lang[/:href]',
                     'defaults' => [
                         'controller' => IndexController::class,
                         'action' => 'lang',
                         'lang' => 'en',
+                        'href' => '/',
                     ],
                     'constraints' => [
                         'lang' => 'nl|en',
--- a/module/Application/src/Controller/IndexController.php
+++ b/module/Application/src/Controller/IndexController.php
@@ -20,9 +20,9 @@ class IndexController extends AbstractActionController
             $session->lang = 'nl';
         }
 
-        $url = $this->getRequest()->getQuery('href');
+        $url = $this->params()->fromRoute('href');
         if (null !== $url) {
-            return $this->redirect()->toUrl($url);
+            return $this->redirect()->toUrl('/' . $url);
         }
 
         if (isset($_SERVER['HTTP_REFERER'])) {
--- a/module/Application/src/Controller/IndexController.php
+++ b/module/Application/src/Controller/IndexController.php
@@ -20,8 +20,8 @@ class IndexController extends AbstractActionController
             $session->lang = 'nl';
         }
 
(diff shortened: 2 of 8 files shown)


module/Application/config/module.config.php
```diff
@@ -25,12 +25,15 @@
             'lang' => [
                 'type' => Segment::class,
                 'options' => [
-                    'route' => '/lang/:lang/',
+                    'route' => '/lang/:lang',
                     'defaults' => [
                         'controller' => IndexController::class,
                         'action' => 'lang',
                         'lang' => 'nl',
```
tomudding, line 32: Perhaps also change this default?

module/Application/view/partial/main-nav.phtml
```diff
@@ -39,14 +39,18 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
                     <ul class="dropdown-menu language <?= ($lang == 'en') ? 'reverse' : '' ?> ">
                         <li>
                             <a class="<?= ($lang == 'en') ? 'disabled' : '' ?>"
-                               href="<?= $this->url('lang', ['lang' => 'en']) ?>">
+                               href="<?= $this->url('lang', ['lang' => 'en']) ?>"
+                               hreflang='en'
+                            >
```
tomudding, line 44: I think this is incorrect usage of `hreflang`, it is supposed to go on a `link` and not an `a` (and it should not be used on pages which redirect).

module/Application/view/partial/main-nav.phtml
```diff
@@ -39,14 +39,18 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
...
                                 <span class="flag-icon flag-icon-en"
                                       style="background-image: url('<?= $this->basePath() ?>/img/en.svg');"></span>English
                             </a>
                         </li>
                         <li>
                             <a class="<?= ($lang == 'nl') ? 'disabled' : '' ?>"
-                               href="<?= $this->url('lang', ['lang' => 'nl']) ?>">
+                               href="<?= $this->url('lang', ['lang' => 'nl']) ?>"
+                               hreflang='nl'
+                            >
```
tomudding, line 53: Same here for the `hreflang`.

module/Application/src/Controller/IndexController.php
```diff
@@ -20,6 +20,11 @@ public function langAction(): Response
             $session->lang = 'nl';
         }
 
+        $url = $this->getRequest()->getQuery()['href'];
+        if (isset($url)) {
```
tomudding, line 24: If `href` is not a key this will still throw an error. Better would be `->getQuery('href')` and then checking `null` (I think).

And a question, where does `href` come from? I don't see it in the code.

tomudding (commented):
The tests fail because you have not yet removed the trailing `/` in the tests (which was changed in the router).

module/Application/view/partial/main-nav.phtml
Koen1999: It is proper usage according to this page: https://www.w3schools.com/TAGs/att_a_hreflang.asp

tomudding: Interesting, didn't know that. All good then!

module/Application/view/layout/layout.phtml
```diff
@@ -9,54 +9,64 @@
...
-    <?= $this->headLink([
+    <?php $this->headLink([
         'as' => 'image', 'rel' => 'preload', 'type' => 'image/svg+xml', 'crossorigin' => 'anonymous',
         'href' => '/img/nl.svg'
     ]); ?>
-    <?= $this->headLink([
+    <?php $this->headLink([
         'as' => 'image', 'rel' => 'preload', 'type' => 'image/svg+xml', 'crossorigin' => 'anonymous',
         'href' => '/img/en.svg'
     ]); ?>
```
tomudding, line 66: Not an issue, but could probably do with one `<?php ?>` for all of them (and possibly chain them too.)

tomudding (approved):

rinkp:
I'm currently experiencing an issue with this. If you switch from Dutch to English or the other way around, you will get a 500 status code. After that, the language does work.
My browsers are set to English by default in case that is relevant. Both Chrome and Firefox do this.
@Gijsdeman is also having this issue

tomudding:
> I'm currently experiencing an issue with this. If you switch from Dutch to English or the other way around, you will get a 500 status code. After that, the language does work. My browsers are set to English by default in case that is relevant. Both Chrome and Firefox do this. @Gijsdeman is also having this issue

This has been fixed.

PR #1373, opened 2022-02-13
Fix authorization issues

### LINKED ISSUE #1302: Improve error handling when trying to authorize the same person for the same meeting again.
Currently, if this happens we see a foreign key violation, which turns into a 500 for the user. This should be improved. Additionally, while I have not been able to confirm this, apparently when authorising someone (no duplicates etc.) an error is also shown.

### LINKED ISSUE #1369: Extra authorizations only limited by frontend, no check on backend
Not allowing people to authorize more than one person is limited to the frontend. You can still submit the form (by going back after first authorization or by looking at this repository for the code). To prevent this, there should also be a check on the backend and since authorizations are very 'special' we should add a CSRF-check.

tomudding:
This fixes #1369 (and as a result also fixes #1302).

Technically there is still a race-condition where you can authorize two (or more) different members. Changing the unique columns in the table is not really helpful, if you remove `recipient` you cannot have a revoked and actual authorization. If you add `revoked` (with `recipient` removed) you can revoke only one authorization.

The latter would work if `$revoked = true` is `NULL` in the database (`NULL` does not care about uniqueness). So we'd probably want to change it to `$isValid = true` (by default) and then set it to `null` when it is no longer valid.

@Koen1999 Let me know what you think about the proposed solution for the race-condition. Though it will not be possible to do it at this point in time, as we have a duplicate record in the database (so we need to discuss with the board what we can do).

### DIFF
--- a/module/Decision/src/Controller/DecisionController.php
+++ b/module/Decision/src/Controller/DecisionController.php
@@ -175,7 +175,7 @@ class DecisionController extends AbstractActionController
             $form->setData($request->getPost()->toArray());
 
             if ($form->isValid()) {
-                if ($this->decisionService->createAuthorization($form->getData())) {
+                if (false !== ($authorization = $this->decisionService->createAuthorization($form->getData()))) {
                     return new ViewModel(
                         [
                             'meeting' => $meeting,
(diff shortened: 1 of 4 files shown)


Koen1999:
I think this looks fine for our purposes.
Another solution could be to have a seperate table for revoked auths, then you can let the database make sure there is no race condition.

Koen1999 (approved):

PR #1376, opened 2022-02-20
Fix issues with photo tagging

### LINKED ISSUE #1315: Tags in photos do not persists locally + comma for first entry
Currently, when tagging someone in a photo the entry is not persisted locally in the DOM. If you close the photo viewer the tag will not be visible. Furthermore, the initial tag shows a stray comma.

### LINKED ISSUE #1374: bug/feature for Tagging members in Pictures
With the beta viewer tagging has become more tiresome again. If possible could (some) of the following suggestions maybe be implemented?

- If there is only a single suggestion it should be auto togged OR if you type a full name it should be auto tagged
- If you have just tagged someone, or have the tag some text box selected it should be auto focused when switching image.

<3

tomudding:
Fixes #1315. Partially but sufficiently fixes #1374: fixes the auto tag on full name functionality and not being able to press `z`.

Automatically focussing on the input element after changing the slide seems practically impossible. Adding `autofocus` does not work, as the page will have a lot of them (one in each `figtags` if allowed) and it will never focus on the correct one.

Using the `focus` event we can add some focus indicator, however, if we then use `blur` event to remove it when the user leaves the input, the input will never re-focus on slide change (as focus status is lost before re-focus code is activated).

Making focussing easier with `tabindex` does not work either, as other elements on the page will take priority (event with `tabindex="1"`).

What does work is `tab`ing through the photo viewer (first the menu, then the arrows, and finally the tag search). Then, when changing slides, pressing `tab` will automatically focus on the tag search again (tested in Chrome and Firefox).


module/Photo/view/partial/tags.phtml
```diff
@@ -7,8 +7,10 @@ $this->scriptUrl()->requireUrl('photo/photo/tag/remove', ['photo_id', 'lidnr'])
     $tagCount = $this->photo->getTags()->count();
     $i = 1;
     ?>
-    <?= (0 !== $tagCount) ? $this->translate('In this photo:') : '' ?>
-    <div class="tagList tagList-<?= $this->photo->getId() ?>">
+    <?php if (0 !== $tagCount): ?>
+        <span class="tag-title"><?= $this->translate('In this photo:') ?></span>
+    <?php endif; ?>
```
Koen1999, line 12: Would it be an idea to add an alternate text in case no tags are present? Something like `No one has been tagged in this photo yet. Tag someone now!`

public/js/autocomplete.min.js
```diff
@@ -8,4 +8,4 @@
...
...elected),e.selectedIndex=-1,e.element.value=e.currentValue,void e.findBestHint()):void e.adjustScroll(e.selectedIndex-1)},moveDown:function(){this.selectedIndex!==this.suggestions.length-1&&this.adjustScroll(this.selectedIndex+1)},adjustScroll:function(e){var t=this,n=t.activate(e),s=n.offsetTop,i=t.suggestionsContainer.scrollTop,o=n.offsetHeight,n=i+t.options.maxHeight-o;s<i?t.suggestionsContainer.scrollTop=s:n<s&&(t.suggestionsContainer.scrollTop=s-t.options.maxHeight+o),t.element.value=t.getValue(t.suggestions[e].value),t.signalHint(null)},onSelect:function(e){var t=this,n=t.options.onSelect,e=t.suggestions[e];t.currentValue=t.getValue(e.value),t.currentValue!==t.element.value&&(t.element.value=t.currentValue),t.signalHint(null),t.suggestions=[],t.selection=e,d.isFunction(n)&&n.call(t.element,e)},getValue:function(e){var t,n=this.options.delimiter;return n?(n=(t=this.currentValue).split(n),t.substr(0,t.length-n[n.length-1].length)+e):e},dispose:function(){var e=this.suggestionsContainer;this.disableKillerFn(),e.parentNode&&e.parentNode.removeChild(e)}},"function"==typeof define&&"object"==typeof define.amd&&define.amd&&define(function(){return l}),e.VanillaAutocomplete=l}(this);
```
Koen1999, line 11: What was changed here?

tomudding: The default value for auto tagging on a complete match, from `false` to `true`, or more specifically from `!1` to `!0`.

The reason why I changed it in the file itself is because it was like that before I switched it from jQuery to vanilla JavaScript. So to make sure there is at least some parity between the two and do not have to change it everywhere we use the autocomplete.

module/Photo/view/partial/tags.phtml
tomudding: Good idea, we just have to make sure we do not `remove()` the element, as we cannot add translations in the JS file which handles the removal of tags.

I'll add it when I have some time this weekend.

tomudding:
This also fixes the issue where pressing on the enter key would submit a form. The form is not needed and has therefore been removed. Additionally, I enabled the option to directly select the first option from the tag list, this should help keyboard users with navigation through the tag list.

Koen1999 (approved):
No further comments!

PR #1388, opened 2022-03-18
Don't try to show album cover if it does not exist

### LINKED ISSUE #1381: Album overview not accessible if album does not have cover
If there exists an album (in the current year) which does not have a cover the page becomes inaccessible.

tomudding:
Fixes #1381.


Koen1999 (changes requested):
I think we should hide the complete HTML element in this case instead of only leaving the src attribute empty

tomudding:
@Koen1999 I have changed the check to hide the `img` tag when the album cover does not exist.

Koen1999 (approved):

PR #1389, opened 2022-03-18
Fix ordering of activity archive

### LINKED ISSUE #1378: Change ordering activity archive to descending by year
Currently, the photo archive is ordered with the most recent on the left, and the activity archive is ordered the other way. Preferably, this would go both in the same order for user-friendliness

tomudding:
Order of the navpills is not the same as in the photo module, which is more logical (decreasing order). Also displays the activities in reverse order (most recent is first).

Also some small quality of life improvements, consisting of small messages if there are no activities/albums in a particular year.

Fixes #1378.


Koen1999 (approved):
Changes look good to me.

PR #1390, opened 2022-03-18
Scroll to last viewed photo and add button from member to real album in photo viewer

### LINKED ISSUE #1379: Add option to go to original album from photos in a user's album
With the new photo viewer, the ability to move from a photo within a user's album (i.e. photos in which the user is tagged) to the corresponding album has been lost. Maybe we can look into making this an option again

tomudding:
Based on #1376, which should be merged first.

Based on the feedback we got, I have added some functionality that will automatically scroll to the last viewed photo when you close the photo viewer (assuming it is not within the viewport). Additionally, this fixes #1379, when viewing a member album there will be an additional button in the photo viewer UI which will open the original album in a new tab.


Koen1999 (approved):
No comments!

PR #1391, opened 2022-03-19
Load tags and voted status dynamically in the photo viewer

### LINKED ISSUE #1384: Improve loading time of large albums
Large albums (300+ photos) take a long time to load due to the amount of queries performed. We should try to find a way to optimize the queries or to dynamically load more photos while scrolling through the album (but then we also have to make sure that it works while having the photo viewer open).

_Originally reported by Bouke._

### LINKED ISSUE #1396: `Set as profile picture!` is shown on pictures you are not tagged in
The new photo viewer shows the "Set as profile picture!" button always, even when not showing your own photos

### LINKED ISSUE #1398: Sharing a shared picture just appends new image to URL
_Originally reported by Samuel._

tomudding:
Based on #1390.

Resolves #1384. Loading times of large albums were dominated by the time it took Doctrine to load information about tags. Unfortunately, we cannot tell Doctrine to only selectively load certain information from the tags (like `lidnr` and `fullName`). To make this possible, I have created a small API-endpoint which can provide that information for any particular image.

This means that there is no upfront cost of loading tags and any actual retrievals will be less costly (as we do not need to load complete entities).

~~Perhaps we should also do the same for votes?~~

**Edit:** this now loads both the tags and whether someone has voted for a photo in one request.

Fixes #1396 and fixes #1398.


public/js/photo.js
```diff
@@ -44,10 +44,32 @@ let Photo = {
...
+                            realTag.parentElement.parentElement.querySelector('.tag-title').classList.add('hidden');
+                            realTag.parentElement.parentElement.querySelector('.no-tag-title').classList.remove('hidden');
+                        } else {
+                            // If there are 1 or more elements left, check if the tag was the last element and if so,
+                            // fix the spacer(s).
+                            if (null === realTag.nextElementSibling) {
+                                realTag.previousElementSibling.querySelector('.tag-spacer').remove();
+
+                                if (2 < realTag.parentElement.childElementCount) {
+                                    // TODO: Cannot be translated.
```
Koen1999, line 62: I think you can assume it is defined. We pass it to most views.

tomudding:
Just a little update on this; it is now rebased to remove some of the old (already merged) commits. I will implement the dynamic loading of votes later today (together with the dynamic tags that should make the album pages fast once again).

module/Photo/src/Mapper/Tag.php
```diff
@@ -44,6 +45,32 @@ public function getTagsByLidnr(int $lidnr): array
...
+     * @return array
+     */
+    public function getTagsByPhoto(int $photoId): array
+    {
+        $rsm = new ResultSetMapping();
+        $rsm->addScalarResult('id', 'id', 'integer')
+            ->addScalarResult('lidnr', 'lidnr', 'integer')
+            ->addScalarResult('fullName', 'fullName');
+
+        $sql = <<<QUERY
```
Koen1999, line 61: Why don't we use the query builder here?

Koen1999 (changes requested):

module/Photo/src/Mapper/Tag.php
tomudding: One of the main reasons is speed, loading the complete `Member` entity for just full name (for which we also have to call the function) brings a lot of overhead from Doctrine. So if we directly want to get the full name from the query (and preventing us from loading the whole entity) we would have to use `CONCAT_WS`. Unfortunately, this is not supported in the query builder/DQL and hence I used a native query here.

tomudding:
I am merging this now as there is an issue present on the website that is fixed in this PR. I have extensively tested the changes in this PR on the test website and locally, and I have not been able to find anything that prevents us from merging this.

If there do appear to be issues later one, please open an issue.

PR #1394, opened 2022-03-21
Show actual error page instead of blank page for unrecoverable PHP errors

### LINKED ISSUE #1380: Add proper error page for hard 5xx errors
Currently, if for whatever reason something critical fails (e.g., no connection with MariaDB) the website will return an empty page (5xx). It would be nice to change that, though I am not sure it is actually possible.

tomudding:
Closes #1380.

To make sure the user has a better experience when an unrecoverable (and thus fatal) `Throwable` is not caught I have introduced a global error page.

This makes sure that users actually get to see that something went wrong, instead of only getting a blank page. The page will only be shown for `Throwable`s that could not be caught by Laminas MVC (e.g., `PDOException`s).

Additionally, I added the page (and NGINX configurations) that I use for maintenance periods in which NGINX is still online, but the actual application not.

500 page:
image


Koen1999 (approved):
Looks good!

PXJesse (commented):
There's a spelling mistake in the error 500 message, it should be "bear with us" rather than "bare" (source: https://www.grammarly.com/blog/bear-with-me/)

tomudding:
> There's a spelling mistake in the error 500 message, it should be "bear with us" rather than "bare" (source: https://www.grammarly.com/blog/bear-with-me/)

Well, that is a bit awkward, but let's look at the bright side, no one has seen the page yet! (I'll update it later this)

PR #1400, opened 2022-03-28
Improve confirmation message for authorizations when recipient has max authorizations

### LINKED ISSUE #1399: It is possible for a member to receive more than 2 authorizations
It is possible for a single member to receive more than 2 authorizations, while according to the bylaws this should be at most 2. Moreover, all authorizations are marked as valid in the frontend

tomudding:
This improves the confirmation message a member receives when they try to authorize another member who already has two authorizations.

It has been made more clear that the recipient has the power (through the eternal memorandum) to choose which authorizations they use if they were to receive more than two authorizations.

Closes #1399.


rinkp (approved):
Text sounds fine to me; I made a minor change in the text, but I'm not sure whether you want to merge this including translations

tomudding:
We generally don't merge translations and stylesheet updates through PRs, as it gets quite hard to review them. But in this case it does not really matter 

Thank you for the better message!

PR #1401, opened 2022-03-30
Sort translation files before computing them

rinkp:
Should prevent quite a few changed lines if you are using different computers to create the files

### DIFF
--- a/translate-helper
+++ b/translate-helper
@@ -3,7 +3,7 @@ dir="$PWD"
 cd `dirname $0`
 
 # parse php files
-find ./module -iname "*.phtml" | xargs xgettext \
+find ./module -iname "*.phtml" -print0 | sort -z | xargs -r0 xgettext \
     --language=PHP \
     --keyword=translate \
     --output=./module/Application/language/gewisweb.pot \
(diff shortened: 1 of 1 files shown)


tomudding (commented):
It does reduce the number of lines changed when switching between machines. However, it appears to change more when you are still on the same machine.

Without changing anything in the source files, repeatedly calling `./translate-helper` (on the same machine) will change the order of obsolete translations. So we might want to remove those before using this.

**Edit:** Not only on the same machine, it just changes the ordering on every new invocation of the helper.

rinkp:
I got lucky when I tested the number of lines changed a few times I think. 

- What is the reason we keep obsolete translations at all? I think it is rather unlikely to introduce the same string a second time after having deleted it once. If there is no use, we can delete it.

- I noticed after running it quite a few times that there are certain orderings that are preferred for the not-used strings, but I haven't found out why it still changes each time.

tomudding:
> I got lucky when I tested the number of lines changed a few times I think.
> 
>     * What is the reason we keep obsolete translations at all? I think it is rather unlikely to introduce the same string a second time after having deleted it once. If there is no use, we can delete it.
I didn't even know old translations were not removed from the PO files.

The easiest way to resolve this is probably

```bash
msgattrib --no-obsolete -o en.po en.po
msgattrib --no-obsolete -o nl.po nl.po
```

After the `msgmerge` instead of having to remember to do it in POEdit.

Koen1999:
I think there is no issue with keeping obsolete translations. It potentially reduces the translation effort for future changes to the website.

tomudding (approved):
This should prevent a lot of unnecessary changes and should make it easier to actually review translations changes.

Though I want to point out that we should continue with only compiling the translation files on master and not in the PR itself.

tomudding:
I will update the translation files on master after merging this.

PR #1406, opened 2022-04-11
Updated dependencies

Koen1999:


tomudding:
Thank you for splitting the different updates across several commits. It will take me a bit longer to completely check this out as the FontAwesome changes are pretty significant.

tomudding (commented):
Some of the icons may look different but I have not seen any problems with migrating to V6 of FontAwesome. I do however feel like that we should be a bit careful since we are still working with Bootstrap 3 (regarding the shims).

Everything else looks good too!

tomudding (approved):
I clearly selected approve...

PR #1413, opened 2022-04-22
Add more claims to the `ApiApp` JWT

### LINKED ISSUE #1403: Request for more user details in JWT token
Dear Web,

As you might know we are working very hard on SudoSOS. We are at the time in development where we are trying to merge it into the current GEWIS ecosystem. One of the main use cases is inactive members who only have a Web account logging into SudoSOS.

SuSOS solves this problem by direct connection to the DB, we would prefer to do it in a nicer way. Would it be possible to get more information from the WEB JWT token to be able to create an account? We only need membership number, first- last name and an e-mail to so.

If you disagree with anything, let me know, I could always use your input.

tomudding:
This adds more claims to the JWT that were requested for SudoSOS, they should also provide some more functionality to other `ApiApp`s.

The claims added:
- `given_name`      => given name or first name
- `family_name`     => given surname or last name
- `email`           => email address
- `membership_type` => type of membership

This closes #1403.

**This is a draft pending a discussion with the IO/privacy responsibles regarding the implementation and its implications.**

### DIFF
--- a/module/User/src/Service/ApiApp.php
+++ b/module/User/src/Service/ApiApp.php
@@ -36,10 +36,15 @@ class ApiApp
         UserModel $user,
     ): string {
         $app = $this->getMapper()->findByAppId($appId);
+        $member = $user->getMember();
 
         $token = [
             'iss' => 'https://gewis.nl/',
-            'lidnr' => $user->getLidnr(),
+            'lidnr' => $member->getLidnr(),
+            'given_name' => $member->getFirstName(),
+            'family_name' => $member->getLastName(),
+            'email' => $member->getEmail(),
+            'membership_type' => $member->getType(),
             'exp' => (new DateTime('+5 min'))->getTimestamp(),
             'iat' => (new DateTime())->getTimestamp(),
             'nonce' => bin2hex(openssl_random_pseudo_bytes(16)),


tomudding:
Copied from de89908f40cea68330c1f86509687dddbb1e4221:

Switch from HS256 to HS512 for signing the authentication JWTs. Recommended key size is 128 bytes (should not be higher) and may never be less than 64 bytes. Should be generated by a CSPRNG, for example:

`openssl rand -base64 128`

Existing keys less than 64 bytes or which are not properly generated will be replaced.

module/User/src/Controller/ApiAuthenticationController.php
```diff
@@ -23,31 +29,74 @@ class ApiAuthenticationController extends AbstractActionController
...
     }
 
-    public function tokenAction(): Response
+    public function tokenAction(): Response|ViewModel
     {
         $appId = $this->params()->fromRoute('appId');
+        $app = $this->apiAppMapper->findByAppId($appId);
+
+        if (null === $app) {
+            return $this->notFoundAction();
```
Koen1999, line 61: It's good practice to first retrieve the identity such that some sort of ACL is performed before revealing information about our database contents.

module/User/src/Model/ApiApp.php
```diff
@@ -41,6 +42,24 @@ class ApiApp
...
+     * The claims that will be present in the JWT. If `null` only the lidnr will be passed along.
+     */
+    #[Column(
+        type: "simple_array",
+        nullable: true,
+        insertable: false,
+        updatable: false,
+        enumType: JWTClaims::class,
+    )]
+    protected array $claims;
```
Koen1999, line 61: Add these to phpstan errors to the baseline

module/User/src/Model/Enums/JWTClaims.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace User\Model\Enums;
+
+use Decision\Model\Member as MemberModel;
+use Laminas\Mvc\I18n\Translator;
+
+/**
```
Koen1999, line 8: Fix this PHP_CS Warning. The enum related ones can be ignored as long as there is no support.

Koen1999 (changes requested):
This PR looks good. But the PhpStan issues need to be solved or suppressed before merging.

module/User/src/Controller/ApiAuthenticationController.php
tomudding: I'll change it but I don't think it is super common to have in external authorization applications. For example, Apple, Facebook, and Twitter will show an error page if the token (app id in our case) is not valid without having to sign in.

module/User/src/Model/Enums/JWTClaims.php
tomudding: I believe this is still an enum issue, it does not understand that the docblock belong to `enum` and not the file. Maybe we should just use the dev version of PHP_CS?

PR #1418, opened 2022-04-28
Fix being unable to sign off external participant

### LINKED ISSUE #1416: Cannot remove external members from subscription list
Specifically the case for [Deloitte Lunch Lecture](https://gewis.nl/admin/activity/participants/5777/1545). When trying to delete an external member, 404 is given for the `externalSignoff` endpoint.

tomudding:
I changed the default repository of the `Signup` mapper at some point to `UserSignup`, but did not account for `ExternalSignup`. Hence, when trying to sign off an external participant from an activity, the `Signup` cannot be found.

Fixes #1416.

### DIFF
--- a/module/Activity/src/Controller/AdminController.php
+++ b/module/Activity/src/Controller/AdminController.php
@@ -389,10 +389,10 @@ class AdminController extends AbstractActionController
         );
     }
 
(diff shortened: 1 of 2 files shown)


PXJesse (approved):
This seems like a good solution to the bug

PR #1424, opened 2022-05-01
Move photos of the week to a `VirtualAlbum` (on a per year basis)

### LINKED ISSUE #1407: Make all "Photos of the Week" a `VirtualAlbum` and serve with the photo viewer
The page which shows all "Photos of the Week" is incredibly slow as it tries to load all photos at once, additionally you cannot really interact with these photos.

I suggest that we all these photos part of a `VirtualAlbum`, that way we can reuse the photo viewer for that page. Going back to the original album from a photo will still be possible by reusing the functionality that exists in the photo viewer for `MemberAlbum`s.

tomudding:
~~**Based on #1391 and should be rebased before merging.**~~

Currently, the Photos of the Week page is completely broken. It didn't use Glide (I don't know why I hadn't changed it) and its layout was no longer properly determined by Masonry. While this could easily be fixed, I thought it would be nicer to completely redo the page.

So that is this.

The Photos of the Week page still shows the current Photo of the Week, but no longer all other Photos of the Week. You can now view the association years (similar to what you see on the normal album overview page) with a thumbnail of the most recent Photo of the Week from that year. Clicking on such an album leads you to the photo viewer.

This works because we now have a new `VirtualAlbum`: `WeeklyAlbum`. This album is used to hold all the Photos of the Week in a certain year, it has an extra attributes `$dates` that holds an array of the `DateTime`s for each Photo of the Week (to prevent having to request it twice).

Because the photos are now in a `VirtualAlbum`, we can reuse the existing photo viewer to show the album. So each year of Photos of the Week can be viewed like any other album.

In the photo viewer, if a photo is a Photo of the Week a small box (like the tag list box) is shown to let the user know in which week of what year the photo was the Photo of the Week.

**Photos of the Week:**
image

**Photos of the Week album**
image

**Photos of the Week album in photo viewer**
image

**Normal album with a Photo of the Week in photo viewer**
image

Closes #1407.

### DIFF
--- a/config/autoload/global.php
+++ b/config/autoload/global.php
@@ -12,6 +12,10 @@
  * file.
  */
 
+use Application\Extensions\Doctrine\{
+    Rand,
+    Year,
+};
 use Laminas\Session\Storage\SessionArrayStorage;
 use Laminas\Session\Validator\HttpUserAgent;
 
@@ -153,7 +157,8 @@ return [
         'configuration' => [
             'orm_default' => [
                 'numeric_functions' => [
-                    'RAND' => 'Application\Extensions\Doctrine\Rand',
+                    'RAND' => Rand::class,
+                    'YEAR' => Year::class,
                 ],
             ],
         ],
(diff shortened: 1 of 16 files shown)


tomudding:
Based on the association years might be more useful.

Koen1999 (approved):
These changes look amazing! Should give a good performance boost.

PR #1454, opened 2022-06-03
Automatically calculate aspect ratio on persist

### LINKED ISSUE #1449: Determine photo aspect ratio while generating thumbnail instead of on-the-fly
The last performance bottleneck for very large albums on the website is the on-the-fly aspect ratio calculation by looking up the pictures.

This will require "manual" calculation of all the old aspect ratios, but this should not be a big problem.

tomudding:
This automatically calculates the aspect ratio of a `Photo` when the entity is persisted (only on the first persist). There is no other way to do this without involving the EntityManager, as Doctrine is strictly follows the data mapper pattern and not the active record pattern. Hence, updating the aspect ratio from within the entity is not possible.

This does still require the manual calculation of the aspect ratios of all old photos.

Fixes #1449.

### DIFF
--- a/module/Photo/src/Model/Photo.php
+++ b/module/Photo/src/Model/Photo.php
@@ -5,7 +5,6 @@ namespace Photo\Model;
 use DateTime;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\ORM\Mapping\{
-    ClassMetadataInfo,
     Column,
     Entity,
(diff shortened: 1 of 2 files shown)


tomudding:
I have "manually" calculated the aspect ratios of the old photos. I am merging this to ensure that we do not have to do it again in the future. I have tested this with a lot of photos.

PR #1456, opened 2022-06-06
Updated policies to make the list complete

### LINKED ISSUE #1438: List all policies on members' page or only link to the public archive
This is something that came up during the 188^th GMM; not all policies are listed on the members' page which can be confusing (if you do not know there are more policies). Therefore, it was suggested that we either include all policies or directly link to the folder in the public archive which contains all the policies.

rinkp:
Upon merging, fixes #1438


tomudding (approved):
Please try to not compile the translation files in the PR itself. We compile them before a release (the same should happen with the stylesheets) to prevent merge conflicts if we have multiple PRs/branches with changes that do not share the same "parent" commit.

PR #1462, opened 2022-06-09
Updated dependencies and dropped dependency on laminas/log

Koen1999:


module/Decision/src/Model/SubDecision/Board/Release.php
```diff
@@ -53,7 +54,7 @@ class Release extends SubDecision
     /**
      * Date of the discharge.
      *
-     * @ORM\Column(type="date")
+     * @Column(type="date")
```
tomudding, line 57: These are leftover from the switch to native attributes instead of the Doctrine annotations, they can and should be removed (which means that we do not need the import). The same applies to the `Album` entity.

tomudding (commented):
Updating `laminas/laminas-inputfilter` to v2.16.0 seems to have broken some part of the type checks by PHPStan. It appears that they are now using custom Psalm types, which PHPStan should understand but clearly doesn't.

We can try changing it to `InputFilterSpecification`, but that is still Psalm specific and therefore has to be in a DocBlock as PHP won't accept it.

module/Decision/src/Model/SubDecision/Board/Release.php
Koen1999: I fixed that

tomudding (approved):

PR #1466, opened 2022-06-11
Added stronger typing using ::class for services

Koen1999:
@tomudding , I now only did this for the translator. But I think this is a direction we should take for all services.
What do you think?

### DIFF
--- a/module/Frontpage/src/Module.php
+++ b/module/Frontpage/src/Module.php
@@ -3,6 +3,7 @@
 namespace Frontpage;
 
 use Doctrine\Laminas\Hydrator\DoctrineObject;
+use Laminas\Mvc\I18n\Translator as MvcTranslator;
 use Frontpage\Form\{
     NewsItem as NewsItemForm,
     Page as PageForm,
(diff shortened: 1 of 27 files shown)


module/Frontpage/src/Module.php
```diff
@@ -48,8 +49,7 @@ public function getServiceConfig(): array
         return [
             'factories' => [
                 'frontpage_service_frontpage' => function (ContainerInterface $container) {
-                    $translator = $container->get('translator');
-                    $aclService = $container->get('decision_service_acl');
```
tomudding, line 52: Should not be removed?

tomudding: The `$aclService` I mean.

tomudding (commented):
Yes, this will make working on the website a lot better in the future. I think it is a good idea to only do this with the `MvcTranslator` to see how it plays out (e.g., we might want to look into potential performance impacts - positive and negative).

module/Frontpage/src/Module.php
Koen1999: I'm not sure why this went wrong while rebasing...Anyway, fixed it

tomudding:
I switched the base branch to remove the commits already on the target branch (to make it easier to do the merge).

PR #1467, opened 2022-06-11
Added Psalm and improved CI

Koen1999:


.github/workflows/static-analysis-psalm.yml
```diff
@@ -0,0 +1,102 @@
+name: PHP Static Analysis with Psalm
```
tomudding, line 1: We might want to split these two static analysis actions into another action (or maybe two) that acts as a composite action where we do the setup.

Koen1999:
@tomudding 

I set an appropriate error level for the project and generated a baseline containing some issues.
The Github action will show new errors, warnings, and informational notices introduced in the PR similar to what PHPStan does.

From what I have seen Psalm has some valid points and actually more than PhpStan at our current error level settings.

This PR also solved some small issues.

Moreover, static code analysis now also covers the test source files.

.github/workflows/static-analysis-psalm.yml
Koen1999: Performance wise this is probably a good idea. But I'm not that good with github actions

Koen1999:
You can see an example PR here: https://github.com/Koen1999/gewisweb/pull/8

.github/workflows/static-analysis-psalm.yml
tomudding: We can leave it like this for now, we just have to make sure that if we change something in the duplicate code we must change it in both actions.

Koen1999:
It works as expected now :D

module/Activity/src/Controller/ActivityController.php
```diff
@@ -290,6 +291,7 @@ public function signupAction(): Response|ViewModel
 
         if ($request->isPost()) {
             $form = $this->signupService->getForm($signupList);
+            /** @var ParametersInterface $postData */
             $postData = $request->getPost();
```
tomudding, line 295: This structure is used in many places (not everywhere) so we might want to do this in more places in the future.

Koen1999: Yes, but somehow these specific cases were problematic according to the static analysis

PR #1473, opened 2022-06-14
Nullable email address and clean-up of legacy code

tomudding:
Implements the changes from https://github.com/GEWIS/gewisdb/pull/160.

Also adds CSRF tokens to the registration and password reset forms. Additionally, this removes all the legacy code we have that has something to do with authentication (all of this code has not been functional for more than a year).


tomudding:
@Koen1999 it has become 2 PRs in one, so my apologies for that. I can split if you'd like.

I have yet to determine how I will fix the issues pointed out by PHP_CS and Psalm. Technically, the e-mail address can never be `null`, because at that point the `User` should not exist (though this is really part of #888 and obviously not yet implemented) and hence it will never be an issue. But I'd also like to have everything in `Decision` reflect what is in gewisdb.

tomudding:
This also fixes an issue where trying to log in without providing any credentials would throw an exception. Though I still think something is wrong with the login form.

module/Decision/src/Model/Member.php
```diff
@@ -708,26 +711,6 @@ public function toArray(): array
         ];
     }
 
-    /**
-     * @return array
-     */
-    public function toApiArray(): array
```
Koen1999, line 714: Why remove this method? It seems like a usable utility function

Koen1999 (approved):
Overall i believe the changes look good.

module/Decision/src/Model/Member.php
tomudding: It was only used in the API that was used by REX. I have removed the API as REX does not longer "exist" and the endpoint is a privacy problem. Without the API there is no need to keep the function and the standard `toArray()` in `Member` is almost the same so that can be used instead.

Koen1999:
Once PhpStan passes, you can merge this.

I'm not sure if Psalm is configured correctly at this point. But if you can get it to pass as well that would be great!

tomudding:
> Once PhpStan passes, you can merge this.
> 
> I'm not sure if Psalm is configured correctly at this point. But if you can get it to pass as well that would be great!

I have updated my last commit to fix a tiny mistake I made, Psalm now fails because it crashes (what I mentioned over WhatsApp).

PR #1488, opened 2022-07-08
Updated dependencies

Koen1999:


tomudding (approved):
A bit sad that it made an ugly commit, but at least it works.

PR #1499, opened 2022-08-04
Switch to Alpine based images

tomudding:
~~I wanted to change a bit more, but I think this is a good start.~~ This replaces all the existing base images for our images with Alpine based images. This should result in images that are generally ~~200-300~~ **500** MB smaller.

~~Further size reductions can be made by combining more steps in the Dockerfiles that allow us to use virtual `apk` packages that can subsequently be deleted. For example, the `pecl` installs require `$PHPIZE_DEPS` (among other things), we could create a new virtual package installing the dependencies, make the extensions, and remove the dependencies again. This should reduce the size of the layer even more.~~

**Important changes:**
- Moves production NodeJS version to LTS (16 instead of 14, parity with dev image),
- Updates ImageMagick to v7,
- Matomo no longer uses the latest version* (to prevent accidental breakages due to unseen updates),
- NGINX no longer uses the latest version (now `stable`), and
- `bash` is no longer available (use `sh`).

*: I know that Matomo v4.11 was released last night, but the creation of the Docker images takes a bit of time, hence this is still on v4.10.1.

### DIFF
--- a/docker/matomo/Dockerfile
+++ b/docker/matomo/Dockerfile
@@ -1,10 +1,8 @@
-FROM matomo:latest as gewisweb_matomo
+FROM matomo:4.10-fpm-alpine as gewisweb_matomo
 
-RUN apt update && apt install -y --no-install-recommends \
-    gettext-base \
-    unzip \
-    && apt upgrade -y --no-install-recommends \
-    && rm -rf /var/lib/apt/lists/*
+RUN apk add --no-cache \
+        gettext \
+        unzip
 
 RUN curl -o LogViewer.zip \
       https://plugins.matomo.org/api/2.0/plugins/LogViewer/download/latest \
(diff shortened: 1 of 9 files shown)


docker/matomo/Dockerfile
```diff
@@ -1,10 +1,8 @@
-FROM matomo:latest as gewisweb_matomo
+FROM matomo:4.10-fpm-alpine as gewisweb_matomo
```
Koen1999, line 1: Is it desirable to pin Matomo to a specific version?

docker/nginx/Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM nginx:latest as gewisweb_nginx
+FROM nginx:stable-alpine as gewisweb_nginx
+
+RUN adduser -D -H -u 1000 -s /bin/bash www-data -G www-data
```
Koen1999, line 3: Same comment about the usage of bash instead of sh as on the PR for gewisdb. We should probably replace the occurrences in the code.
afbeelding

Koen1999 (changes requested):
Overall, the changes look very promising. I tested it, and it works for as far as I can tell. With some minor adjustments, this is ready to merge.

docker/matomo/Dockerfile
tomudding: Yes and no, I want to prevent the image from updating automatically to ensure we do not miss the update. Because we have had an update that broke the analytics for a week because we missed the update. This is mostly because of database updates that need to be performed.

I must admit that this will increase our maintenance workload, but that is still less than the time it takes to debug the instance when something breaks due to updates.

docker/nginx/Dockerfile
tomudding: :/

PR #1500, opened 2022-08-10
Fixed some code issues

Koen1999:


module/User/src/Mapper/NewUser.php
```diff
@@ -57,9 +57,9 @@ public function getByCode(string $code): ?NewUserModel
      *
      * @param MemberModel $member
      *
-     * @return int
+     * @psalm-return list<mixed>
      */
-    public function deleteByMember(MemberModel $member): int
+    public function deleteByMember(MemberModel $member): array
```
tomudding, line 62: Doctrine :'( Maybe we should experimentally determine what is in the array?

module/User/src/Permissions/Assertion/IsAfterMembershipEnded.php
```diff
@@ -7,8 +7,7 @@
 use Laminas\Permissions\Acl\Assertion\AssertionInterface;
 use Laminas\Permissions\Acl\Resource\ResourceInterface;
 use Laminas\Permissions\Acl\Role\RoleInterface;
-use Photo\Model\Album;
-use Photo\Model\Photo;
+use Photo\Model\{Album, Photo};
```
tomudding, line 10:
```php
use Photo\Model\{
    Album as AlbumModel,
    Photo as PhotoModel,
};
```

tomudding (changes requested):

PR #1501, opened 2022-08-16
Require constructor promotion

tomudding:
Thanks to Rink!


tomudding:
This does not change any functional code, hence I am merging this without further reviews.

PR #1502, opened 2022-08-22
SuSOS --> SudoSOS

### LINKED ISSUE #1448: Change SuSOS to SudoSOS and remove legacy login service
It will take some time before these changes actually have to be done, but just to make sure we do not forget about it. I think SuSOS is only present in two places:

- [ ] The main menu and
- [ ] The members page.

We can already remove the legacy login service as it has not worked in ages, but I think it is a good idea to do it in conjunction the other changes.

rinkp:
Make SudoSOS the new default

SuSOS will be hidden for non-active members. Active members will keep the SuSOS option in the menu since the chances of them needing it are higher and they most likely are aware of SuSOS/SudoSOS

Upon merging, closes #1448. Legacy login was removed in #1473

### DIFF
--- a/module/Application/view/partial/main-nav.phtml
+++ b/module/Application/view/partial/main-nav.phtml
@@ -95,9 +95,16 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
                                     <?= $this->translate('My information') ?>
                                 </a>
                             </li>
+                            <?php if ($this->acl('decision_service_acl')->isAllowed('organ', 'edit')): ?>
+                                <li>
+                                    <a href="https://gewis.nl/susos/account/websitelogin.php">
+                                        <?= $this->translate('SuSOS') ?>
+                                    </a>
+                                </li>
+                            <?php endif; ?>
                             <li>
-                                <a href="https://gewis.nl/susos/account/websitelogin.php">
-                                    <?= $this->translate('SuSOS') ?>
+                                <a href="<?= $this->url('user_token', ['appId' => 'sudosos']) ?>">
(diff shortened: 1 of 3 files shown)


module/Application/view/partial/main-nav.phtml
```diff
@@ -95,9 +95,16 @@ $lang = $this->plugin('translate')->getTranslator()->getLocale();
...
+                                <li>
+                                    <a href="https://gewis.nl/susos/account/websitelogin.php">
+                                        <?= $this->translate('SuSOS') ?>
+                                    </a>
+                                </li>
+                            <?php endif; ?>
                             <li>
-                                <a href="https://gewis.nl/susos/account/websitelogin.php">
-                                    <?= $this->translate('SuSOS') ?>
+                                <a href="/token/sudosos">
```
tomudding, line 106: Can you change this to use the URL builder?

module/Decision/view/decision/member/index.phtml
```diff
@@ -42,9 +42,9 @@ $meetings = $this->meetings;
                         <div class="big-icon fas fa-image"></div>
                         <div class="panel-heading"><?= $this->translate('My photo\'s') ?></div>
                     </a>
-                    <a href="https://gewis.nl/susos/account/websitelogin.php" class="panel panel-image align-middle">
+                    <a href="/token/sudosos" class="panel panel-image align-middle">
```
tomudding, line 45: Same here.

tomudding (changes requested):

tomudding (approved):

PR #1506, opened 2022-08-31
Implement `hidden` members

### LINKED ISSUE #1504: Implement `hidden` accounts
### What would you like?

Per request of the board (@BHenkemans)
Hide birthdays and lock logins for specific `hidden` accounts.

Tags of these members in photos should still be shown

### Why is this needed?

Some members unfortunately have passed away but still are shown on the homepage when it is their birthday. Some members have been deleted such that everything but their member ID and password have been deleted, but those can still sign in. 

### Other information

_No response_

rinkp:
Closes #1504


module/Decision/src/Mapper/Member.php
```diff
@@ -50,10 +50,12 @@ public function searchByName(
...
 
         $sql = <<<QUERY
-            SELECT `lidnr`, CONCAT_WS(' ', `firstName`, IF(LENGTH(`middleName`), `middleName`, NULL), `lastName`) as `fullName`, `generation`
+            SELECT `lidnr`,
+                    CONCAT_WS(' ', `firstName`, IF(LENGTH(`middleName`), `middleName`, NULL), `lastName`) as `fullName`,
+                    `generation`
             FROM `Member`
-            WHERE CONCAT(LOWER(`firstName`), ' ', LOWER(`lastName`)) LIKE :name
-            OR CONCAT(LOWER(`firstName`), ' ', LOWER(`middleName`), ' ', LOWER(`lastName`)) LIKE :name
+            WHERE CONCAT_WS(' ', `firstName`, IF(LENGTH(`middleName`), `middleName`, NULL), `lastName`) LIKE :name
```
tomudding, line 57: This will prevent users from searching for people who have a middle name without also providing said middle name (partially or complete). Let's say we have a member `Jane van Doe`, and Alex is trying to find their profile yet they only now their first and last name, searching for `Jane D(oe)` will not yield any results (with the assumption that their are many `Jane`s).

module/User/src/Permissions/Assertion/IsAfterMembershipEnded.php
```diff
@@ -7,8 +7,7 @@
 use Laminas\Permissions\Acl\Assertion\AssertionInterface;
 use Laminas\Permissions\Acl\Resource\ResourceInterface;
 use Laminas\Permissions\Acl\Role\RoleInterface;
-use Photo\Model\Album;
-use Photo\Model\Photo;
+use Photo\Model\{Album, Photo};
```
tomudding, line 10: Can you put these on two separate lines?

```php
use Photo\Model\{
    Album,
    Photo,
};
```

Also applies to `BaseControllerTest.php`

tomudding (changes requested):

module/Decision/src/Mapper/Member.php
```diff
@@ -50,10 +50,16 @@ public function searchByName(
...
+            SELECT `lidnr`,
+                CONCAT_WS(' ', `firstName`, IF(LENGTH(`middleName`), `middleName`, NULL), `lastName`) as `fullName`,
+                `generation`
             FROM `Member`
-            WHERE CONCAT(LOWER(`firstName`), ' ', LOWER(`lastName`)) LIKE :name
-            OR CONCAT(LOWER(`firstName`), ' ', LOWER(`middleName`), ' ', LOWER(`lastName`)) LIKE :name
+            WHERE
+                (
+                CONCAT(LOWER(`firstName`), ' ', LOWER(`lastName`)) LIKE :name
+                OR CONCAT(LOWER(`firstName`), ' ', LOWER(`middleName`), ' ', LOWER(`lastName`)) LIKE :name
```
rinkp, line 60: I don't think there is a specific SQL statement formatting standard, but I settled for parentheses-are-the-only-character-on-a-line and indent-on-every-keyword

module/User/src/Permissions/Assertion/IsAfterMembershipEnded.php
rinkp: Actually I undid this change at all to conflicting changes with pull #1509. The other is fixed

tomudding (approved):

PR #1509, opened 2022-09-03
Allow `graduates` to see albums they are tagged in

### LINKED ISSUE #1472: Member's member album incorrectly limited for graduates

rinkp:
Also fixes #1472 provided your own member album is not empty


tomudding:
I fixed an issue in de945d4 that has existed for some time. I also made a change to ensure that sub-albums are not shown when a `graduate` is not supposed to have access for it.

Additionally, I change it so that albums without a cover default to a generic cover. The same applies to images that cannot be loaded:

image

tomudding:
Each additional album adds 2 queries on the yearly view. In an album itself you only get 4-6 additional queries (not related to anything). If there are sub-albums that are hidden you get an additional 4-6 queries (with 1 per sub-album).

PR #1512, opened 2022-09-12
Ensure correct timezone for all Docker containers

### LINKED ISSUE #1511: Cron jobs do not execute at expected time
### Current behaviour

The `web` container (and the others) defaults to UTC for all things date and time. This means that depending on summer time the jobs we have scheduled are executed 1 or 2 hours later than expected.

### Desired behaviour

The container uses a compatible timezone for the Netherlands (e.g., Europe/Amsterdam), which ensures that the cron jobs are executed when expected.

### Steps to reproduce

Just run the container and observe when the jobs are executed.

### Website version

7d26865

_No response_

_No response_

### Other information

_No response_

tomudding:
Fixes #1511.

### DIFF
--- a/docker/glide/Dockerfile
+++ b/docker/glide/Dockerfile
@@ -8,6 +8,7 @@ RUN apk add --no-cache --virtual .build-deps \
         libpng-dev \
         libwebp-dev \
         libzip-dev \
+        tzdata \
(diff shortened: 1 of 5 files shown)


tomudding:
This does not change any other existing behaviour as PHP already uses the correct timezone.

Issue #591, opened 2016-04-14
Ability to set a deadline for meeting authorizations


jszanto:

Koen1999:
Can you provide a description, perhaps including why this is neccesary?

tomudding:
This is not in the HR, so we cannot do this. #1524 added timestamps that should help with any issues that come up about when something was created and revoked or not.

PR #1524, opened 2022-10-11
Authorization revocation

tomudding:
Allows someone to revoke their authorization for a GMM.

Ideally, we implement #591 by adding a mandatory `startTime` to GMMs in [GEWIS/gewisdb](/GEWIS/gewisdb), though that is out of scope for this PR.

Unrelated, but this PR also removes all mentions of the Web Commissie.


tomudding:
PHPStan is misleading, that is how we always do it and Laminas broke it (it is still valid).

PR #1527, opened 2022-10-18
Fix generation of album covers without sufficient photos

### LINKED ISSUE #577: Photo: Generating album covers fails for albums which only contain subalbums
Even though this was specifically kept in mind when creating the album cover code, it somehow still fails.

### LINKED ISSUE #611: Album cover generation only uses the first subalbum

### LINKED ISSUE #1503: Albums with only subsubalbums don't have cover pictures
### Current behaviour

https://github.com/GEWIS/gewisweb/blob/80b304fbc1cdc6179b594d14a920e0fea2d4260f/module/Photo/src/Service/AlbumCover.php#L99

We only check direct subalbums, not recursively for subalbums until we have found photos

### Desired behaviour

Generating a cover for an album should have a cover photo with pictures from the topmost subalbum

### Steps to reproduce

Have an album with 0 photos, containing only subalbums with 0 photos having subsubalbums with more than 0 photos. 
Generating an album preview results in an empty picture

### Website version

cf7b469

_No response_

_No response_

### Other information

_No response_

tomudding:
Changes how the cover photo of an album is generated. This correctly uses all sub-albums that have photos (in a specific layer) to generate a cover.

This fixes #1503, fixes #611, and fixes #577. 

---
**OLD:**

This uses a native SQL to accomplish this, unfortunately, this does not support an early-abort (as recursive CTEs cannot `COUNT()` on results).

```sql
WITH RECURSIVE `subAlbumContents` (`depth`, `id`, `photo_id`) AS (
    -- select the root album and all its contents
    SELECT 0 AS `depth`, `a`.`id`, `p`.`id` AS `photo_id`
    FROM `Album` `a`
    LEFT JOIN `Photo` `p` ON `p`.`album_id` = `a`.`id`
    WHERE `a`.`id` = :album_id

    UNION ALL

    -- subsequently (and recursively) select its children and their contents
    SELECT `s`.`depth` + 1 AS `depth`, `a`.`id`, `p`.`id` AS `photo_id`
    FROM `Album` `a`
    LEFT JOIN `Photo` `p` ON `p`.`album_id` = `a`.`id`
    INNER JOIN `subAlbumContents` `s` ON `s`.`id` = `a`.`parent_id`
    -- to a limit of :depth INCLUSIVELY (even though we use a strict less than), + 1 from the SELECT must be accounted for
    WHERE `depth` < :depth
)
SELECT `photo_id`
FROM `subAlbumContents`
WHERE `photo_id` IS NOT NULL
    -- use first layer of albums that is not empty
    AND `depth` = (
        SELECT MIN(`depth`)
        FROM `subAlbumContents`
        WHERE `photo_id` IS NOT NULL
    )
-- RAND() is incredibly inefficient, but no better way to do this (scales linear with number of results in `subAlbumContents`)
ORDER BY RAND()
LIMIT :limit;
```

Because there is no early-abort, this can get quite slow for large albums and/or large max depths. I have tested it on the production data and it can return results in 0.01 to 0.02 seconds for large albums (including many sub-albums).

### DIFF
--- a/module/Photo/src/Controller/AlbumAdminController.php
+++ b/module/Photo/src/Controller/AlbumAdminController.php
@@ -210,9 +210,14 @@ class AlbumAdminController extends AbstractActionController
 
         if ($request->isPost()) {
             $albumId = $this->params()->fromRoute('album_id');
-            $this->albumService->generateAlbumCover($albumId);
+            if (null !== ($cover = $this->albumService->generateAlbumCover($albumId))) {
+                return new JsonModel([
+                    'success' => true,
+                    'coverPath' => $cover,
+                ]);
+            }
         }
 
-        return new JsonModel([]);
+        return new JsonModel(['success' => false]);
     }
 }
--- a/module/Photo/src/Mapper/Photo.php
+++ b/module/Photo/src/Mapper/Photo.php
@@ -3,6 +3,7 @@
 namespace Photo\Mapper;
 
 use Application\Mapper\BaseMapper;
+use Doctrine\ORM\Query\ResultSetMapping;
 use Photo\Model\{
     Album as AlbumModel,
     MemberAlbum as MemberAlbumModel,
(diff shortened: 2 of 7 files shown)


Koen1999:
Avoiding usage of the query builder significantly reduces the readability and maintainability of the code. Unless this change is unavoidable, I recommend refactoring.

tomudding:
> Avoiding usage of the query builder significantly reduces the readability and maintainability of the code. Unless this change is unavoidable, I recommend refactoring.

No I will refactor it and do the actual recursion in PHP to get the early-abort functionality we need.

PR #1531, opened 2022-11-01
Fix destroyed decisions not working as expected

tomudding:
Fixed some issues we had in the past.

**Note:** this should not be merged as is. This is far from performant (I have seen the number of queries increase 5x in certain scenarios).


tomudding:
This biggest problem with the number of queries is when you are loading actual organ members. First we need the `OrganMember`, then `Installation` subdecision and then we need the `Decision` to determine if it was destroyed or not.

tomudding:
Not solvable at this point in time.

PR #1535, opened 2022-11-19
Allow courses to be updated

### LINKED ISSUE #1042: Improve error message when adding courses
As per the e-mail from Koen, the error message(s) when adding courses are not very descriptive.

We can probably do something with the form validation to have proper error messages appear (like in almost every other module).

tomudding:
As requested by @BHenkemans, this adds functionality to be able to update and delete courses, view documents for specific courses and be able to delete them.

This removes a lot of things that are no longer needed. Courses can no longer have parents or children. Additionally, URLs, year, and quartile have been removed.

image

image

Closes #1042.


tomudding:
PHPStan is expected, as the PR baseline is not up-to-date.

PR #1536, opened 2022-11-20
Generate a `Message-Id` for e-mails

### LINKED ISSUE #1534: Generate `message-id` for e-mails
Apparently `laminas/laminas-mail` does not do this by itself.

tomudding:
`laminas/laminas-mail` does not do this automatically and not having a `Message-Id` can be bad for the spam score of our e-mails.

Fixes #1534.


tomudding:
Here is an example:
image

PR #1537, opened 2022-12-04
Dependencies 2022-12

tomudding:
Long due, but was previously not possible due to many bugs in Doctrine ORM.


tomudding:
The Psalm action does not care about the updated baseline as it is doing the PR baseline test. But once merged it will be good to go.

tomudding:
This is a good preparation for PHP 8.2 which will be released on the 8th. Almost all Laminas package have support for it, only a few other packages do not yet fully support PHP 8.2 (according to Packegist).

PR #1541, opened 2022-12-08
Upgrade to PHP 8.2.0

tomudding:
**The official PHP Docker images have not yet been updated. So this will currently not work. I expect the images to be updated in a few hours.**

Blocking dependencies:
- ~~`friendsofphp/php-cs-fixer` (no update announced)~~ (only affected when manually called)

Upgradable dependencies:
- `doctrine/doctrine-module` can be upgraded from 5.2.1 to 5.3.0
- `doctrine/doctrine-orm-module` can be upgraded to 5.3.0
- `laminas/laminas-code` can be upgraded from 4.7.1 to 4.8.0
- `laminas/laminas-component-installer` can be upgraded from 2.8.0 to 3.1.0
- `laminas/laminas-developer-tools` can be upgraded from 2.5.0 to 2.6.0
- `laminas/laminas-cache-storage-deprecated-factory` can be upgraded
- `laminas/laminas-text` can be upgraded to 2.10.0
- `slam/phpstan-laminas-framework` can be upgraded to 1.4.0


tomudding:
Tested locally, including with PHPStan and Psalm. All errors have been resolved and the baselines updated where necessary. All appears to be well, nothing is broken as the upgrade to PHP 8.2 was very minimal.

The two failing actions are due to the incompatible composer.json on master.

PR #1561, opened 2023-01-29
Font and SVG preloading (fixes #1560)

### LINKED ISSUE #1560: Font preloading issue
### Current behaviour

1. https://gewis.nl/fonts/lato/lato-v22-latin-italic.woff2 is not preloaded, because the font is not referenced for preload
2. https://gewis.nl/fonts/gewisicons/gewisicons.ttf?37l19n is not preloaded, most likely because of the additional = in https://github.com/GEWIS/gewisweb/blob/34e1aae4f23d63d4ba65a0ab61994059acac0558/module/Application/view/layout/layout.phtml#L58

### Desired behaviour

_No response_

### Steps to reproduce

Firefox: https://www.webpagetest.org/result/230129_BiDc82_8V7/1/details/#waterfall_view_step1

Chrome: https://www.webpagetest.org/result/230129_BiDc61_8T7/1/details/#waterfall_view_step1

### Website version

6b63c7311a13a54703922bc619ddffc4e612aded

_No response_

Chrome, Firefox, Microsoft Edge

### Other information

I can submit a PR if desired

rinkp:
Actually, I saw the SVG resources also did not get used; the second commit fixes that as well

### DIFF
--- a/module/Application/view/layout/layout.phtml
+++ b/module/Application/view/layout/layout.phtml
@@ -47,6 +47,10 @@
     ]); ?>
     <?php $this->headLink([
         'as' => 'font', 'rel' => 'preload', 'type' => 'font/woff2', 'crossorigin' => 'anonymous',
+        'href' => '/fonts/lato/lato-v22-latin-italic.woff2'
+    ]); ?>
+    <?php $this->headLink([
+        'as' => 'font', 'rel' => 'preload', 'type' => 'font/woff2', 'crossorigin' => 'anonymous',
         'href' => '/fonts/fontawesome/fa-solid-900.woff2'
     ]); ?>
     <?php $this->headLink([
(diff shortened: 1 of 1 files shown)


module/Application/view/layout/layout.phtml
```diff
@@ -55,14 +59,14 @@
     ]); ?>
     <?php $this->headLink([
         'as' => 'font', 'rel' => 'preload', 'type' => 'font/ttf', 'crossorigin' => 'anonymous',
-        'href' => '/fonts/gewisicons/gewisicons.ttf?37l19n='
+        'href' => '/fonts/gewisicons/gewisicons.ttf?37l19n'
     ]); ?>
     <?php $this->headLink([
-        'as' => 'image', 'rel' => 'preload', 'type' => 'image/svg+xml', 'crossorigin' => 'anonymous',
```
Koen1999, line 61: Why is the crossorigin removed? Shouldn't links have this attribute?

Koen1999: This bug report on Chromium seems related: https://bugs.chromium.org/p/chromium/issues/detail?id=1065069

rinkp: I had that one open for a while as well as a Stackoverflow for that issue

SVG resources are documents in HTML (as opposed to images). Hence, you can interact with their contacts as with an embedded HTML. The way the SVG is included enforces it to be included later like that. I removed it entirely since 'use-credentials' did not seem appropriate and the only two allowed values are 'anonymous' and the former. My testing showed that Chrome accepts this and does not do a second request for the two SVGs.

This actually allows for interesting things such as modifying SVG colours with JavaScript so they are current origin objects. With this fix, preloaded resources are used

Koen1999: Well, I suppose crossorigin doesn't matter a lot since it's hosted on the same domain.

Koen1999 (approved):

Issue #1558, opened 2023-01-26
Remove `*.gewis.nl` from `img-src` CSP directive


tomudding:

rinkp:
Sounds like a reasonable thing to do. Do you know whether the connecr-src for supremum.gewis.nl is still used now infimum are cached?

tomudding:
> Sounds like a reasonable thing to do. Do you know whether the connecr-src for supremum.gewis.nl is still used now infimum are cached?

No, it seems I have been a bit sloppy with updating it. This and the `https://fonts.googleapis.com` like you mentioned will be removed.

Issue #1556, opened 2023-01-18
Restrict availability of session cookies on subdomains


tomudding:
Both `GEWISSESSID` and `GEWISSESSTOKEN` are currently availability on subdomains, this is not necessary.

tomudding:
Reported by @rinkp.

tomudding:
Requires removal of the specific domain, to prevent browsers from creating a wildcard.

rinkp:
Maybe we can combine this with a rollover of the key because then we guarantee all cookies to be set again in any case. (since a few weeks, wiki.gewis.nl is now also abiding by this standard with the similarly named `GEWIKISESSTOKEN` (XSRF token was already hostonly)

tomudding:
> Maybe we can combine this with a rollover of the key because then we guarantee all cookies to be set again in any case. (since a few weeks, wiki.gewis.nl is now also abiding by this standard with the similarly named `GEWIKISESSTOKEN` (XSRF token was already hostonly)

Yes that is a good idea!

PR #1562, opened 2023-02-02
Security improvements for cookies and CSP

tomudding:
Due to how browsers interpret the `Domain` attribute of cookies, our session cookies were accessible from subdomains. This is no longer the intended behaviour, hence we remove the domain from cookies to ensure only the host is used instead of `*.gewis.nl`. For more information, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#domaindomain-value.

In terms of the CSP, the following changes were made:
- `connect-src` for the Supremum website is not needed as we use a local API for this, see GH-1220.
- `img-src` for `*.gewis.nl` is too broad and has therefore been removed.
- `style-src` for Google Fonts is no longer needed since we moved to locally hosted fonts, see GH-1352.

Fixes GH-1556 and fixes GH-1558.


tomudding:
When deploying this, we should rotate the keys for `GEWISSESSTOKEN` to ensure that all old cookies are invalidated.

PR #1579, opened 2023-02-03
Fix subnet for university NAT

tomudding:
A minor typo was made which caused `guest`s to not be able to see exams and summaries when on the university's NAT'ed subnet.

Discovered this together with @rinkp.


tomudding:
I don't know why the actions do not work, but they are not needed for this PR.

Issue #1330, opened 2021-12-28
Implement breached password check when resetting or activating accounts


tomudding:
A validator for this was introduced in `laminas/laminas-validator`, see [undisclosed passwords validator](https://docs.laminas.dev/laminas-validator/validators/undisclosed-password/). This can be used to check new passwords against the Pwned Passwords API from Troy Hunt.

Koen1999:
It's amazing that they actually added this feature!

tomudding:
I have talked with the CBC about this and we have devised a better solution than checking the actual API of Pwned Passwords. As there is plenty of RAM too spare, we can implement an actual Bloom Filter in Redis such that we can check the passwords locally.

We do not even have to generate the bloom filter ourselves, as we can get a copy from [here](https://scotthelme.co.uk/re-bloom-pwned-passwords-v8/).

So to summarise:
1) New Docker container with Redis (and `redisbloom`).
2) Permanently storing the bloom filter dump and load it every time the container is restarted.
3) Query passwords via [palicao/phpRebloom](https://github.com/palicao/phpRebloom) in gewisweb.

```php
use Palicao\PhpRebloom\BloomFilter;
use Palicao\PhpRebloom\RedisClient;
use Palicao\PhpRebloom\RedisConnectionParams;

$bloomFilter = new BloomFilter(
	new RedisClient(
		new Redis(),
		new RedisConnectionParams('redis', 6379),
	)
);

$exists = $bloomFilter->exists('pwned-bloom', strtoupper(sha1($password)));
```

PR #1580, opened 2023-02-04
Add checks for Pwned Passwords and more QoL password improvements

tomudding:
First and foremost, this implements checks against a Pwned Passwords API when passwords are handled (registrations, resets, changes, and logins). This currently uses the GEWIS version of the API (see https://pwned-passwords.gewis.nl, note: only available from the campus). If a password is deemed to be in a public data breach it cannot be used. If a user logs in with a known pwned password they must reset their password before they can log in.

All related e-mails now also use the new e-mail template by Stijl.

The hacky `EntityManager` code in `LoginAttempt` has been removed by not detaching `User`-type entities during the authentication process.

The rate limits have been slightly adjusted and the validity of links has been limited to 24 hours.

Some pictures:
image

image

image

image

Closes GH-1330.


tomudding:
Most of the issues from the Actions are due to the new PhpStorm UI, which does not really do a good job of highlighting errors.

tomudding:
PHP_CS is a non-issue, as it curently does not support `readonly` classes.

Issue #1590, opened 2023-02-13
Change page title to something else than 'GEWIS Website'


tomudding:
'GEWIS Website' is maybe a bit too much.

PR #1603, opened 2023-02-27
Change website title to "Study Association GEWIS"

tomudding:
This is better than "GEWIS Website". The Dutch translation is "Studievereniging GEWIS".

Closes GH-1590.


tomudding:
I have been a bit too agressive with resetting some translations. Will be fixed.

PR #1606, opened 2023-02-28
Add `Keyholder`s

tomudding:
This is a parity update for GEWIS/gewisdb#256. Allows board members to view whether someone has a key code (or rather if such a decision has been made).


tomudding:
I may have been a bit inconsistent in the naming of the actual decision in GEWISDB, so that is why it is like this.

PR #1610, opened 2023-03-28
Compare meeting type values for meeting selection document upload

tomudding:
Otherwise, all documents that are uploaded are always uploaded to the latest available meeting instead of the selected meeting.

Fixes ABC-2303-157.


tomudding:
Changes nothing critical, so the checks are not necessary.

Issue #1627, opened 2023-04-23
Move towards `declare(strict_types=1);`


tomudding:
Similar plans exist for GEWISDB, however, because that is a lot more important it is taking a bit longer. To test everything and to ensure that it is indeed possible to switch we should also do this for GEWISWEB.

Issue #1294, opened 2021-12-03
Typing `$this` in template files


tomudding:
Laminas allows for doing

```php
/**
 * @var Laminas\View\Renderer\PhpRenderer $this
 */
```

in the `.phtml` template files. This enables PhpStorm (or any other modern IDE) to autocomplete `$this` with all the Laminas view stuff. In places where custom ViewHelpers are used, one can also concatenate `HelperTraits`:

```php
/**
 * @var Laminas\View\Renderer\PhpRenderer|Laminas\Mvc\Plugin\FlashMessenger\View\HelperTrait $this
 */
```

This might be useful to make editing of template files easier.

tomudding:
For forms:
```php
\Laminas\Form\View\HelperTrait
```

For translating:
```php
\Laminas\I18n\View\HelperTrait
```

PR #1628, opened 2023-04-23
More and better typing

tomudding:
These changes bring a lot to our code base. First and foremost, `declare(strict_type=1)` is now always required in a PHP file (including the `.phtml` view templates). On top of that, I have added correct typing to most of the view templates themselves, this means that auto-complete can now also work in templates. It should be noted that typing in view templates is not automatically available, this still requires manual insertion of type definitions.

I have also changed the definition of `BaseMapper` to be a template such that we can get proper return types from it. On top of that I am making it mandatory (though not yet enforced nor fixed in the code base) that complex `array` return types should be correctly typed.

Closes GH-1294 and closes GH-1627.


tomudding:
I have tested almost every aspect of the website to ensure that everything still works after enforcing the strict types. I believe I have fixed most if not all of the issues that were caused by it.

tomudding:
I have also updated the Psalm baseline to remove some of the issues we had. I have also updated the `MeetingDocument` baseline, however, it will fail as long as it is not on master (so failure of Psalm is expected).

Issue #1639, opened 2023-05-15
Two organs with the same name don't show up on profile pages


rinkp:
### Current behaviour

If a member has been in 2 organs with the same name (e.g. https://gewis.nl/association/avc/RvA and https://gewis.nl/association/rva/RvA), it will not show up on their profile page as such.

### Steps to reproduce

If you have been in the first, it will show up properly (e.g. [7533](https://gewis.nl/member/7533)). If you have only been in the second, it will also show up ([9838](https://gewis.nl/member/9838)), but if you were in both ([9093](https://gewis.nl/member/9093), [8994](https://gewis.nl/member/8994)), it won't show up.

### Website version

cddb27a7e65bf45df93bb597d159e34f0739ae81

PR #1640, opened 2023-05-15
Fix organs not being distinct when sharing a name

tomudding:
Note that this only fixes this when displaying a list of organs on the profile of a member. A similar problem was addressed in GH-982 but can never be completely fixed.

Fixes GH-1639.


tomudding:
> A similar problem was addressed in https://github.com/GEWIS/gewisweb/pull/982 but can never be completely fixed.

Unless we start using the foundation decision in the URL instead of the abbreviation, but that is ugly for pages that are accessible to non-members.

PR #1644, opened 2023-05-18
Updated dependencies

Koen1999:
It's just another regular jump. I did not update `doctrine/orm` since you mentioned we were waiting for a patch. Originally I wanted to update the service manager as well, but there are other dependencies that require modifications first.


tomudding (approved):
Works as expected; all the FontAwesome icons are still shown (they have a tendency to change names). ORM will have to wait.

PR #1647, opened 2023-05-25
Fix issue with sorting `Organ`s by hash

tomudding:
The original code from GEWISDB (as copied in 2015) contained a tiny mistake in `getDecisionNumber()`, the subdecision number was returned instead of the decision number.

This was fixed in 2017 in GEWISDB, however, the fix was never copied to the website.


tomudding:
The fix by Justin was made in https://github.com/GEWIS/gewisdb/commit/797a1ebbf1f3d4fb6d7bc2536c180dde4ac4c453.

Issue #1585, opened 2023-02-06
`doctrine/orm` regression for entities with `BackedEnum` in association mapping


tomudding:
### Current behaviour

> Error: Object of class Decision\Model\Enums\MeetingTypes could not be converted to string in /code/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php:1575 Stack trace: 
#0 /code/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php(1575): implode(' ', Array) 
#1 /code/vendor/doctrine/orm/lib/Doctrine/ORM/UnitOfWork.php(3298): Doctrine\ORM\UnitOfWork->addToIdentityMap(Object(DoctrineORMModule\Proxy\__CG__\Decision\Model\Meeting)) 
#2 /code/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php(692): Doctrine\ORM\UnitOfWork->registerManaged(Object(DoctrineORMModule\Proxy\__CG__\Decision\Model\Meeting), Array, Array) 
#3 /code/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(267): Doctrine\ORM\Internal\Hydration\AbstractHydrator->registerManaged(Object(Doctrine\ORM\Mapping\ClassMetadata), Object(DoctrineORMModule\Proxy\__CG__\Decision\Model\Meeting), Array) 
#4 /code/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(498): Doctrine\ORM\Internal\Hydration\ObjectHydrator->getEntity(Array, 'r') 
#5 /code/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php(149): Doctrine\ORM\Internal\Hydration\ObjectHydrator->hydrateRowData(Array, Array) 
#6 /code/vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php(270): Doctrine\ORM\Internal\Hydration\ObjectHydrator->hydrateAllData() 
[...]

PR #1652, opened 2023-06-03
Update dependencies

tomudding:
Most important part of this update is the release of Doctrine ORM version 2.15.2 which contains a fix for a bug that impacted the database (join columns could not contain enums).

This actually fixes GH-1585.


tomudding:
Tested and works again as expected.

PR #1654, opened 2023-06-08
Move to unified GEWISPHP Coding Standards

tomudding:
This will require a lot of changes, but it should make the whole codebase a lot better.


tomudding:
Almost all of the remaining PHP CodeSniffer errors are related to `Traversable` (i.e., `array`-like) return definitions, which I have not yet added or suppressed.

Issue #1656, opened 2023-06-24
`translatePlural` not recognised when using `xgettext`


tomudding:
The `translate-helper` does not recognise the usage of `translatePlural`. This means that these strings are never translated. An example of this would be the "Upcoming meeting(s)" string on the Members page, which has not yet been translated.

PR #1657, opened 2023-06-24
Update translations and fix `translate-helper`

tomudding:
The `translate-helper` now properly recognised the usage of `translatePlural`, which is also reflected in any capable PO editor.

Fixes GH-1656.


tomudding:
_I don't know why MariaDB keeps shutting down by unknown forces._

PR #1663, opened 2023-07-02
Do not show birthdays of recently expired people

tomudding:
The expiration date of a membership/graduate status is 'exclusive'.


tomudding:
Let's see if this can (initially) also not be merged just like with the commit on GEWISDB.

Issue #1383, opened 2022-03-14
Improve approve organ page


LeonVreling:
Currently, the board needs to approve changes an organ makes to their page on the GEWIS website. This process is, however, not that nice in my personal experience.

One thing that could be improved is a preview of the image that you are approving, cause now you can only view the text. We could maybe make a new page that shows the old version on the left side, and the new proposed changes on the right side. This also makes the checking of the text easier.

Let me know what you think about this 😄

tomudding:
I have always wanted to properly implement a Git-like diff system (which would also be used in the Activity module), however, I have not found a package that can do this for natural text (and I most certainly do not have the time to implement something like that myself).

If we get something working, we can also use it in the upcoming update to the company module, as that also heavily relies on diffs.

tomudding:
Perhaps [lolli42/finediff](https://packagist.org/packages/lolli42/finediff) can do the diff thing, however, images might be a bit more difficult.

tomudding:
[jfcherng/php-diff](https://packagist.org/packages/jfcherng/php-diff) is probably the best package we can get if we want to do this in PHP. We can potentially also move this to something JavaScript based (if performance becomes an issue).

tomudding:
![image](https://github.com/GEWIS/gewisweb/assets/4983571/34628e91-e54e-4e69-9ffc-76c6ddccdcaa)

Side by side for larger HTML-based textual diffs. Inline for simple text diffs. Unified/combined for single line diffs.

PR #1669, opened 2023-07-09
Improve reviewing of update proposals by using proper diffs

tomudding:
Uses `php-diff` to generate different styles of diffs for different kinds of texts:"
- Combined: single-line text
- Inline: longer multi-line text (e.g., descriptions)
- SideBySide: long complex text (e.g., HTML pages)

Does require a small patch to ensure that we always generate a diff, even when the inputs are equal (see [GEWIS/gewisweb-php-diff](https://github.com/GEWIS/gewisweb-php-diff)).

This forms the basis for GH-1383 (which requires some extensive changes).


tomudding:
The constant most certainly exists.

Issue #1676, opened 2023-08-08
Adding number fields to activities is not possible


rinkp:
### Current behaviour

When editing an (unapproved) activity, it is not possible to add a numerical signupfield. Neither board members nor the creator can edit this

When first adding a text field and then changing the type, the following error is shown because the min and max fields are not shown and don't have a default value:
![image](https://github.com/GEWIS/gewisweb/assets/17340320/a3b1a171-acea-47c6-af7e-ad31d355b19f)

### Steps to reproduce

1. Create an activity as a regular user with a signuplist
2. Try to add a signup list field
3. A post is submitted to the same page and your field is gone without any feedback
4. Posting again will not create the field instead saving the activity as is

### Website version

715f02d06e9dcd8389cd4761efac55139baf463e

### Other information

First reported by Lucy

tomudding:
This is a JavaScript issue from what I can see, the `select` requires an additional update before displaying the fields.

tomudding:
Actual cause is that the `click` event is on the `<select>` and not the `<option>`s.

PR #1678, opened 2023-08-08
Fix `<select>` for signup list field type not updating correctly

tomudding:
The wrong event was used, `click` does not cause an (additional) update when the actual type is selected. This works properly with `change`, as the `<select>`s are still dynamic we need a global event listener for `change` on the page.

This fixes GH-1676.


tomudding:
This has been the case for almost a year and no one noticed :(

Issue #1682, opened 2023-08-18
Improved i18n routing


tomudding:
### What would you like?

In GH-1370 the i18n aspects of the website were improved, including alternate `hreflang`s in the `<head></head>` for translations. Unfortunately, this negatively impacted search engine results due to the additional redirect. The functionality was disabled in bbfd63e2362ef3eb1f6ee16775b5cc87be17db4a to mitigate this.

Because we cannot go the route of the additional redirect, I suggest implementing a new router stack that is able to handle `https://gewis.nl/en/{actual route}` and `https://gewis.nl/nl/{actual route}`.

I still have to figure out how to make the old URLs work* but parsing the `/:lang/` part is relatively easily (as the route stack matcher will just have to change the `baseUrl` to include the `:lang/` part to have the normal route stack matching handle everything else).

This also includes the automatic detection of language and changing the URL accordingly.

tomudding:
On top of this we can also use the `TranslatorAwareTreeRouteStack` to translate the traditional routes (after the new route stack as proposed above). This would mean that you have, for example, `gewis.nl/en/activity` and `gewis.nl/nl/activiteit`.

tomudding:
I don't know which language Google prefers when indexing. Maybe to make sure that Google does not get mad, if your preferred language is English you can stay on `/` (and not be redirect to `/en/`) and if Dutch you automatically go to `/nl/`. `/en/` will of course work too.

rinkp:
> I don't know which language Google prefers when indexing. Maybe to make sure that Google does not get mad, if your preferred language is English you can stay on `/` (and not be redirect to `/en/`) and if Dutch you automatically go to `/nl/`. `/en/` will of course work too.

I think having `/en/` and `/nl`/ would be a good idea. Currently the default is English so we can keep the homepage like that.

rinkp:
When we do this, we may consider having both an English and a Dutch title for custom pages, but on the other hand I don't think the router supports that, so then we would need an if-else everywhere we generate URLs (such as navigation)

tomudding:
> When we do this, we may consider having both an English and a Dutch **_title_** for custom pages, but on the other hand I don't think the router supports that, so then we would need an if-else everywhere we generate URLs (such as navigation)

Title or route? For routing of custom pages we can just add an additional set of "routes". Based on the `:lang` from `gewis.nl/:lang/:category[/:sub_category][/:name][/]` we can determine whether we need to look up the Dutch route or the English route.

[...]

PR #1686, opened 2023-08-20
Add language aware route stack to allow language specific links

tomudding:
This route stack can be used to parse and handle the routing of links which include a language identifier, for example, `en` and `nl`.

The code to determine the preferred language of the visitor has been removed from `onBootstrap()` to reduce its overhead. It has been moved to the new route stack to handle everything during the route matching phase.

A listener has been added for `MvcEvent::EVENT_ROUTE` to ensure that we can get the language from the route matching into the actual translators and relevant i18n code.

The new route stack also enables the automatic inclusion of the preferred language in the generation of links through the `url()` view helper. Manually changing the `language` parameter is fully supported too.

This also ensures that alternate `hreflang`s can be properly provided. If there is a need to overwrite the values, this can be done using the `hrefLang()` view helper.

All custom `Page`s are now using `FrontpageLocalisedText` for the translations. This also means that the routes are now "translatable". To make this possible, the `Page` form has been extended to accept pairs for the routes instead of single values. To make sure that you can never have two distinct pages with the same routes, the validation of the form has been extended.

On creation of new pages the routes are checked against all known routes from the database. If a page is edited, it is determined whether the route needs to be checked (against the database).

However, this means that you cannot use the Dutch route when viewing the website in English and vice-versa. So `/en/some/important/page` works but `/nl/some/important/page` not (unless you use English routes for Dutch pages). The language selector ensures that you can easily switch between the different languages, even on custom pages. To enable this, the `LanguageAwareTreeRouteStack` can handle replacing parts of the routing parameters for custom pages.

This also changes some of the hardcoded URLs for certain custom pages to be more descriptive of the actual page.

---

These changes come with a minor performance penalty. Locally, worst-case scenarios show up to a 20% increase in processing time. However, on average it is less than 5% (or even faster).

---

There is currently no support for translatable routes even though we extend the `TranslatorAwareTreeRouteStack`. On top of that, future changes will add support for language specific routes for custom pages.

---

This closes GH-1682.


tomudding:
Psalm issue is already in the baseline.

tomudding (commented):
I do not see anything that will break something in production, the database does require an extensive migration but this should not be an issue.

Additionally, I have extensively tested this and have not been able to completely break it.

Issue #1688, opened 2023-08-24
No-downtime sync with GEWISDB


rinkp:
### What would you like?

Currently, when syncing data with `gewisdb`, the website may throw errors for a small amount of time. This prevents syncing more often and is not a good user experience. 

### Other information

This is related to the use of the TRUNCATE statement which always does a commit and thus tables may be empty for a while before the transaction with the inserts is committed.

Issue #1474, opened 2022-06-15
Reportdb import script can cause loss of data if failure occurs


tomudding:
Our beloved script to import data from reportdb has another flaw. If an failure occurs, the transaction is never rolled back and auto committed on exit. This means that tables can be empty after truncation.

There should be a `ROLLBACK;` statement somewhere in this `catch`:

https://github.com/GEWIS/gewisweb/blob/ef1057922fc98d2d65a131d660a91840a521d194/importdb.php#L101-L107

tomudding:
Though I would personally re-order the file a bit, such that a rollback will not occur when the connection cannot be established.

tomudding:
Doing a transaction and commit (+ potential rollback) per table might be better, but you can be left with incomplete or potentially incompatible data.

PR #1689, opened 2023-08-24
Implement atomic database sync

rinkp:
With this pull request, the database sync no longer truncates all data in the table, but instead uses inserts, updates and deletes. 

This fixes GH-1688 (as well as fixes GH-1474)

## Changes
The `TRUNCATE` statement is removed (which had an implicit `COMMIT`, causing the website to be unavailable for a short period of time during sync and is replaced by `INSERT ... ON DUPLICATE KEY UPDATE` in combination with a `DELETE FROM` statement that removes no longer existing entries. Foreign key checks are still disabled because the foreign keys do not have any cascading settings.

## Consequences
We assume that the primary key does not change over the lifetime of a record since we are not able to pick up changes to that. Hence, we are not able to update the `User` object when a `Member` changes `lidnr`. However, this assumption was also made with the previous solution. 

## Why would we want this
It enhances user experience by not throwing errors during the database sync. Furthermore, it paves the way into more frequent (or on-demand) syncs because sync becomes seamless. These may allow members to sign in without waiting a day when their email address has been updated or they have just been registered. 

Optionally, a partial sync could be done separately which does not remove any rows but instead only relies on inserts and updates, but tests show that the delete statements do add little overhead. 

## Why would we not want this
The script becomes more advanced and does not work for newer (non-MariaDB) MySQL out of the box. To use MySQL 8.0 and newer, it is needed to make the following changes:

```diff
function ($a) {
-    return $a . '=VALUES(' . $a . ')';
+    return $a . '=new.' . $a;
},
```

```diff
$sql = sprintf(
-    'INSERT IGNORE INTO %s %s VALUES %s ON DUPLICATE KEY UPDATE %s',
+    'INSERT IGNORE INTO %s %s VALUES %s as new ON DUPLICATE KEY UPDATE %s',
    $table,
```

If we are ever going to have auto incrementing columns in the database, we need to also make sure that the `AUTO_INCREMENT` value is reset to the max of that column to prevent exhausting available numbers. The reason for this is that the `INSERT` statement always reserves a new incrementing value. Currently, we do not have any default value set to `AUTO_INCREMENT` in the MySQL side. 

## Tests performed
This pull request has been tested on tables with 10k simple entries and 12 regular ones (and some extra `sleep(5)` statements to simulate slowness). I suggest testing this on test.gewis.nl to see what happens in practice. Groups of rows have been edited with simple mysql statements and deleted to see how the sync behaves. 

```python
import secrets
for i in range(12,10000):
    print("INSERT INTO \"public\".\"member\" (\"lidnr\", \"email\", \"lastname\", \"middlename\", \"initials\", \"firstname\", \"generation\", \"type\", \"changedon\", \"membershipendson\", \"birth\", \"expiration\", \"paid\", \"supremum\", \"authenticationkey\") VALUES ('" + str(i) + "', 'test" + str(i) + "@gewis.nl', 'Mechanical Engineering', 'van', 'M.', 'Marco', '2023', 'external', '2023-08-22', '2023-07-01', '2000-01-01', '2024-07-01', '15', NULL, '" + secrets.token_hex(64) + "');")
```

### DIFF
--- a/config/autoload/gewisdb.local.php.dist
+++ b/config/autoload/gewisdb.local.php.dist
@@ -8,4 +8,6 @@ return [
     'user'     => getenv('DOCKER_DB2_USERNAME'),
     'password' => getenv('DOCKER_DB2_PASSWORD'),
     'dbname'   => getenv('DOCKER_DB2_DATABASE'),
+    'charset' => 'utf8',
+    'collate' => 'utf8_unicode_ci',
 ];
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -36,6 +36,8 @@ services:
             context: .
         depends_on:
             - mysql
+        extra_hosts:
+            - "host.docker.internal:host-gateway"
     postfix:
         image: mailhog/mailhog
         env_file:
--- a/importdb.php
+++ b/importdb.php
@@ -1,135 +1,236 @@
 <?php
+
+declare(strict_types=1);
+
 /**
- * This is a separate script that copies the GEWIS Report Database to the Web
- * database.
+ * This is a separate script that copies the GEWIS Report Database to the website database.
  *
  * It is a simple PostgreSQL to MySQL copy script.
  */
 
-echo "Commencing sync with gewisdb\n";
+echo 'Commencing sync with GEWISDB...' . PHP_EOL;
 
 try {
-// connections
+    // Setting up connections.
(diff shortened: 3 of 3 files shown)


tomudding (approved):
Thank you @rinkp! I have changed a few things here and there to hopefully make it easier to see what is actually happening. This also includes the necessary changes for GH-1474.

Issue #1708, opened 2023-09-13
Synchronisation of GEWISDB does not always remove decisions


tomudding:
### Current behaviour

The sync script to get data from GEWISDB does not always remove decisions when they were deleted in GEWISDB.

### Steps to reproduce

Add decisions, sync, remove decisions, sync. Not all removed decisions are actually removed.

### Website version

020e67e039a2498cc709bf982f75c82d013ea7d0

tomudding:
https://github.com/GEWIS/gewisweb/blob/020e67e039a2498cc709bf982f75c82d013ea7d0/importdb.php#L194-L199

The query and `$pks[$table][0]` do not account for composite keys (as @rinkp already thought would be the issue).

PR #1709, opened 2023-09-13
Fix multi-column primary keys in importdb script

rinkp:
Should fix GH-1708


tomudding (approved):
```sh
Syncing table "SubDecision"...
............................................................

Inserted or updated 0 rows in "SubDecision" (updates count double)
Deleted 5 rows from "SubDecision"
```

Issue #1701, opened 2023-09-09
Links to other associaitions on Education Page are broken


BHenkemans:
### Current behaviour

It was reported that most of the links on the education are broken / behind a login. So they are no use to our students anymore. Normally, I would ask the board to fix this, but this page cannot be changed by the board.

### Steps to reproduce

Go to the education page and try the links of 

- SvBMT Protagoras
- e.t.s.v. Thor
- Jan Pieter Minckelers
- Intermate

### Website version

v4.0 (dccb9614f)

### What operating are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Firefox

tomudding:
Sad to see other associations putting things behind a login. We can just remove that whole section.

PR #1716, opened 2023-09-26
Removed links of education pages which are now behind a login wall.

BHenkemans:
Closes GH-1701.


tomudding:
The merge by @BHenkemans was intentional as we were testing which privileges he had (owner privileges).

Issue #1726, opened 2023-10-03
On course page allow linking to previous iterations of courses


BHenkemans:
### What would you like?

I would like to have the previous iterations (or future iterations) of course to be added to the course page. Something like this.

![image](https://github.com/GEWIS/gewisweb/assets/38168308/c970cb35-1d8a-47fc-b891-2df785226280)

### Why is this needed?

Due to the bachelor college redesign, a lot of courses have changed slightly but not enough to make previous documents irrelevant. Therefore, informing students of the previous iterations of the course could be useful

### Other information

It would require the Educational Officer link the courses, but he himself was interested in this feature too, so I am sure he wont mind too much (@JortvD ;))

JortvD:
Yeah I had a very similar idea for the similar courses. My way to do it would be to create a many-to-many table of course ID to course ID (and joined with the course info), where one course points to other courses that are similar. That way you can also have some courses point to another course but not point back. Then in the admin panel the easiest (MVP) solution would be to have single text field with comma seperated course IDs and a check on the form if the courses exist and if the mappings are unique.

tomudding:
Added through GH-1727.

PR #1727, opened 2023-10-04
Add similar courses in course page

JortvD:
Created to solve #1726

### DIFF
--- a/module/Education/src/Service/Course.php
+++ b/module/Education/src/Service/Course.php
@@ -43,9 +43,9 @@ use function strlen;
 use function unlink;
 
 /**
- * Exam service.
+ * Course service.
  */
-class Exam
+class Course
 {
     protected ?BulkForm $bulkForm = null;
 
--- a/module/Education/src/Service/Course.php
+++ b/module/Education/src/Service/Course.php
@@ -486,6 +486,8 @@ class Course
 
     /**
      * Save a course.
+     *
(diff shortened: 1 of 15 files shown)


Koen1999:
Is there a way to populate `$similarCoursesFrom` in the model?

JortvD:
> Is there a way to populate `$similarCoursesFrom` in the model?

I have to preface that I'm a bit new to Doctrine ORM, but isn't it the case that `$similarCoursesFrom` is populated due to the fact it is the inverse of `$similarCoursesTo`? So if entries are added to the SimilarCourse table it will populate the related collections for those courses.

module/Education/src/Service/Course.php
```diff
@@ -485,13 +485,60 @@ public function getCourseForm(?CourseModel $course = null): CourseForm
...
+     * } $data
      */
-    public function saveCourse(CourseModel $course): bool
+    public function saveCourse(array $data): CourseModel
     {
+        $course = new CourseModel();
+        $course->setCode($data['code']);
+        $course->setName($data['name']);
+
+        $similarCoursesCodes = explode(',', $data['similar']);
```
Koen1999, line 502: I see you added it to the form, good! Will be quite some manual work though.

JortvD: Yep, but that's the job :)

tomudding (approved):
Looks good @JortvD! Because you will be using this interface it is fine for now, however, I do think we should make the selection process of courses easier in the future. For example, with an auto-complete search function.

Issue #1730, opened 2023-10-13
Add links to Supremum website in infimum panel


noelvandevrie:
### What would you like?

I'd like to see two links in the footer, one to allow users to directly go to the infimum submission page and the other to search.

PR #1731, opened 2023-10-13
Added links to Supremum website in infimum panel

noelvandevrie:
Allows users to directly submit an infimum or to search for one.

Closes GH-1730.


tomudding (approved):
Thank you @noelvandevrie!

PR #1749, opened 2023-11-01
Automatically expire `UserRole`s

tomudding:
To ensure that people do not retain privileges on the website for too longer, `UserRole`s will not automatically expire on a specified date.


tomudding:
@githubactions where are you

Issue #1482, opened 2022-06-28
`michelf/php-markdown` deprecation notice on internal method


tomudding:
Not an issue for us, but just adding it here to keep track of it.

> Deprecated: preg_replace(): Passing null to parameter &num;3 ($subject) of type array|string is deprecated in /code/vendor/michelf/php-markdown/Michelf/Markdown.php on line 233

Koen1999:
Perhaps we can pass this down to MaglMarkdown or the mentioned dependency

tomudding:
> Perhaps we can pass this down to MaglMarkdown or the mentioned dependency

There is a pull request open for it in the dependency, but the package does not appear to have had any activity since November last year. I am sure it will resolve itself at some point (before the PHP foundation decides to make this an actual error), but it is something we can keep an eye on.

tomudding:
The dependency has seen some action since June, however, this problem is still not resolved (PR is still open).

tomudding:
Due to the lack of updates to `MaglMarkdown` (and I do not want to create a pull request), I will create `gewisweb-markdown` which essentially forks the original. It should require minimal maintenance.

PR #1754, opened 2023-11-12
Fix name and content of `Activity` being `null` in news

tomudding:
Regardless if available translations, we always want to display an activity in the news feed if it is present.

Fixes GH-1482.


tomudding:
GH-1755

PR #1766, opened 2023-11-19
Remove `iconv` for decision search highlighting

tomudding:
Unfortunately, by using `iconv` there is a greater chance that the resulting (converted) text is either short or longer. For example, by using the euro symbol (€) we artifically increase the length of the texts we are comparing:

```php
iconv('UTF-8', 'ASCII//TRANSLIT//IGNORE', '€');
// 'EUR'
```

This is problematic, as it will result in incorrectly alignments of `<mark>`. While this can be mitigated by carefully calculating offsets for the offsets this quickly makes it more difficult to keep maintaining this functionality. Especially when there need to be more of these exceptions.

image

Only using the transliterator with `Any-Latin; Latin-ASCII` seems to preserve the length of the comparing elements and allow for searching accented/special characters. There are characters that are not part of/exist in `Latin-ASCII`, however, these characters are probably never used in the setting of the association.

This is a bug fix for GH-1764.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,6 @@
         "ext-exif": "*",
         "ext-fileinfo": "*",
         "ext-gd": "*",
-        "ext-iconv": "*",
         "ext-intl": "*",
         "ext-imagick": "^3.5.0",
         "ext-mbstring": "*",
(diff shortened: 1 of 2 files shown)


tomudding:
`€` is not the only symbol that causes problems for us, things like special quotes `’` have the same effect.

PR #1772, opened 2023-12-24
[GEWISDB] Move any `Member` association to `SubDecision` itself

tomudding:
Changes from GEWISDB upstream, see GEWIS/gewisdb#363. Necessary for GDPR export functionality.


tomudding:
Copying things is difficult 😞

PR #1779, opened 2024-01-09
Implement watermark tag

JortvD:
As specified in private communication


tomudding (approved):
Works as intended and performance hit is nothing compared to how long it already takes to make the PDFs.

I personally do not agree with _again_ increasing the opacity of the diagonal watermark. Just 1.5 years ago we already went from `0.65` to `0.5` because the then-EO wanted that. I do not want to have that discussion again, so I'll merge this as is with value of `0.35`.

Pending the last Actions run I will merge this in a bit.

PR #1789, opened 2024-02-08
Fix BAC website URL

Yoronex:
Replaced the old bac.gewis.nl link with the brand new and better mannendispuutdebac.nl.


tomudding (changes requested):
While I approve of buying random domain names when you do not need them, I do not approve of this change.

Issue #1791, opened 2024-02-22
ATHENA Function ordering on their fraternity page is wrong


Yaneriaf:
### Current behaviour

The current ordering is 
Chair
Treasurer
Vice-Chair
Secretary

### Desired behaviour

Preferably, this ordering would be
Chair
Secretary
Treasurer
Vice-Chair

This would be in line with the other fraternities (and organs for that matter) (except I.V.V, but that is due to people having double functions.)

### Steps to reproduce

Open the website
Open the fraternity page of ATHENA
It is there.

### Website version

v4.1

### What operating are you seeing the problem on?

Windows

### What browsers are you seeing the problem on?

Firefox

### Other information

Would be a nice first issue :)

tomudding:
Sorting of the functions happens here:

https://github.com/GEWIS/gewisweb/blob/d2091d798d33fd11cac2aa9f10b59669b0b1c223/module/Decision/src/Service/Organ.php#L410-L421

It only tries to put the chair first (or people with more functions).

PR #1795, opened 2024-03-14
Fix ordering of functions for committees and organs

tomudding:
Ordering is now guaranteed to follow:
- Chair
- Secretary
- Treasurer
- Vice-Chair

If someone has multiple functions then the highest function is used for the actual ordering. So someone who is both secretary and vice- chair will be above someone who is treasurer. Likewise, if someone is treasurer and vice-chair they will still be ranked below chair and secretary.

Fixes GH-1791.


tomudding:
Last PHP_CodeSniffer issue is not a problem, as we do not yet have the latest version it detects the type of the `const` as the name.

PR #1803, opened 2024-03-26
Fix to solve issue of leaving page during upload of images

### LINKED ISSUE #135: Photo: Show warning when leaving page during photo upload

Martijn-vanDijk:
Fixes #135 

Just a fix to make it not leave the page when uploading pictures to albums. (or atleast give a warning)

### DIFF
--- a/module/Photo/view/photo/album-admin/add.phtml
+++ b/module/Photo/view/photo/album-admin/add.phtml
@@ -93,6 +93,27 @@ $this->breadcrumbs()
     document.addEventListener("DOMContentLoaded", function() {
         Dropzone.options.imageDropzone = {
             acceptedFiles: 'image/*',
+            init: function () {
+                let uploading = false;
+
+                window.addEventListener("beforeunload", function (e) {
+                    if (uploading) {
+                        // Most modern browsers no longer support setting custom dialog messages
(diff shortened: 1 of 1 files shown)


module/Photo/view/photo/album-admin/add.phtml
```diff
@@ -93,6 +93,27 @@ $this->breadcrumbs()
...
+            init: function () {
+                let uploading = false;
+
+                window.addEventListener("beforeunload", function (e) {
+                    if (uploading) {
+                        // Most modern browsers no longer support setting custom dialog messages
+                        // for this dialog, but for backwards compatibility this could be nice.
+                        const warning = "You are currently uploading files, are you sure you want to leave?";
+                        e.returnValue = warning;
+                        return warning;
```
tomudding, line 105: This works great! However, I suggest adding

```js
e.preventDefault();
```

at the top inside the `if`-statement. That is the new way of actually enabling this event. The other options, setting `e.returnValue` and returning `warning`, are deprecated (imho the `return warning;` can be removed as `e.returnValue` has been supported in most major browsers since 2017 or earlier).

tomudding (changes requested):

PR #1807, opened 2024-03-30
Update dependencies

tomudding:
Removes the need for our `php-diff` patches as the functionality to display the whole diff when there are no changes has been added to the upstream.

Should also fix the failing PHP_CodeSniffer actions.

https://github.com/GEWIS/gewisdb/pull/384


tomudding:
Don't know why Psalm is complaining. There is no unused baseline entry as there is no entry in the baseline for that "error".

PR #1827, opened 2024-05-11
[GEWISDB] Change `SubDecision` `number` property to `sequence`

tomudding:
Changes from GEWISDB upstream, see GEWIS/gewisdb#394.


tomudding:
PHPStan is expected to fail, see https://github.com/GEWIS/gewisdb/pull/394.

PR #1880, opened 2024-07-31
fix: exploding login redirect links

tomudding:
# Description

Thought I had already fixed this, but this was unfortunately not the case as I still saw Amazon and Meta having issues with this.

Quite simple, `redirect_to` already in the URL? Than we provide a new clean URL.

## Types of changes

- [X] Bug fix _(non-breaking change which fixes an issue)_


tomudding:
Not the nicest solution but it works

PR #1912, opened 2024-10-20
chore: add Doctrine migrations

tomudding:
# Description

Adds support for migrations while using Doctrine ORM. The initial migration is based on the current `main`.

Normally, a `migrations:rollup` would be used to ensure that this migration is successful. However, in our case it is necessary to export ALL data (but not structure) from the database, drop everything in the database, and only then perform the migration.

This ensures that _all_ FKs are the same for newer migrations preventing headaches in the future.

---

Convenience methods have been added for `make` to make using the migrations easier. However, in production everything must go directly through `./orm`.

## Types of changes


tomudding:
Part of a series of enhancements to hopefully make maintainability easier.

Issue #1890, opened 2024-08-23
Other Organs Not Visible on Committee Page


BHenkemans:
### What would you like?

Currently, "other organs" (RvA, KCCs, and AVWs) are not visible on the committee page. However, some people may want to get information about these organs (for instance, to contact the organ or one of their members). The only way to do this right now is to guess someone in that organ and use the link of that organ on their member page.

### Why is this needed?

The current method makes it difficult to find these organs if you don't know anyone in the body.

BHenkemans:
I suggest making another page or a tab on the current committee page if making a distinction is even desirable.

tomudding:
The `avc|kcc|avw|rva` are relatively new; see GH-1057. Before that, everything was grouped under `avc` (not that you could view those). The other types do not appear under the `Committees` section, as it logically follows that `committee` is a specific type and not also an umbrella for `avc|kcc|avw|rva`.

I am firmly against adding another entry to the navbar. As such, there are two probable solutions: We can either consolidate all "organs" into one navbar entry and have a submenu for each "flavour" or have a nav-pill bar on the `committees` page, similar to how we allow filtering on meeting types. However, I also do not like these two options either.

BHenkemans:
> However, I also do not like these two options either.

Would it hurt to pretend that these are "just regular committees"? As I mentioned before, for most members, these are just committees. The types (in my eyes) are purely regulatory.

tomudding:
> Would it hurt to pretend that these are "just regular committees"? As I mentioned before, for most members, these are just committees. The types (in my eyes) are purely regulatory.

I am getting old, so I like to complain more. I think the second option of adding a nav-pill bar is the best solution.

PR #1924, opened 2024-11-27
feat(frontpage): add overview for GMM bodies

tomudding:
# Description

Apparently this was brought up at the last GMM, but unfortunately it remains difficult to communicate things. Fortunately, there was an issue open for this before, so it was already clear what needed to be done without me receiving the additional information.

This adds a new dropdown to the 'Association' menu for "GMM Bodies":
- Advisory Boards,
- Financial Audit Committees,
- GMM Taskforces, and
- GMM Committees.

Each links to a new page for an overview of that specific type of GMM body. The page also includes a historical overview of those GMM bodies.

No other new functionality has been added, as everything necessary for this was already implemented.

---

While we have always used "organs" instead of "bodies" this is an incorrect translations (like many of our translations in the past). This is the first attempt to try and fix that.

---

Opted to not add notice that the Financial Audit Committees before 2021 were actually GMM Committees.

Closes GH-1890.

## Types of changes

### DIFF
--- a/module/Application/language/en.po
+++ b/module/Application/language/en.po
@@ -8,8 +8,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: GEWISweb 0.1.0-dev\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-13 21:40+0100\n"
-"PO-Revision-Date: 2024-11-13 21:41+0100\n"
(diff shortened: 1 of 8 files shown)


tomudding:
I went for the additional entry in the menu bar, even though I was against it a few months ago. I checked with our sister associations how they do this; only one association has something similar.

PR #1903, opened 2024-09-20
wip: improved `OrganInformation` approval

tomudding:
# Description

## Related issues/external references

Closes GH-1383, closes GH-1858.

## Types of changes

- [X] Bug fix _(non-breaking change which fixes an issue)_
- [X] New feature _(non-breaking change which adds functionality)_
- [X] Breaking change _(fix or feature that would cause existing functionality to change)_
- [ ] Documentation improvement _(no changes to code)_
- [ ] Other _(please specify)_

PR #1934, opened 2024-12-07
feat: improved cards for bodies, companies, and jobs

tomudding:
# Description

Started out as a rewrite for the broken cards used to display BM and GMM bodies. I believe I broke their spacing many years ago when refactoring something else, but I never had the need to fix it (as people stopped complaining).

The cards for companies always worked but looking back at the Sass code, they were just interfering with each other in which the body overview lost.

Cards are now unified and follow the newer Boostrap styling more closely.

---

Intention was to fix the styling for BM/GMM bodies where the abbreviation of the body was not visible due to the usage of picture with a white(-ish) background. This has been resolved by moving all text back into the body of the card instead of on top of the cover image. Sizing and grid layout are handled through CSS.

If there is no cover image, the GEWIS logo is now shown (similar to how this is done in the Photo module).

---

**Before:**
image

**After:**
image

## Types of changes


tomudding:
This was supposed to be part of GH-1903; however, since that is taking longer than expected, I am moving this forward (note: these changes were never on that branch, only in my head up to a few hours ago).

Issue #1824, opened 2024-04-30
Move everything to a `Factory`


tomudding:
Instead of doing

```php
'module_type_name' => static function (ContainerInterface $container) {
    // ...
},
```

we should do it properly by using a `Factory`:

```php
ImportantClass::class => ImportantClassFactory::class,
```

This makes it easier to see, remember, and use the actual classes. Furthermore, it helps PHPStan actually checking if what we are `$container->get()`ing exists and removes a lot of `@var` type hints.

PR #1951, opened 2024-12-25
chore: move everything to factories

tomudding:
# Description

Closes GH-1824.

## Types of changes


tomudding:
Bit brain-dead (if I was not already) from repetitively creating so many files. Unfortunately, I just remembered that most of the factories are not needed at all because all mappers can use a common factory, and so can most forms (the ones that only need a translator).

---

Psalm errors about `WatermarkService` can be ignored due to change in baseline. PHPStand also fails in a similar way because class names have changed.

Issue #1921, opened 2024-11-20
Marking participants of activities present/not present


tomudding:
### What would you like?

Functionality that allows organisers of an activity to mark participants of their activity as present/not present.

### Why is this needed?

C4 has asked for this to make their work easier for lunch lectures. This functionality can also be useful for other organs in the association, so it must be available to all.

### Other information

I think marking someone present should only be possible from 30 minutes before the activity starts to 30 minutes after the activity has ended. There is no point in marking someone present/not present outside this window (perhaps up to an hour after but not more to prevent abuse - maybe it is should use `TimestampableTrait`?).

tomudding:
Closed via GH-1973

PR #1973, opened 2025-01-18
feature: attendance mode

LuukBlankenstijn:
- Docker bind mount for development
- Stimulus support
- Typescript support
- Add attendance mode where activity admins can take attendance of participants

# Description
Started working on attendance mode. Really wanted to use Stimulus, so added it, including typescript. Then while developing I got really frustrated by the constant rebuilding of the dev container to see changes, so changed it, so now the host system is mounted in the docker container. This means changes will take effect immediately, without doing anything. It made the development experience a lot nicer.

The feature I added was the attendance mode. In the admin-overview of an activity at a signup-list, there is now a button 30 minutes before to 30 minutes after the activity that allows the user to go into signup mode.
image
Here they can search for participants that have not been marked present yet and mark them present. This will be send to the server. The server will respond with information to sync the list in the client. Since I did most of it in the frontend with Stimulus, the UI is really responsive.

All of this happened kind of at the same time, this is the reason it is combined in one pull request. Separating it would take an immense amount of time.

Fixes issue GH-1921

Fixes GH-NNN.

## Types of changes

- [X] Dev-environment


tomudding (approved):
Thank you @LuukBlankenstijn!

I have a few small remarks (see below), but I have already resolved them while fixing the Psalm and PHPStan errors. Will push the changes in a bit.

tomudding:
Both Psalm errors should be resolved once merged (because the action using the old baseline).

tomudding (approved):

PR #2016, opened 2025-04-27
fix(attendance-mode): fix wrong default value for Signup entity

LuukBlankenstijn:
# Description
A small bug I found that caused me a lot of confusion. To fix everything, it would be nice to run the following query against the live database: ``UPDATE SIGNUP SET drawn = TRUE``. Since the drawing functionality has not really been used yet, it does not matter this resets everything.

## Types of changes

- Bug fix _(non-breaking change which fixes an issue)_


tomudding:
Maybe I do not understand the functionality, but the default should be `false` not `true`? Because you cannot be drawn by default, however, then the following should be set to `true`?

https://github.com/GEWIS/gewisweb/blob/0da869798b56f3a20790acce3cd114142aaddd7f/module/Activity/src/Controller/AdminController.php#L619

LuukBlankenstijn:
Well, right now the default is true. Since there is no policy (implemented) (yet), everyone is "drawn" in on default, and then you can "draw" them out in the admin interface. The actual draw is made by the used by hand.

PR #2162, opened 2026-08-15
feat(frontpage): build out the module around polls, news and pages

### LINKED ISSUE #1729: Info toast for student e-mail address users
### What would you like?

From ABC-2304-100, after Bootstrap 5 migration is finished we should add a toast to let members know they are still using their student e-mail address.

### Why is this needed?

Because we cannot reach them after they graduate.

### Other information

_No response_

### LINKED ISSUE #1784: Reactions on poll comments
### What would you like?

I think it would be nice to allow anonymous reactions to poll comments. The options should be "Like", "Love", "Insightful", and "Funny". No negative options because I am not waiting for unnecessary whining from people who are not happy with a negative reaction to their comment.

You can react only once per comment but you can change your reaction. Once the poll has ended the reactions are set in stone and can no longer be altered.

### Why is this needed?

_No response_

### Other information

_No response_

### LINKED ISSUE #1812: Allow replies to poll comments
### What would you like?

Allow replying to (at least) a top-level comment on a poll. These replies are then shown directly under this comment with some identation or special styling to show that they are a reply.

This will likely also require a change in how we display the datetime and author of a comment. Perhaps we should move to a "social media" approach to displaying this information (maybe time being shorter like 1min. ago, 2 days ago).

### Why is this needed?

Sometimes there are extensive discussions in the comments which result in an unclear order that you have to read the comments in.

### Other information

_No response_

### LINKED ISSUE #1816: Rotate birthday picture
### What would you like?

When multiple people share their birthday, do not show just one picture but have a few on rotation (from the different people).

### Why is this needed?

Seeing one picture the whole day is a bit too much.

### Other information

_No response_

### LINKED ISSUE #2046: Restrict URL slugs to not start with numbers or symbols
### What would you like?

Currently, most URL slugs need to adhere to this pattern:

```regex
/^[0-9a-zA-Z_\-\.]*$/
```

However, that should be:

```regex
/^[a-z][0-9a-z\_\-]*$/
```

### Why is this needed?

There is no need for slugs to start with a number or symbol. Furthermore, there should not be a literal `.` in the URL.

### Other information

_No response_

### LINKED ISSUE #2050: Change custom page URL slug pattern
### What would you like?

```regex
/^[a-z][0-9a-z\_\-]*$/
```

### Why is this needed?

_No response_

### Other information

Controller + form + migration

tomudding:
# Description

The frontpage module was the last area of the website that was still shaped the way the Laminas application left it. The poll, news and page entities were ported, but nothing much was built on top of them: there was no public poll surface, no news rendering beyond the entity, no page administration at all, the front page showed only the agenda, the photo of the week and the birthdays, and `/admin` was an empty shell. This branch finishes the module.

Polls now follow the revision workflow the activities, companies, vacancies, and body pages already use. A poll used to be approved by writing down who approved it, which gave the board no way to say no and no way to explain itself. The question and its answers now live on a `PollRevision` the board decides on, while the poll itself keeps what must survive a decision: the votes, the discussion, and the date it closes. A poll is never edited, so there is no draft, no cloner, and no edit lock, and requests for changes are withheld: a question members have started answering cannot be rewritten underneath them. Asking again after a rejection writes a new revision from scratch with the previous wording carried over. The reviewer fills in the closing date when approving, because approving a question is also scheduling it, and taking a poll down early moves that date to today rather than throwing away what was asked. Submitting raises the usual board notification, and the legacy email goes away with it.

Members get the full poll surface: the current question, an archive of everything the association was ever asked (searchable and paginated, since there are well over a thousand), and a page for putting a question forward. Answering and talking underneath happen in place. The widget swaps to the results once this reader has answered, so the bars never steer an answer that hasn't come yet. Comments gain reactions and one level of replies, both of which are the substance of GH-1784 and GH-1812. A comment is signed with a name the member picks, and only the board sees which member that was. A month after a poll closes, the answers stop being anybody's in particular: each option keeps its tally as a number, the rows saying which member gave it are removed, and the reactions keep their counts and lose the members behind them. That runs daily on the GDPR schedule and stamps the poll, so a second run leaves it alone.

News gains a category so a reader can narrow it to the part of the association they care about, plus a full admin round trip and a public archive at `/news` with a page per item, grouped by month and narrowed by year, category and search the way the activity and photo overviews already read. The members area carries a short list of the latest items. Nothing names who wrote an item, because news is the association speaking rather than a person.

The front page is rebuilt around that news. The panorama becomes a hero carrying the association's name and the two things a visitor is most likely to want next; underneath it, the news runs down the page with the agenda, the photo of the week, the birthdays, the current poll, and the infimum in a rail beside it. What the board pinned is set apart by a strip saying so, and the company banners sit inside the feed rather than under it, so the companies that paid for the space are read past rather than scrolled to. The birthday panel used to show a photo of whichever member had been tagged most, meaning it was about one of them all day; now it cycles through a photo of each of them, and there is one of GH-1816. The server renders every photo, hiding all but the first, so a reader without JavaScript still gets one.

The infimum comes from the Supremum's own API, and nothing here asks for it while a page is being drawn. A cron fetches a fresh one every five minutes, keeps it, publishes it on a shared members topic, and the page draws with a placeholder that fills in from an endpoint of ours. The application's single EventSource re-dispatches the rotation as a DOM event the panel listens for, so there is no second connection and no page reload. A failed fetch is cached as well, so a down API costs one request in a while rather than one for every reader who arrives.

Members reachable only at their university address now get a notice asking them to have the board or the secretary change it (GH-1729). It runs alongside the real announcements, so it closes the same way and stays closed. There is no link on it, since changing the address is something the board does rather than something a page can offer.

The page an administrator lands on now names every review queue they may deal with, how much is in it and how long the oldest submission has been sitting, with the five that have waited longest across all queues listed together. A count says how much there is; only the oldest says whether anything has been forgotten. Each block is gated on its own, because any active member can reach the page. Revisions gained a real `submittedAt` for this, stamped when they enter the submitted state and backfilled from `createdAt`.

Custom pages could be read but not written, unless you edited the database. There is now a two-step form for making, changing, and taking down a page, with the address written as the URL it becomes and the content written in the same editor the rest of the site uses, kept as HTML. A page is addressed by its own words rather than by an ID, so each segment is held to the shape of a URL segment: three to thirty-two characters, lower case, starting with a letter, and nothing in it but letters, digits, underscores and hyphens. That closes GH-2050 and the last part of GH-2046 that was still outside the shared slug rule, and a migration brings the addresses already stored down to lower case. Two pages may not answer to the same address in the same language, and no page may take an address the application already answers to itself.

A page is stored as HTML and rendered as is, which until now meant whoever could write one could put anything on the website, up to and including a script that runs for every reader. What a page may contain is now settled in configuration with `symfony/html-sanitizer` and applied on the way in, wherever a save comes from. Structure, text, tables, links, our own images and the two video hosts the content security policy already allows to be framed all pass; everything else is dropped by default. We keep an image only when it is served from our own image pipeline, so a page cannot depend on a file somewhere else on the internet that can change size, move, or disappear. A one-off command puts the pages written before any of this through the same sanitiser, and is deliberately not scheduled.

Two things to note for deployment. The key the Laminas application reads as `DOCKER_SUPREMUM_API_KEY` iY` alongside a host it can be pointed at, which brings it in line with how the GEWISDB API is configured; production has to be given the new name. There are three migrations: one for polls, one for news, and one for pages, the last of which rewrites stored page addresses.

Fixes GH-1729.
Fixes GH-1784.
Fixes GH-1812.
Fixes GH-1816.
Fixes GH-2046.
Fixes GH-2050.

## Types of changes
- [X] Breaking change _(fix or feature that would cause existing functionality to change)_
- [X] Other _(please specify)_: an environment variable is renamed and has to be set in production before this is deployed.


tomudding:
The first PR where the core logic is largely the work of Claude. Unfortunately, it required a lot of fixes, so I think I will go back to my usual workflow of having Claude do only smaller, clearly defined parts.

